PR #2, opened 2026-01-11
MILAN Advertise State Controller Implementation 

### LINKED ISSUE #1: Implement ADP Advertise Controller
- FPGA Module for driving the ADP Advertise State Machine

cemaldogann:
- Advertise state implementation of the Milan Specs v1.2 #1


hdl/adp/KL_advertise_controller.sv
```diff
@@ -0,0 +1,189 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Oguz Kahraman <oguz.kahraman@kebag-logic.com>
```
OguzKahramn: name should be changed

hdl/adp/KL_advertise_controller.sv
```diff
@@ -0,0 +1,189 @@
...
+    input wire clk_i, //! Global clock
+    input wire rst_n, //! Active-low Reset 
+    input wire [63:0] rcv_entity_id_i,  //! entity_info.rcv_entity_id - from KL_adp_parser
+    input wire [63:0] entity_id_i,      //! entity_id of the ATDECC entity - from memory map 
+    input wire [63:0] grandmaster_id_i, //! grandmaster_id - from PTP module
+
+    input wire rcv_adp_discover_i, //! ADP Discovery packet is received and rcv_entity_id_i field is valid- from KL_adp_parser
+    input wire link_down_i, //! Link down event from Upper Management Module (TBD)
+    input wire link_up_i,   //! Link up event from Upper Management Module (TBD)
+    input wire shutdown_i,  //! Shutdown evenet from Upper Management Module (TBD)
```
OguzKahramn: maybe we can create a type for this `status` so we can have 1 line input, or we can use one-hot and [2:0] bit input could be better imo

hdl/adp/KL_advertise_controller.sv
```diff
@@ -0,0 +1,189 @@
...
+`default_nettype none
+import adp_pkg::*;
+
+module KL_advertise_controller
+  (
+    input wire clk_i, //! Global clock
+    input wire rst_n, //! Active-low Reset 
+    input wire [63:0] rcv_entity_id_i,  //! entity_info.rcv_entity_id - from KL_adp_parser
+    input wire [63:0] entity_id_i,      //! entity_id of the ATDECC entity - from memory map 
+    input wire [63:0] grandmaster_id_i, //! grandmaster_id - from PTP module
```
OguzKahramn: we can just input `entity_info_t` from `adp_parser` for these - it will be shorter and scalable imo - if different components are useful which are stored in `entity_info_t`

hdl/adp/KL_advertise_controller.sv
```diff
@@ -0,0 +1,189 @@
...
+  //! for the ATDECC Entity.
+  always_ff @(posedge clk_i) begin : grandmaster_control
+    if (!rst_n) begin
+      grandmaster_id_r <= 64'd0;
+      gm_change_r <= 1'd0;
+    end
+    else begin
+      grandmaster_id_r <= grandmaster_id_i;
+      if (grandmaster_id_r != grandmaster_id_i)
+        gm_change_r <= 1'd1;
```
OguzKahramn: this register is redundant imo, I would register `grandmaster_id_i` or `entity_info.grandmaster_id`  and assign it on `assign` block:

` assign advertise_event_o.GM_CHANGE = grandmaster_id_i  != grandmaster_id_r;`

hdl/adp/KL_advertise_controller.sv
```diff
@@ -0,0 +1,189 @@
...
+            cntrller_state <= IDLE_S;
+        end
+
+        UPDATE_EVENT_S : begin
+          rcv_adp_discover_r <= 1'd1;
+          cntrller_state <= IDLE_S;
+        end
+      endcase
+    end
+  end
```
OguzKahramn: just an idea: Maybe the state machine is necessary idk but this can be done in this way:

`assign advertise_event_o.RCV_ADP_DISCOVER = rcv_adp_discover_i & (rcv_entity_id_i == 64'd0 || rcv_entity_id_i == entity_id_i);`

hdl/adp/KL_counter.sv
```diff
@@ -0,0 +1,151 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Oguz Kahraman <oguz.kahraman@kebag-logic.com>
```
OguzKahramn: name should be changed

hdl/adp/KL_advertise_controller.sv
```diff
@@ -0,0 +1,189 @@
...
+    .completed_o   ( tmr_delay_completed_r )
+  );
+// --------------- MAIN LOGIC --------------- //
+
+  //! Free-running counter logic - value in between 0-4
+  always_ff @(posedge clk_i) begin : counter_logic
+    if (!rst_n) 
+      zero_four_sec_cnt_r <= 3'd0;
+    else begin
+      if (zero_four_sec_cnt_r == 3'd4)
```
OguzKahramn: we can store contstant value as localparam and then we can declare the counter length in scaleble way:
```
localparam int  MAX_SECOND = 'd5;
logic [$clog2(MAX_SECOND)-1:0] zero_four_sec_cnt_r;
```

hdl/adp/KL_advertise_controller.sv
cemaldogann: When the parsing is finished on the KL_ADP_Parser, it gives the rcv_adp_discover with the all parsed fields with the entity_it as well. Since KL_ADP_Parser needs clock cycle to go idle, i believe your approach can be done and better!!

hdl/adp/KL_advertise_controller.sv
cemaldogann: copy paste :((

hdl/adp/KL_advertise_controller.sv
cemaldogann: You are right. For now, I suggest that we keep it that way and when the upper management module is designed, we can definitely do that

hdl/adp/KL_advertise_controller.sv
```diff
@@ -0,0 +1,132 @@
...
+
+    input wire rcv_adp_discover_i, //! ADP Discovery packet is received and rcv_entity_id_i field is valid- from KL_adp_parser
+    input wire link_down_i, //! Link down event from Upper Management Module (TBD)
+    input wire link_up_i,   //! Link up event from Upper Management Module (TBD)
+    input wire shutdown_i,  //! Shutdown evenet from Upper Management Module (TBD)
+
+    input wire start_tmr_delay_i,     //! Start TMR_DELAY counter command - from KL_advertise_state Module
+    input wire start_tmr_advertise_i, //! Start TMR_ADVERTISE counter command - from KL_advertise_state Module
+    input wire stop_tmr_delay_i,      //! Stop TMR_DELAY counter command from KL_advertise_state Module
+    input wire stop_tmr_advertise_i,  //! Stop TMR_ADVERTISE counter command from KL_advertise_state Module
```
OguzKahramn: imo, very long list of input. I'm wondering somehow we can create  `typedef` for clear port list, my suggestion to create header files for each module of AVB (not hdl module, like ADP.svh, AECP.svh etc). In this way, it will be easier to scale in the future and also it will be easier to understand in hdl read/development for protocol level.  My suggestion:

* Adding similar semantic signals together in a struct and use this structs to not make port noise in hdl, this struct does not have to be in the specification, just for internal usage. The same struct can be used in `parser` module as well to assign values so it will decrease port list in this module as well.

create or edit(if it already exists) file `adp.svh` :

```
typedef struct packed {
  logic start_delay;
  logic start_advertise;
  logic stop_delay;
  logic stop_advertise;
} tmr_events_t;
```

we can use this struct in the port list:
```
...
input tmr_events_t tmr_event_i,
...
```

and feed/assign signal into the module:

```
if (tmr_event_i.start_delay)
  start_delay();
else if (tmr_event_i.start_advertise)
  start_advertise();
...
```

hdl/adp/KL_counter.sv
```diff
@@ -0,0 +1,151 @@
...
+    input wire clk_i,
+    input wire rst_n,
+    input wire start_i,
+    input wire stop_i,
+    input wire [2:0] delay_value_i,
+
+    output wire completed_o
+  );
+
+  typedef enum bit [2:0] {
```
OguzKahramn, line 38: it is ok but if the number of states increase, then this needs to be updated as well, I would use int therefore just adding state would be necessary

hdl/adp/KL_counter.sv
```diff
@@ -0,0 +1,151 @@
...
+        end
+
+        FINISH_S : begin
+          completed_r <= 1'd0;
+          high_counter_val_r <= 16'd0;
+          med_counter_val_r <= 16'd0;
+          low_counter_val_r <= 16'd0;
+          cnter_state <= START_S;
+        end
+      endcase
```
OguzKahramn, line 108: default state needs to be added

tb/utests/ptp_timestamp/tb_ptp_ts_core.tcl
```diff
@@ -37,7 +37,7 @@ variable script_file
 set script_file "tb_ptp_ts_core.tcl"
 
 # Create project
-create_project ${_xil_proj_name_} ./${_xil_proj_name_} -part xc7a35tcpg236-1
+create_project ${_xil_proj_name_} ./${_xil_proj_name_} -part xc7z020iclg400-1L
```
OguzKahramn, line 40: is it intended to change ?

kl-verificator (changes requested):
## Review: MILAN Advertise State Controller ⚠️ Request Changes

### Recommendation: **Close this PR** — it is fully superseded by PR #8

PR #8 (`03_adp_advertise_state`) includes every file from this PR plus additional modules. Merging both would create conflicts and confusion. Please close PR #2 and continue with PR #8.

---

### Issues found (carry forward to PR #8)

**1. `adp_advertise_event_t` is not `packed`** (`adp_pkg.sv`)
```sv
typedef struct {          // ← missing 'packed'
  logic RCV_ADP_DISCOVER;
  logic TMR_ADVERTISE;
  ...
} adp_advertise_event_t;
```
Non-packed structs have implementation-defined bit ordering. Compare with `link_status_t` and `tmr_events_t` which are correctly `packed`. This was also raised in PR review comments. Fix: add `packed`.

**2. TMR_DELAY startup range deviates from spec**
Milan spec §5.6.3.5.2: startup with link up → TMR_DELAY = **0–2 seconds**.
Milan spec §5.6.3.5.3: LINK_UP event → TMR_DELAY = **0–4 seconds**.
The free-running `zero_four_sec_cnt_r` always gives 0–4 regardless of context. Document this deviation or add a startup flag to cap at 2 s on first LINK_UP.

**3. No guard against simultaneous LINK_UP + LINK_DOWN**
Milan spec §5.5.1.4.2 requires LINK_UP and LINK_DOWN to be mutually exclusive. Add an `assert` or input guard.

---

### Simulation result
All 5 unit tests PASS under Verilator 5.048 with clock-scaled counter.

Issue #67, opened 2026-08-16
AECP: SET_CONFIGURATION / SET_STREAM_FORMAT / SET_STREAM_INFO / START+STOP_STREAMING unimplemented — the STREAM_IS_RUNNING predicate has no wire into the engine


Mister-M-alt:
## Summary

Four Milan v1.2 SHALL commands share one missing predicate and are therefore one
issue: `SET_CONFIGURATION` (`0x0006`), `SET_STREAM_FORMAT` (`0x0008`),
`SET_STREAM_INFO` (`0x000E`) and `START`/`STOP_STREAMING` (`0x0022`/`0x0023`). All
five answer the conformant `NOT_IMPLEMENTED` echo today.

The predicate is **"is this stream running?"**, and Milan spells it two different
ways depending on direction. Getting that distinction wrong is the defect
[#58](https://github.com/kebag-logic/milan-fpga/issues/58) recorded against the
now-deleted plane; this issue is that clause reading carried forward onto the
protocol processor, plus the three commands #58 did not cover.

Landed context (VERSION `0x004C`): `KL_aecp_dyn_state.sv` holds the settable
state, including a `started/stopped` bit per Stream Input and a presentation-time
offset per Stream Output, and publishes them to the fabric. `CHECK_LOCK` and the
full-body refusal stubs (`E_LOCKED4`, `E_BADARG4`, `E_TIZ4NS`) exist and are
tested. What is missing is the **streaming predicate** and the four dispatch arms.

## The predicate, settled

Milan uses two different words and they are not interchangeable:

| Direction | Milan's word | What it means |
|---|---|---|
| Stream **Input** | **bound** | the pure ACMP bind level. §5.3.8.2 is explicit that bound *"does not necessarily mean that data is flowing"*. **Not** reservation, **not** started. |
| Stream **Output** | **streaming** | declaring a Talker Advertise **and** a Listener Ready or Ready Failed registered (§5.3.7.3) |

Both already exist in the processor as the wires the hardware obeys — the ACMP
listener's bind record and the SRP talker declaration state. Neither is wired to
the AECP engine yet, and that wire is the bulk of this issue.

The refusal is **value-independent**: a same-format `SET_STREAM_FORMAT` on a bound
input refuses too. The pre-bind window stays fully writable, so the ordinary
controller flow (set the format, then connect) is untouched.

```mermaid
flowchart TD
    CMD[SET_STREAM_FORMAT] --> L{locked by another controller?}
[...]

Issue #70, opened 2026-08-16
Saved state: nothing survives a power cycle — the NVM face is a blank-flash stub and reports a restore that never happened


Mister-M-alt:
## Refined 2026-08-22 (maintainer instruction: refine the On Hold tickets; READY, sequenced behind #69)

**Decision:** this ticket keeps the saved-state acceptance and now also owns the clean re-landing of the NVM power-cut coverage; #178 is consolidated here and closed.

**State of the world (measured 2026-08-22):**
- Donor `main` is `44489453`, which carries protocol-processor PR #13 (`tb/nvm_port`: a torn commit cannot lose other records; 90/90) together with the #69 notification work. There is no donor commit with one without the other, so the pin bump is #69's; this ticket starts its parent-side work once that pin is on `dev`.
- The parent side is the actual gap: the NVM face is a blank-flash responder (reads 0xFF, writes discarded) and `A_PP_STAT` reports a successful restore of zero records. Nothing persists.

**Work (parent side, after the #69 pin lands):**
1. Decide and record the backing store publicly before implementing: a dedicated QSPI flash region behind the existing SPI-flash controller (the boot flash already holds bitstream + kernel; a sector-aligned record region is the natural home), versus a DRAM shadow committed to flash by firmware/driver. Measure the area of the flash path before committing (USER: keep usage low, re-use the existing controller).
2. Wire the processor's NVM port to that store; a restore walk that finds blank flash must report "nothing restored", never success.
3. Persist and restore the eight Milan items and the binding state listed below; prove power-cut safety with the donor's nvm_port semantics in the parent (a torn commit cannot lose other records).
4. Parent tests: a Verilator bench that writes, power-cycles (reset with retained flash model), restores, and fails when any record is missing; negative control by reverting the store.
5. Docs: `docs/reference/REGISTER_MAP.md` (`A_PP_STAT` semantics), the persistence page, the feature ledger.

Acceptance as in the original body. Closed into this ticket: #178 (re-land lane).

---

## Summary

**Nothing in this device survives a power cycle.**

Milan v1.2 names eight things a PAAD-AE shall save in non-volatile memory and
restore. The processor's NVM face is a blank-flash responder: reads return `0xFF`,
writes are accepted and discarded, erase completes. A restore walk always finds
blank flash and completes with zero records — and reports success, which is worse
than failing, because `A_PP_STAT` then shows a restore that never happened.

This blocks the saved-state compliance tests completely. They are among the few
[...]

Issue #69, opened 2026-08-16
AECP: the unsolicited-notification trigger set is empty, and the departing-controller monitor (Milan 5.4.5.3) has no originator — registry lands, nothing fires


Mister-M-alt:
## Refined 2026-08-22 (maintainer instruction: refine the On Hold tickets; this one resumes as READY)

**Decision:** the held implementation lane resumes. This ticket is the compliance task AND the implementation lane; the separate re-land ticket #183 and the dependency record #169 are consolidated here and closed.

**State of the world (measured 2026-08-22):**
- The reviewed donor mechanism is merged and durable: protocol-processor PR #11 ("Implement AECP notification triggers and monitoring") and PR #12 ("Fix notification review findings") are on donor `main`, which is `44489453` today (10 commits past the parent pin `a25b5cc9`; 13 HDL files changed, +1448/-200). The same donor `main` carries PR #13 (`tb/nvm_port` power-cut coverage, #70), so ONE pin bump serves both tickets; #70 follows this one.
- The bump adds four top-level inputs to `hdl/top/protocol_processor_top.sv` that the parent instance at `hdl/milan/KL_pp_shadow.sv:864` does not drive: `ctr_change_i` (one served descriptor counter changed), `ctr_change_desc_type_i[15:0]`, `ctr_change_desc_index_i[15:0]`, `gsi_asp_chg_i`. The lint ratchet fails `PINMISSING` on them (measured on the retired PR #145: 0 with all four connected, 3 with one removed). A tie-off is NOT acceptable (it makes the trigger permanently silent while the gate goes green); each must be driven by the parent signal that carries the clause's meaning (the descriptor-counter change from the datapath's counters, the AS_PATH / GET_STREAM_INFO change).
- Source lists are derived now (#170, `scripts/pp_srcs.py` in every consumer), so the manifest omissions that broke the Yosys OOC flow on #145 cannot recur by hand; the OOC script `syn/ooc/pp_shadow_ooc.tcl` must still be checked against the derived list.
- Parent VERSION and the register map must move if the CSR face changes; `docs/reference/MILAN_FEATURE_STATUS.md` and `docs/reference/REGISTER_MAP.md` follow.

**Re-land scope (from #183, now here):**
- Start from current `dev`; pin `protocol-processor` to `44489453` (or the then-current donor `main`) in its own commit; drive the four inputs with real parent signals and document each driver with its clause.
[...]

Issue #68, opened 2026-08-16
AECP: ADD_AUDIO_MAPPINGS / REMOVE_AUDIO_MAPPINGS unimplemented — dynamic input mappings are mandatory for a Milan listener and the validation is all-or-nothing


Mister-M-alt:
## Summary

`ADD_AUDIO_MAPPINGS` (`0x002C`) and `REMOVE_AUDIO_MAPPINGS` (`0x002D`) are Milan
v1.2 SHALL commands and both answer the conformant `NOT_IMPLEMENTED` echo.

The **read** half already works: `GET_AUDIO_MAP` (`0x002B`) has served both port
directions since VERSION `0x0048`, off the integrator's map store through the
audio-map gather face. What is missing is the write path and, more than that, the
**validation** — Milan turns several things IEEE leaves vendor-defined into
SHALLs, and every one of them is all-or-nothing.

Milan §5.3.3.7 makes this mandatory rather than optional for inputs:

> The Stream Port Input of a Configuration shall not contain any AUDIO_MAP
> descriptor.
> Note: this is because a PAAD-AE has to support dynamic mappings on input.

So a Milan listener has **no static input mappings at all** — dynamic mappings are
the only way channels reach clusters, and a controller that cannot add them cannot
route audio.

## The validation rules, all SHALLs

IEEE §7.4.45 leaves mapping validity "vendor defined". Milan does not:

> A PAAD-AE shall treat as invalid a mapping that references a channel of a Stream
> Input/Output that does not exist in the currently set format for this Stream
> Input/Output.

> …shall return the BAD_ARGUMENTS error code … if the command contains two
> different mappings that reference the same cluster's channel and two different
> stream's channels that are not redundant. **In this case, no mapping shall be
> added.**

> A PAAD-AE shall ignore duplicate mappings that may be present in a
> REMOVE_AUDIO_MAPPINGS command.

> …it shall not allow removing a mapping that references a channel of a currently
> streaming Stream Output. In this case, the PAAD-AE shall return a BAD_ARGUMENTS
> error code and no mapping shall be removed.

And one that is only a **SHOULD**, so it must not be graded as a SHALL:

> A PAAD-AE **should** treat as invalid a mapping that references a channel of a
> Stream Input/Output whose sampling rate is different from the sampling rate of
> the Audio Unit, if none of the SSRC and ASRC bits are set…

```mermaid
flowchart TD
[...]

PR #77, opened 2026-08-16
AECP: the STREAM_IS_RUNNING predicate and SET_CONFIGURATION (VERSION 0x004D)

### LINKED ISSUE #82: AECP: READ_DESCRIPTOR(ENTITY).current_configuration and GET_CONFIGURATION can disagree — IEEE 7.4.8.2 calls them equivalent
## Summary

`GET_CONFIGURATION` (`0x0007`) and `READ_DESCRIPTOR(ENTITY)` (`0x0004`) can
answer with **different values for the same field**. IEEE 1722.1-2021 §7.4.8.2
says they may not:

> The `configuration_index` field is set to descriptor_index of the current
> configuration. This is equivalent to the `current_configuration` field in the
> ENTITY descriptor.

Since VERSION `0x004D`, `SET_CONFIGURATION` stores the index in the AECP dynamic
state store and `E_GCFG` overlays that store on the way out. `E_RDESC` was not
changed: it still `COPY_BUF`s the ENTITY descriptor straight out of the
descriptor image, so `current_configuration` is whatever
`avdecc/gen_aem_store.py` baked in.

Measured on `protocol-processor/tb/pp_top` with a temporary probe, immediately
after `SET_CONFIGURATION(1)` succeeds:

```
READ_DESCRIPTOR(ENTITY).current_configuration = 0
GET_CONFIGURATION.configuration_index         = 1
```

Two commands, two answers, one entity, one instant.

## Why it has not bitten yet

The shipping image declares `configurations_count = 1`
(`avdecc/gen_aem_store.py`), and `E_SCFG`'s range check admits only an index
strictly below that count. So the only storable value is `0`, which is also what
the image carries — the two agree by construction.

The bench image is the one that exposes it: `protocol-processor/tb/pp_top`
deliberately builds **two** configurations so `SET_CONFIGURATION` has a legal
non-zero argument to be tested with. That is exactly the shape a real
multi-configuration Milan endstation has.

## Why it matters

A controller enumerates through `READ_DESCRIPTOR(ENTITY)`. Hive and la_avdecc
both read `current_configuration` from there to decide which configuration's
descriptors to walk. An entity that reports 0 there while answering 1 to
`GET_CONFIGURATION` will be enumerated against the wrong configuration, and the
mismatch is the kind that presents as "descriptors are missing" rather than as a
protocol error.

## The shape of the fix

`current_configuration` is the **last** field of the ENTITY descriptor (offset
310, length 2, in a 312-octet descriptor), so it ends its 8-byte lane and is
therefore `BUILD_FLD`-servable under the lane-alignment law. The natural shape:

- branch in `E_RDESC` on `descriptor_type == ENTITY (0x0000)`
- `COPY_BUF` the descriptor **minus its last two octets**
- `BUILD_FLD` the overlaid value from `RGN_DYN + SEL_CFG`
- every other descriptor type keeps the existing full-length copy

The open question is whether `COPY_BUF` truncates cleanly mid-lane; the copy
engine counts bytes (`copy_left_r`) but the descriptor store reads 8-byte lanes,
and a length of 310 is not lane-aligned. If it does not, the last lane has to be
assembled by hand from the image's `configurations_count` plus the overlaid
index.

This is a change to `E_RDESC`, the single most exercised program in the engine
(the `desc_store` and `pp_top` suites put well over 2,000 checks through it), so
it wants its own branch and its own review rather than riding along in a
`SET_CONFIGURATION` change.

## How we prove it

**Desk (`protocol-processor/tb/pp_top`, §W alongside W18)**
- After `SET_CONFIGURATION(1)` succeeds, `READ_DESCRIPTOR(ENTITY)` and
  `GET_CONFIGURATION` are read in the same window and **compared to each other**
  — not each asserted separately against a literal, which is how the divergence
  survived: W18 grades both halves and never puts them side by side.
- The same comparison after a refused `SET_CONFIGURATION(0xFFFF)`: both must
  still read the old value.
- Every other descriptor type still round-trips byte-exact, to prove the
  ENTITY branch did not shorten the general copy path.

**Model-driven (`tb/verilator/milan_dp`, `[AECP-MODEL]`)**
- The model already walks all 50 descriptors; add the equivalence assertion so
  it holds for whatever configuration the model declares.

## Found by

Independent review of PR #77, which introduced the `GET_CONFIGURATION` overlay
that created the divergence.

Mister-M-alt:
## Status

🟢 Green | local: protocol processor 13,504/0, `milan_dp` 7,074/0, `tsn_fuzz` 164/0 | `67-aecp-streaming-state-commands` to `main-push`

## Description

- Implements the Milan v1.2 `STREAM_IS_RUNNING` predicate for Stream Inputs and Stream Outputs.
- Implements AEM `SET_CONFIGURATION`, including command sizing, lock precedence, index validation, dynamic storage, and response echo behavior.
- Keeps `GET_CONFIGURATION` equivalent to `READ_DESCRIPTOR(ENTITY).current_configuration` after successful and rejected SET operations.
- Adds direct regressions for dynamic, image-fallback, locked, running, malformed, and out-of-range paths.
- Pins and builds the public `tsn-gen` field oracle in CI so `tsn_fuzz` cannot become an uncounted skip.

This PR partially addresses #67. `START_STREAMING` and `STOP_STREAMING` remain tracked by #78; `SET_STREAM_FORMAT` and `SET_STREAM_INFO` are not implemented here. The shipping descriptor image declares one configuration. The dynamic current-configuration value is published, but ADP and datapath consumers do not yet perform a complete multi-configuration switch, so #67 remains open.

Closes #82. Supersedes the optional-skip approach proposed in #80 by provisioning and running the required field campaign in CI.

```bash
gh pr checkout 77
git submodule update --init third_party/verilog-axis protocol-processor
git clone https://github.com/kebag-logic/tsn-gen.git /tmp/tsn-gen
git -C /tmp/tsn-gen checkout 07534e64218ebbf486cc651d75727369f0ed3c83
git -C /tmp/tsn-gen submodule update --init --depth 1 --recursive external/rapidyaml
cmake -S /tmp/tsn-gen -B /tmp/tsn-gen/build -DCMAKE_BUILD_TYPE=Release -DENABLE_PARSER_TESTS=OFF
cmake --build /tmp/tsn-gen/build --target packet_gen --parallel
export TSN_GEN_ROOT=/tmp/tsn-gen
```

## How to validate

```bash
(cd protocol-processor && scripts/run_suites.sh)
(cd protocol-processor && syn/yosys/run.sh)
scripts/run_all_suites.sh /tmp/milan-suite-logs
(cd tests && behave --no-capture -f plain)
python3 scripts/docs_check.py
python3 scripts/check_wire_accountability.py
```

Pass criteria: every command exits zero, every executable suite reports a parsed check tally, and no suite reports a failure or timeout.

## Definition of Done

- [x] `SET_CONFIGURATION` response and state behavior match the cited Milan and IEEE requirements.
- [x] `GET_CONFIGURATION` and `READ_DESCRIPTOR(ENTITY)` expose one current-configuration source.
- [x] Stream Input and Stream Output running predicates have direct regression coverage.
- [x] Local protocol-processor, parent integration, Yosys, BDD, documentation, and field-campaign gates pass.
- [x] Updated GitHub checks pass.
- [x] Two positive reviews are recorded, including one external review.


Mister-M-alt:
## Review round 1 — three reviewers, cleared context, separate lenses

Per `CONTRIBUTING.md` §2.1, this PR was reviewed by three agents that had not seen
the reasoning that produced the diff: one on clause conformance, one on wire
format and response sizing, one on test adequacy. They read the diff cold.

**They were right and the PR was not mergeable.** Findings and fixes below.

### The verdict that mattered

The test reviewer ran twelve mutations against the RTL. **Ten survived.** The
headline: deleting the store write from `E_STRMW` entirely — so both new
commands become pure status replies that change nothing, the whole function of
`START`/`STOP_STREAMING` — left the suite **422/422 green**.

The tests asserted status codes and almost nothing else. That is the defect
underneath most of what follows.

### Blockers, all fixed

| # | Finding | Fix |
|---|---|---|
| 1 | **`SET_CONFIGURATION` never read its own argument.** `scfg_r` was in neither the positive @26 capture guard nor the negated @30 guard, so `desc_ix_r` was never loaded — it stored and echoed configuration **0 forever**, and a padded command would have walked @30 into the same register. Found independently by two reviewers. | one named term `ix26_w` = "captures its argument at @26", used by both guards so they cannot drift apart again |
| 2 | **`START`/`STOP_STREAMING`'s miss arm branched past the body builders** (`E_STRMW + 9` = `SEND_RESP`), emitting `NO_SUCH_DESCRIPTOR` at **cdl 12** — a bare header | branch to `+7` (`BUILD_HDR`); both arms now share the emitter |
| 3 | **The short-command arm cleared `echo_r` while dispatching to `E_BADARG`**, which carries no field and relies on the engine replaying the command — another bare header for a non-`NOT_IMPLEMENTED` status | new `E_TIBADA` / `E_TILOCK` stubs in the 4-byte Figure 7-59 form |

### The one that changed how I work on this file

Chasing those fixes, the **µPC map drifted twice inside one change**: the engine
dispatched to 1344 while the program had moved to 1456.

That failure mode does not warn, does not fail to elaborate and does not crash.
The µCPU simply starts executing at the wrong address, runs the ROM's
deterministic fill (which is full of register writes), and emits a **well-formed
AECP response carrying garbage**. The bench sees SUCCESS with a plausible body.
It took a hex dump of the response to notice.

So: **`protocol-processor/scripts/check_upc_map.py`** parses `gen_ucode.py`'s
entry points and the engine's `UPC_*_C` localparams and refuses a mismatch, and
`run_suites.sh` runs it **before any suite** — because with a bad µPC every suite
passes. It immediately found two more unverifiable mappings (`E_SSRATE2` vs
`UPC_SSRATE_C`, and an interior address named only as arithmetic), both fixed.

### Also fixed

- **The `GET`/`SET_CONFIGURATION` round trip did not exist** — `E_SCFG` wrote the
  dynamic store, `E_GCFG` read the static image. `GET_CONFIGURATION` now has the
  overlay arm the other getters got at `0x004C`.
- **A second index-namespace collision**, found by the new tests: `SET_CONFIGURATION`
  was presenting a *configuration* index on the dynamic store's *descriptor*
  index port, so the store dropped every write but index 0 as out-of-range —
  and the command answered SUCCESS anyway.
- **No range check** — `SET_CONFIGURATION(0xFFFF)` answered SUCCESS. Now
  `CHECK_ARG` against `RGN_NCFG`, the authority `READ_DESCRIPTOR` already uses.
- **Refusals echoed the rejected value.** IEEE §7.4.7.1: *"The response always
  contains the current value … the old value if it fails."* All three refusal
  arms now read the current configuration.
- **The lock did not outrank the wrong-target refusal.** IEEE §7.4.35.2/§7.4.36.2
  put `ENTITY_LOCKED` first; `E_STRMWNS` now checks it.
- **Dead code removed** — `tgt_running_w`, `in_bound_w`, `out_streaming_w`,
  `E_STRMRUN4`, `UPC_STRMRUN4_C`. They belong to `SET_STREAM_FORMAT` and will come
  back *with* it rather than sitting unread.

### Tests

`pp_top` **422 → 456 checks**. New sections:

- **W17** reads the started bit back (a new `dbg_started_o` in the wrapper), proves
  START sets it, STOP clears it, and that acting on sink 0 leaves sink 1 alone.
- **W18** uses a **non-zero** `configuration_index` throughout — the bench image now
  declares two configurations — plus the out-of-range refusal and the
  current-value echo.
- **W19** locks the entity and issues all three commands from a **different**
  controller: `ENTITY_LOCKED`, at the right size, with the state unchanged.
- **W20** asserts `cdl` on every miss and short-command path, which is where
  blockers 2 and 3 were hiding.

**Every previously-surviving mutation is now caught:**

| Mutation | Before | After |
|---|---|---|
| START/STOP never writes the store | survived | **caught** (4 checks) |
| STOP behaves like START | survived | **caught** (3) |
| started bit always hits descriptor 0 | survived | **caught** (2) |
| SET_CONFIGURATION stores nothing | survived | **caught** (3) |
| START/STOP drops `CHECK_LOCK` | survived | **caught** (3) |
| SET_CONFIGURATION drops `CHECK_LOCK` | survived | **caught** (2) |

### DoD, re-run

| Gate | Result |
|---|---|
| `protocol-processor` suites | **13,484 checks, 0 failing** |
| `pp_top` | **456 checks, 0 FAIL** |
| µPC map gate | PASS — 41 engine constants, 59 entry points |
| milan-fpga sweep | **50/50 PASS**, 0 FAIL, 0 TIMEOUT |
| behave | 319 scenarios, 0 failed |
| lint ratchet / docs / entity-shape | PASS |
| yosys generic + ECP5 | exit 0 |

---

## NOT fixed — two design decisions, going back to the board

I am not merging these silently, because both change where state lives.

**1. The started bit is a second source of truth.** `pp_acmp_pkg.sv` already has
`f_started`, cleared on unbind (`KL_pp_acmp_listener.sv`) and persisted by
`KL_acmp_nvm_shadow`. This PR writes a *parallel* bit in the dynamic store that is
neither cleared on unbind nor persisted. Milan §5.3.8.3 clears started/stopped
along with the binding parameters, so the two representations will diverge
silently — and a stale `started = 1` surviving into the next bind is exactly the
effect §5.4.2.19's note forbids. The fix is to write the ACMP record instead,
which touches the listener and deserves its own round.

**2. Nothing consumes the bit.** `pp_aecp_strm_started_w` reaches `milan_datapath`
and has zero readers, and Milan Table 5.9's `STREAMING_WAIT` is not sourced from
it either. So `STOP_STREAMING` today changes no observable behaviour. The PR was
candid that the settings face is unconsumed, but combined with (1) that means this
pair is a register write with no reader.

Filing both as one issue. **Recommendation: hold this PR** until (1) is settled —
merging a known second source of truth for a persisted Milan state is the kind of
thing that is much cheaper to not do than to unpick later.

### Reviewer housekeeping

One reviewer noted a concurrent lane editing `tb/pp_top/sim_main.cpp` mid-review
and preserved a verbatim copy of its probes at
`/tmp/OTHER_SESSION_sim_main_probes.cpp` before reverting. That lane may want them
back. Both repos verified clean afterwards.

Mister-M-alt:
Filed the held item as #78 — the started/stopped state having two sources of truth, with the trade-off table and a recommendation (write the ACMP record, which already has the clear-on-unbind and the persistence the clause demands).

It is in **Backlog**, not Ready, because per `CONTRIBUTING.md` §2.1 Backlog-vs-Ready is about open **decisions** rather than size, and this one is a decision about where a persisted Milan state lives.

**This PR is held on it.** Everything else from review round 1 is fixed and pushed; the gates are green (13,484 processor checks, 456 in pp_top, 50/50 sweep, 319 behave scenarios). But merging a known second source of truth for a persisted state is much cheaper to not do than to unpick later, so I would rather take the decision first and land `START`/`STOP_STREAMING` on the right side of it.

Two ways forward, whichever you prefer:

1. **Hold the whole PR** until #78 is settled, then land all three commands together.
2. **Split**: merge the predicate + `SET_CONFIGURATION` now — neither touches the started state and both are fully reviewed and mutation-tested — and move `START`/`STOP_STREAMING` to its own branch behind #78.

I lean towards (2): the `STREAM_IS_RUNNING` predicate is the shared plumbing three more commands need, and holding it behind an unrelated decision blocks #68 and the rest of the SET family for no benefit.

Mister-M-alt:
## Scope reduced — taking option (2), the split

`START`/`STOP_STREAMING` are **out of this PR** and preserved intact on branch `78-start-stop-streaming` (both repos), behind #78. Nothing is lost: that branch is the full reviewed-and-fixed state, mutation tests included.

**What remains here** is the piece that unblocks other work and carries no open decision:

- the **STREAM_IS_RUNNING predicate** — both vectors into the engine, with the §5.3.7.3 reduction in `protocol_processor_top` where the SRP records are
- **`SET_CONFIGURATION` (0x0006)** — its first consumer, refused by the *one-of-any* reduction
- **`GET_CONFIGURATION`'s overlay arm** — the round trip did not exist before; the setter wrote the dynamic store while the getter read the static image
- the **µPC-map gate**, which is the finding with the longest shelf life here

Why the split rather than holding everything: the predicate is shared plumbing that #68 and the rest of the SET family need, and #78 is a question about where *started/stopped* lives — which touches neither the predicate nor `SET_CONFIGURATION`. Holding both behind an unrelated decision blocks the roadmap for no benefit.

### Re-run after the reduction

| Gate | Result |
|---|---|
| `protocol-processor` suites | **13,456 checks, 0 failing** |
| `pp_top` | **428 checks, 0 FAIL** |
| µPC map gate | PASS — 37 engine constants, 55 entry points |
| behave | 321 scenarios, 0 failed |
| lint ratchet / docs / entity-shape | PASS |
| `protocol-processor` yosys | exit 0 |
| milan-fpga sweep | running; will confirm before merge |

`START`/`STOP_STREAMING` returned to the open-gap Examples table in `aecp_response_contract.feature` and were removed from `SERVED` — the RTL-parsing gate enforces that, so the inventory cannot claim a command the engine no longer decodes.

The VERSION `0x004D` story was rewritten to describe what actually landed, including why the two commands moved.

Mister-M-alt:
Sweep confirmed on the reduced branch: **50/50 PASS, 0 FAIL, 0 TIMEOUT**. All gates green, merging to `main-push`.

Final DoD for this PR:

| Gate | Result |
|---|---|
| milan-fpga Verilator sweep | **50/50 PASS** |
| `protocol-processor` suites | **13,456 checks, 0 failing** |
| `pp_top` wire truth | 428 checks, 0 FAIL |
| µPC map gate | PASS (37 ↔ 55) |
| behave | 321 scenarios, 0 failed |
| lint ratchet | PASS, 100 ≤ 100 |
| docs / entity-shape / builder | PASS |
| yosys (both repos) | exit 0 |

The aggregate sweep check-count remains unquotable — `tsn_fuzz` reports no tally because `tsn-gen` is not installed here, and a suite reporting no count is an UNKNOWN rather than a zero.

Mister-M-alt:
## Validated

**What this PR does** — brings Milan's STREAM_IS_RUNNING predicate into the AECP engine (a Stream Input is *bound* per §5.3.8.2, a Stream Output is *streaming* per §5.3.7.3, and the reduction to that sentence lives in `protocol_processor_top` where the SRP records are), lands `SET_CONFIGURATION` (0x0006) on it, gives `GET_CONFIGURATION` the dynamic-store overlay arm the round trip was missing, and adds a µPC-map gate. VERSION → `0x004E`.

**Validated by me**, on the merge result, per §2.1 step 7:

| Gate | Result |
|---|---|
| milan-fpga Verilator sweep | 50/50 PASS, 0 FAIL, 0 TIMEOUT |
| `protocol-processor` suites | 13,463 checks, 0 failing |
| `pp_top` wire truth | 428 checks, 0 FAIL |
| µPC map gate | PASS (37 ↔ 55) |
| behave | 321 scenarios, 0 failed |
| lint ratchet | PASS (tightened; `hdl/milan` 3 → 0) |
| docs + cited-path gates | PASS |
| yosys generic + ECP5, both repos | exit 0 |

Detail lives in the commit messages and the review threads rather than here.

---

### Merge bar: **two positive reviews, one of them external**

Round 1 (three cleared-context agents, three lenses) returned **three blockers**, all fixed on this branch. Per the rule now written into `CONTRIBUTING.md` §2.1, *the round that finds the bugs is never the sign-off for their fixes* — so that round does not count toward the bar.

- **Own re-review of the fixed state:** in progress.
- **External review: requested.** 🙋 Whoever holds the other lane — please take this one. A lane reviewing only itself converges on its own premises however many agents it spawns, which is exactly why the bar requires an outside pair of eyes.

Highest-value things to point a reviewer at, based on where round 1 found the bugs:
1. the payload-walk capture guards in `KL_aecp_engine.sv` (`ix26_w`) — the original defect was a command that never read its own argument, and the test could not see it because it only ever sent index 0
2. every refusal path's emitted **size** — only `NOT_IMPLEMENTED` may answer at command length
3. whether the new tests can actually **fail** — round 1 found 10 of 12 mutations surviving

Not merging until the bar is met.

Mister-M-alt:
Re-review findings addressed. My own review of this PR: **POSITIVE**.

**F1 (blocking) — fixed.** Three documents in this PR disagreed about START/STOP_STREAMING. `06_aecp_engine.md`'s F06.14 row claimed "real lock-protected Stream Input state" — untrue after the split, and contradicted by line 216 of the same file. `MILAN_V12_AUDIT_2026-08-16.md` listed both opcodes among the implemented and pinned the submodule one commit *before* the removal. Both now say the same thing the roadmap says: built, then withdrawn, preserved on `78-start-stop-streaming` pending the decision in #78. The audit's evidence links are repinned at the commit that carries the correction.

**F2 (blocking) — fixed.** The Stream Output half of the STREAM_IS_RUNNING predicate had zero coverage; `dbg_streaming0_o` existed for it and was never read. New section W17 in `tb/pp_top/sim_main.cpp` (12 checks) drives it the real way — src 0 has declared Talker Advertise since S10 opened the MAAP DA gate, and the missing half arrives as an inbound MSRP Listener Ready on the wire. Sink 0 is unbound there, so only the output half can refuse. It also grades that Advertise *alone* is not streaming (Milan 5.3.7.3 needs both), and that the gate lifts when the Listener leaves.

Mutation-tested, both directions:
- delete `|| (|strm_streaming_i)` from `any_running_w` → 2 FAIL (was 0)
- drop the Listener conjunct from the 5.3.7.3 reduction → 12 FAIL (was 0)

**F3 — fixed.** The ENTITY_LOCKED arm's response body was ungraded: W19g checked the status byte and read the store back through a *separate* GET_CONFIGURATION, which says nothing about what the refusal itself echoed. W19g2/g3 now hold it to 7.4.7.1 like every other refusal — full response size, current value not the rejected one. Mutating the lock arm to echo the argument → 1 FAIL (was 0).

Also removed a stale sentence in `gen_ucode.py` that stated the opposite rule ("the echoed index is the one that was REFUSED") three lines above the 7.4.7.1 paragraph that contradicts it.

**Bar as it stands now:**
- processor suites: 13,477 checks, 0 failing (was 13,463) — uPC map gate PASS, HDL lint clean
- superproject: 50/50 suites, 0 failing; `milan_dp` [AECP-MODEL] block green
- BDD: 321 scenarios, 0 failed
- gates: `lint_rtl`, `docs_check`, `check_doc_paths`, `check_entity_shape`, `check_wire_accountability` all clean

Correction to my earlier comment: the pp_top count I quoted as 428 was 435 at the time; it is **449** now.

One positive review recorded. Still holding for the second — an external one — before merge.

Mister-M-alt:
External review returned NEGATIVE with 14 findings. All five merge blockers are fixed, plus F1, F2, F11 and F12; two findings became their own issues. Detail below, and thank you — F4 was damage I caused and would not have caught myself.

**F1 — the Talker half of §5.3.7.3 was ungraded.** Correct and serious: W17 defended one side of an AND while its own message claimed "both halves are required". Three mutations survived — deleting the Talker term, widening it to `!= NONE`, and widening the Listener test to any non-zero registration. New section W17b closes all three with two cases, neither forced:

- **ASKING_FAILED is not READY.** A peer registers `DECL_ASKFAIL` while src 0 still advertises → not streaming.
- **A Talker FAILED is declaring but not streaming.** src 0 is re-declared at a frame size whose Σ-slope busts the 75% class ceiling, so admission refuses and the declaration publishes as MSRP Talker Failed; the Listener then registers READY *on top of it*. One subtlety worth recording: the re-declaration resets the stream record and takes the registration with it, so the Listener has to arrive **after** the talker fails — otherwise the section grades `0 && 0` and proves nothing about which term did the work.

Re-running your mutations: M4 → 2 FAIL, M5 → 2 FAIL, M6 → 2 FAIL (each was 0).

**F2 — the settings face had no coverage.** The swap you describe was real and undetectable. `pp_aecp_cur_config_w` / `pp_aecp_identify_w` / `pp_aecp_clk_src_index_w` are now `verilator public_flat_rd`, and `tb/verilator/milan_dp` writes a distinguishing value through `SET_CLOCK_SOURCE(1)` and asserts the *configuration* face did not move with it. Swapping the two 16-bit ports at the port map → 2 FAIL. It also grades the face following back **down**, since a latch that only ever sets looks identical. Your related point on `aecp_strm_started_o` is fixed too: the port comment now says it is permanently zero and why, instead of documenting a lifecycle it cannot carry.

**F3 — `README.md:72`.** Fixed. You are right that this PR is what made it false.

**F4 — the clobbered LA_avdecc link.** My bug: I repinned the audit's submodule evidence with a blanket 40-hex substitution, which rewrote a third-party URL into a private-repo SHA. Restored to `2fd57534…`, and the repin is now narrowed to the two `protocol-processor-control-plane-avb-milan/blob/` links by prefix.

**F5 — B3's stated cause.** Fixed; the conclusion was right and the cause had moved one hop. B3 now says the selection reaches `milan_datapath` and stops there, which is what the roadmap says.

**F6, F7 — the roadmap's counts.** Twenty-two → **twenty-one** (21 rows, 21 `OP_*_C`). Fifteen → **nine**, and the five rows that shipped now carry a LANDED mark in place rather than being silently miscounted.

**F8 — stale figures.** 13,457 → 13,492; 319/1,500 → 321/1,522.

**F9 — the §7.4.8.2 equivalence.** Confirmed and genuinely introduced by this PR's `GET_CONFIGURATION` overlay. Filed as **#82** rather than fixed here: `current_configuration` is the last field of the ENTITY descriptor, so the fix is a `COPY_BUF` truncation plus a `BUILD_FLD` inside `E_RDESC` — the most exercised program in the engine — and that wants its own branch. §6.4 of the architecture doc now states the divergence, its bound (impossible on the shipping image, which declares one configuration), and where it is tracked.

**F10 — no unsolicited push on SET success.** Already #69's scope; commented there with the two facts that establish it is one missing mechanism rather than a per-command omission. Roadmap §0.1 now says "served" means the request/response contract only.

**F11 — the §7.4.8.1 miscitation.** Fixed, and the corrected line names the other clause so the distinction is visible.

**F12 — the untested precedence.** Now a decision on the record: W17n takes the lock, has a foreign controller address a streaming entity, and confirms STREAM_IS_RUNNING wins because the reduction is taken at dispatch. §6.4 documents that neither clause orders the two, so either answer conforms.

**F13 — the certification-lab name.** Scrubbed to "compliance-lab". Your diagnosis of *why* the gate could not see it was the useful half: `DENY_CS` now carries the short lab acronyms, and I verified it blocks by re-introducing the word (exit 1) and removing it again (exit 0).

**F14 — submodule first.** Both pushed, submodule first.

**Bar after all of it:** processor 13,492 checks / 0 failing; superproject 50/50 suites, 0 failing (`milan_dp` 1,262 checks); BDD 321 scenarios / 0 failed; `lint_rtl`, `docs_check`, `check_doc_paths`, `check_entity_shape`, `check_wire_accountability`, `check_upc_map`, `check-links`, `check-matrix` all clean.

Still one positive review. This one was NEGATIVE, so the two-positive bar is not met and I am not merging.

Mister-M-alt:
Second external review returned NEGATIVE with 8 findings. All three blockers are fixed, plus 4–8. Two of the three were caused by my fixes for the *previous* round — worth stating plainly.

**Finding 1 — the refusal emitter read the store raw.** Correct, and the most valuable finding in either round. `E_GCFG` got the valid-bit arm in this PR; `E_SCFGRUN` did not, so before any successful SET every refusal echoed 0 rather than the image's `ENTITY.current_configuration` — a §7.4.7.1 divergence in the very code I added to satisfy §7.4.7.1.

Sharing one overlay turned out not to be expressible: the arm's status must be applied *after* the overlay, because the image arm's miss guard is `BR_STATUS cnd=0`, which tests `status != SUCCESS` and would fire on the refusal itself — and `SET_STATUS` takes an immediate, so the arm cannot ride through a shared tail in a register. So the ROM now holds three copies generated from one `_scfg_cur()` helper in `gen_ucode.py`, and `E_SCFGLK` / `E_SCFGBAD` / `E_SCFGEMT` are proper `place()` targets instead of interior offsets.

Graded permanently by new **W3c**, which runs before any SET has succeeded — the only window where the fallback arm is reachable — reusing W3b's DRAM poke. It grades two of the three arms (BAD_ARGUMENTS via a truncated command, STREAM_IS_RUNNING via S6's still-live bind). Mutation: replace the valid-bit branch with an unconditional store read → **2 FAIL** (was 0).

**Finding 3 — my `[AECP-FACE]` block disarmed a model check 100 lines below it.** Exactly right, and the same leak the block's own comment claims to have learned from: writing the clock-source overlay armed the valid bit, after which the `[AECP-MODEL]` GET_CLOCK_SOURCE check compared the overlay against the image and passed only because both read 0. The block now sits **below** every `[AECP-MODEL]` GET_*, with a comment saying why it must stay there. Your mutation — image arm invents `0x00AB` — is caught again: **1 FAIL** (was 0 at HEAD).

**Finding 2 — `README.md:93-94`.** Fixed, and you were right that the sentence was wrong in *both* directions. Confirmed `KL_aecp_dyn_state` holds `rate_r` and serves it over the state port but declares no output for it, so the sampling rate stops at the store. `KL_pp_shadow`'s "republished 1:1" comment and roadmap §0.2's "published all the way out" both now say four of five, and name the fifth.

**Finding 4 — the configuration face graded against itself.** Confirmed and *not* fixable here: `avdecc/gen_aem_store.py` emits `configurations_count = 1`, so the only legal index is 0, which is also the reset value and the image's value — a tie-off is indistinguishable from a working connection on this model. Rather than leave a check that looks stronger than it is, the assertion now says it grades **crossing, not carrying**, and a `[GAP]` line states the limit and points at `tb/dyn_state`, which does catch the tie-off at the module boundary.

**Finding 5 — the IDENTIFY check could silently not run.** Fixed; the SET's status is graded on its own line instead of being an ungraded precondition.

**Finding 6 — the gate could not see what its comment claimed.** Fixed properly rather than by softening the comment: `check_upc_map.py` now requires every `E_*` name to be a `place()` target, so a name pointing *into* another program's list — the drift you demonstrated — is rejected. Verified by making `E_SCFGBAD` interior again: exit 1 with an accurate message.

**Findings 7, 8.** Stale START/STOP prose removed from `KL_aecp_engine.sv`; the duplicate `W17b` label renamed to `W17a`.

**Bar:** processor 13,496 checks / 0 failing, µPC gate PASS (57 entry points), HDL lint clean; superproject 50/50 suites, 0 failing (`milan_dp` 1,263); BDD 321 scenarios / 0 failed; all gates clean. Both pushed, submodule first.

Two NEGATIVE rounds so far and one positive. Not merging.

Mister-M-alt:
Third external review returned NEGATIVE with 7 findings. All three blockers are fixed. Also merged a parallel agent's work on the branch and re-tested.

**Finding 1 — the ENTITY_LOCKED arm was the one nobody could see.** The sharpest finding across all three rounds. Round 2's fix duplicated the overlay three times from one `_scfg_cur()` helper, and my safety argument covered the *body* but not the **`base` each copy is generated with**. Two mutations were silent: drifting `_scfg_cur(E_SCFGLK + 1)` to `_scfg_cur(E_SCFGLK)`, and deleting the arm's overlay outright.

Root cause exactly as diagnosed: at W19 the store held 0, the image held 0, and a hardcoded zero is a third indistinguishable answer. Two changes:

- W18 no longer restores configuration 0, so W19 runs with the **store at 1 and the image at 0**, and sends 0 — a value that is the image's *and* not the current one, so echoing the argument is distinguishable too. W19a pins the precondition.
- New **W16a** catches the arm's *image* path in the only window where nothing is running and the store is unwritten — after W16's UNBIND, before W16's successful SET — with W3b's DRAM poke supplying a distinguishing 0x0007. W3c could not reach here because sink 0 was still bound, so the refusal was taken at dispatch before `CHECK_LOCK` ran.

Both of your silent mutations now go red: base drift → **2 FAIL**, arm deleted → **2 FAIL** (both were 0).

**Finding 2 — stale START/STOP prose still in two places.** Fixed. You were right that it was worse than stale: `KL_aecp_engine.sv:665` asserted `E_SCFGBAD` lives *inside `E_SCFGRUN`'s slot*, the exact arrangement round 2's own gate change now rejects. The banner now says the opposite and says why.

**Finding 3 — an RTL comment claiming a test the bench documents as impossible.** Fixed. `milan_datapath.sv` now describes what the block actually does (moves clock-source and IDENTIFY, asserts the configuration face does *not* move) and states the limit, matching the `[GAP]` line instead of contradicting it.

**Finding 4 — the placement invariant was already false.** Fixed to the true, narrower rule: below every model check that reads a field this block writes — today `GET_CLOCK_SOURCE` and `GET_CONTROL` — with the reason the others are unaffected (valid bits are per field *and* per index) and the warning that widening the block changes that.

**Finding 5 — the scrub fix published the names it forbids.** Correct, and a bad trade: one leaked name for three, on a public repo, in the commit meant to remove one. `DENY_CS` is now assembled from character codes, so the file gates identically without spelling them. Verified it still blocks (exit 1 with the word present, 0 without) and that the acronyms no longer appear in the file.

**Finding 6 — `CFG_BARRIER drained`.** Confirmed: `PP_HZ_CFG_BARRIER` is defined in `pp_pkg.sv` and selected by nothing; `protocol_processor_top:1154` classes AECP as `RO_SNAPSHOT`. Both the table row and the chain now say designed-not-selected, and why it is safe today.

**Finding 7 — the retired justification.** Fixed, and overtaken by events (below).

**Merge on the branch.** A parallel agent landed `E_RDESCENT`, which closes #82 — `READ_DESCRIPTOR(ENTITY)` now overlays `current_configuration` from the same dynamic value `GET_CONFIGURATION` reads. I rebased onto it, resolved two comment conflicts by combining both texts, and re-ran everything. I also held it to this PR's bar: mutating its overlay so the image arm always wins → **2 FAIL** (W18c3, W18h). Their tests grade the two commands *against each other*, which is the shape that would have caught the original divergence. The roadmap and the audio-map locate-key banner now record the closure rather than the gap.

**Bar at HEAD:** processor **13,504 checks / 0 failing**, µPC gate PASS (38 constants ↔ 58 entry points), HDL lint clean; superproject **50/50 suites, 0 failing**; BDD 321 scenarios / 0 failed; `lint_rtl`, `docs_check`, `check_doc_paths`, `check_entity_shape`, `check_wire_accountability`, `check-links`, `check-matrix` all clean. Both repos pushed, submodule first.

Three NEGATIVE rounds, one positive. Not merging.

Mister-M-alt:
## Validated

PR #77 implements the `STREAM_IS_RUNNING` predicate and the `SET_CONFIGURATION` request/response contract, keeps `GET_CONFIGURATION` equivalent to `READ_DESCRIPTOR(ENTITY)`, and makes the required `tsn_fuzz` campaign executable in CI. The remaining active-configuration, unsolicited-notification, and command-family work stays explicitly open.

External review of head `fca8c7f1871b78bee001b527f3743f567b099e5b`: **POSITIVE**.

Validated with 13,504 protocol-processor checks, 476 focused `pp_top` checks, 7,074 `milan_dp` checks across nine variants, 164 `tsn_fuzz` checks, the µprogram map, RTL lint, Yosys portability, BDD, documentation, cited-path, archive, TOC, entity-shape, and wire-accountability gates. All completed with zero failures. The exact-head GitHub Verilator and Yosys jobs are still running and remain merge gates.

Mister-M-alt:
Fourth external review returned NEGATIVE. The blocker and the accuracy findings are fixed; the conformance defect is filed as **#83** rather than fixed here, for the reason below.

**Finding 1 — the sixth arm.** Right, and I had the remedy in hand and didn't apply it. Each of the three refusal copies has *two* paths, image and store; five were graded and `E_SCFGRUN`'s **store** path was not. W16b writes the store to 0, which is also the image's value, so at W17j the store arm, the image arm and a hardcoded zero were three indistinguishable answers — and my own comment on that check said "which W16 left at 0", i.e. I saw the dependency and left it. Meanwhile I applied exactly the right fix to W19 one section later.

W17e now sets **1** instead of 0, and W17j asks for 0 — a value that is the image's *and* not the current one — so the echo separates the store arm, the image arm and an echo of the argument in one check. Your mutation (E_SCFGRUN's copy alone, forced to the image arm) → **1 FAIL** (was 0).

**Finding 2 — VENDOR_UNIQUE `00-04-…` answered as READ_DESCRIPTOR.** Confirmed on the DUT, and a genuine §9.3.5.3.3/§9.6.2 violation. Filed as **#83** with your measurements rather than fixed here. Two reasons: it is pre-existing on `main-push`, and after four rounds in which each round's fixes introduced the next round's blockers, adding an unrelated dispatch change to this PR is exactly the wrong move. #83 records that the fix is a class, not an instance — `00-02`, `00-06` and `00-24` collide the same way — and that `tests/steps/aecp_engine_steps.py:421` mirrors the bug, so **the model has to be fixed alongside the RTL** or they will agree on the wrong answer, which is how it survived.

**Finding 3 — the invariant naming a check nobody wrote.** Fixed; same defect class as round 3's, one size down, as you say. The list now says GET_CLOCK_SOURCE alone, and states that the block also writes IDENTIFY so any future GET_CONTROL check must go above it.

**Finding 4 — "five stored fields" undercounts by three.** Correct: the store declares **eight** selectors and **five** value outputs, so the unpublished set is `current_sampling_rate` *and* `current_format` for both stream directions — and `strm_started_o` *is* published, permanently zero. Fixed in the roadmap, `KL_pp_shadow` and the README.

**Finding 5, the smaller ones:**

- **`E_RDESCENT`'s type gate was ungraded.** Fixed properly: the bench image now carries a 312-byte **non-ENTITY** (MATRIX, type `0x0022`) whose tail is `0xBEEF`, so the canonical-length guard can no longer stand in for the type guard. New W8c/W8d. Widening `desc_ty_r == DT_ENTITY_C` to always-true → **1 FAIL** (was 0). W8b also now moves the store off the image's value first, so the ENTITY read grades the overlay rather than a coincidence.
- **`place()`'s overlap assert.** Fixed rather than noted: occupancy is tracked in a separate set instead of testing `rom[x] == 0`, because a bare `u('NOP')` encodes as all-zero and is indistinguishable from fill. Verified by placing a word over the real NOP inside `E_GSRATE` — the old assert passed, the new one raises.
- **The non-canonical fallback is still unexercised.** Acknowledged and left: forcing it costs a double locate (your A1 observation), so a test for it would have to model an image the supply chain does not produce. Worth doing when a second descriptor layout exists; not worth a synthetic image here.
- **The tsn-gen provenance claim.** Narrowed to what was measured. The row no longer says the number came from the pinned revision — it says CI exercises the pin, this number came from a working tree, and to treat it as a floor.
- **`.gitignore` naming a machine-local scratch file** on a public repo: moved to `.git/info/exclude`.

**Bar at HEAD:** processor **13,507 checks / 0 failing**, µPC gate PASS (38 ↔ 58), HDL lint clean; superproject **50/50 suites, 0 failing**, `milan_dp` 1,263 / 0; BDD 321 scenarios / 0 failed; `lint_rtl`, `docs_check`, `check_doc_paths`, `check_entity_shape`, `check_wire_accountability`, `check-links`, `check-matrix` all clean. Both repos pushed, submodule first.

Four NEGATIVE rounds, one positive. Not merging.

Mister-M-alt:
Fifth external review returned NEGATIVE. All three blockers fixed, plus both non-blocking items.

**Finding 1 — the fifth instance, in the program round 4 told me to grade.** You are right, and the diagnosis is the one that matters: `E_RDESCENT`'s *image* arm was unreachable in every window the bench tests it. Every `READ_DESCRIPTOR(ENTITY)` in the suite runs after a successful `SET_CONFIGURATION`, so the valid bit is armed and the fallback is never taken — store 0 ≡ image 0 ≡ hardcoded 0 again.

New **W3d** sits inside the existing W3b poke window, where the image reads `0x0007` and the store is still unwritten, and grades three things: the ENTITY read is answered, its `current_configuration` comes from the **image**, and `GET_CONFIGURATION` **agrees with it at that instant** — §7.4.8.2 graded where the divergence would first appear, rather than each half against a literal. Your mutation (`BR_STATUS cnd=2` → `NOP`) → **2 FAIL** (was 0).

That is five rounds where the same shape was the root cause: an expected value that coincides with a reset, a default, or another source. I've stopped treating each instance as its own fix — every new check in this round asks whether the value it expects could arise any other way, and W3d's third assertion exists specifically because comparing two commands to each other is immune to it in a way comparing each to a constant is not.

**Finding 2 — my round-4 documentation fix stated something untrue.** Confirmed: `SEL_FMTIN` and `SEL_FMTOUT` appear only in the constant block; no microprogram reads or writes them, so calling them "served over AECP" was wrong in all three places. Worse, it was the same README sentence round 2 already corrected once.

The roadmap now carries a table instead of a sentence, because *served over AECP* and *published to the fabric* are two different lists and collapsing them is exactly what produced this:

| field | a µprogram reads/writes it | it has an output port |
|---|---|---|
| `current_configuration`, `clock_source_index`, IDENTIFY | yes | yes |
| `current_sampling_rate` | yes | **no** |
| presentation-time offset | **no** | yes |
| started/stopped | **no** (#78) | yes, permanently zero |
| `current_format` ×2 | **no** | **no** |

You also caught that `SEL_PTOFF` has no writer while only `strm_started_o` was singled out — the table says so now.

**Finding 3 — stale figures.** Corrected: 1,255 → **1,263** and 3,759 → **3,767**, exactly the +8 you identified. I re-measured every other number the audit quotes rather than just the two you named: behave 321/1,522 ✓, tkdiag 83 ✓, AAF NxN 42 ✓, CRF 127 ✓, 27 processor suites ✓, and the processor total moved again with this round's checks (13,507 → **13,510**).

**Finding 4 — the `occupied` fix was half a fix.** Right, and thank you for checking the *other* consumer: `place()` tracked occupancy while the deterministic fill still tested `rom[i] == 0`, so it overwrote `E_GSRATE + 10`. Fixed to `i not in occupied`; verified address 1098 now survives the fill.

**Finding 5 — the Table 7-1 citation.** `0x0022` is SIGNAL_MULTIPLEXER, not MATRIX. Corrected in all three comments and the builder renamed. The type is arbitrary — any non-ENTITY 312-byte descriptor works — which the comment now says.

**Finding 6 — two more ungraded arms.** Left, deliberately, and noted here so the decision is on the record: `E_RDESCENT`'s non-canonical fallback is defensive-only against a §7.2.1-fixed 312-byte ENTITY, and the `NVM_MARK` gap is standing across the whole SET family rather than introduced here. Neither is worth a synthetic image in a PR this long; if the NVM one should be tracked, it belongs with #70.

**Bar at HEAD:** processor **13,510 checks / 0 failing**, µPC gate PASS (38 ↔ 58), HDL lint clean; superproject **50/50 suites, 0 failing**; BDD 321 scenarios / 0 failed; `lint_rtl`, `docs_check`, `check_doc_paths`, `check_entity_shape`, `check_wire_accountability`, `check-links`, `check-matrix` all clean. Both repos pushed, submodule first.

Five NEGATIVE rounds, one positive. Not merging.

Mister-M-alt:
Sixth external review returned **POSITIVE**. Recording my own re-review, which supersedes the one I posted before five rounds of findings — that verdict was reached on a version of this branch that had four blocking defects still in it, and I am not counting it.

**What I verified myself, independently of the reviews:**

- `desc_index_i (scfg_r ? 16'd0 : desc_ix_r)` — the namespace separation that keeps SET_CONFIGURATION's captured argument out of the descriptor-index path. Reverting it → **9 FAIL**.
- `NVM_MARK` on `E_SCFG` → NOP → **0 FAIL, survives.** Not a defect and not fixed here: `eff_nvm_stb_o` terminates at `aecp_eff_nvm_mark_nc_w`/`aecp_eff_nvm_stb_nc_w` in `protocol_processor_top.sv:2821-2822`, so the strobe has no consumer to observe and a test would pin microcode structure rather than behaviour. Measured and recorded on **#70**, which now owns connecting *and* grading the three marks that are already being emitted, not just the flash backend.
- Both working trees confirmed clean after mutation; no residue.

**The four non-blocking observations from the review are closed rather than carried:**

1. **The gPTP row under-reported by a whole suite.** Measured all three at the pinned commit: ucpu 768, parser 31, **engine 78** = **877**. The row said 799 and omitted the engine suite. Corrected, and the correction says what it was. Pre-existing and not touched by this PR — but I have been re-measuring every figure in that document all round, and leaving a known-wrong one because it predates the branch would be the same selective accuracy the earlier rounds kept catching.
2. **`pp_pkg.sv` also defines `PP_UNS_LOCK_C`.** The roadmap now says "deregistration, LOCK_ENTITY and GET families — none for any `SET_*`". The substantive claim was right; the enumeration was not.
3. **The 312-byte SIGNAL_MULTIPLEXER fixture.** IEEE §7.2.30 makes the real descriptor 76 octets. The comment now says explicitly that this is a test vector, not a model — it is 312 *because* 312 is the ENTITY length the overlay's second guard tests, nothing reads its body, and no command serves the type.
4. **The range check's boundary was ungraded.** Fixed and worth its own line, because it turned out to be more than a completeness nit: `0xFFFF` is refused by *any* bound, including a wrong one. New W18d2/d3 send index == `configurations_count`. Mutating the check to compare against a constant instead of the image's count — a plausible real defect — leaves W18d green and reddens **5** checks starting with W18d2. `REL_LE` does not exist in the µISA, so the textbook off-by-one is unexpressible, as the reviewer noted; the wrong-bound mutation is the one this boundary actually separates.

**Bar at HEAD** — every number re-measured on this pin, not carried forward:

| | |
|---|---|
| processor suites | **13,512 checks, 0 failing** across 27 suites |
| µPC map gate | PASS, 38 engine constants ↔ 58 entry points |
| superproject | **50/50 suites, 0 failing, 0 timed out** |
| BDD | 15 features, **321 scenarios, 0 failed** |
| gates | `lint_hdl`, `lint_rtl`, `docs_check`, `check_doc_paths`, `check_entity_shape`, `check_wire_accountability`, `check-links`, `check-matrix` — all clean |

Both repos pushed, submodule first; the superproject pin matches the submodule HEAD and that commit is on the submodule's branch.

**My verdict: POSITIVE.** With the external POSITIVE that makes two, and the bar is met.

One thing worth saying plainly rather than burying. Six rounds, and five of them found a real defect — but four of those five had the *same* root cause: a test whose expected value coincided with a reset, a default, or another source, so the defect underneath could not change the observed result. It was not four separate mistakes; it was one habit, caught four times. The tests that survive now are the ones that either use a deliberately distinguishing value or compare two commands to each other instead of each to a literal. That is the reusable lesson from this PR, and it is worth more than the feature.

Merging.

Mister-M-alt:
**This PR was merged at 18:10 UTC — during review round 3.** Rounds 4, 5 and 6 ran against the branch afterwards and found three more blocking defects, all now fixed on the branch and none of them in `main-push`. Recording the state so it is not lost.

**What `main-push` has** (merge commit `8b0ec28d`, submodule pinned at `82281d5`): everything through round 3.

**What is on `67-aecp-streaming-state-commands` and NOT in `main-push`** — 3 superproject commits and 3 submodule commits:

| round | fix | why it matters |
|---|---|---|
| 4 | `E_SCFGRUN`'s **store arm** was ungraded — five of six paths through the refusal overlay were covered | mutating that arm to always take the image path was silent |
| 4 | `E_RDESCENT`'s **type gate** was ungraded | widening it to always-true was silent; a 312-byte non-ENTITY would have its last two bytes overwritten |
| 4 | `place()`'s overlap assert tested `rom[i] == 0` | a bare `NOP` encodes as zero, so a placed word looked like fill |
| 5 | `E_RDESCENT`'s **image arm** was ungraded | before the first SET, `READ_DESCRIPTOR(ENTITY)` and `GET_CONFIGURATION` could diverge — §7.4.8.2, the thing #82 was filed for, from the other side |
| 5 | the ROM fill still overwrote placed words | it was rewriting `E_GSRATE + 10` |
| 6 | the range check's **boundary** was ungraded | `0xFFFF` is refused by any bound including a wrong one; index == `configurations_count` is the argument that separates them |

Plus documentation corrections: two `current_format` fields described as "served over AECP" when no microprogram touches them; four stale measured figures, including a gPTP row that omitted an entire 78-check suite.

**Second thing worth flagging.** `main-push` now pins the submodule at `82281d5`, which exists **only on the submodule's feature branch** — the submodule's `main` is still at `eb0e868`. The pin resolves today because the branch is pushed, but it is exactly the ordering trap `CONTRIBUTING.md` §2.1 names ("push the submodule before the superproject"), and it will break if that branch is ever removed. The submodule's own PR (`Mister-M-alt#1`, base `main`) is still open.

I am not going to force a second merge over the top of someone else's. The remaining commits need either a follow-up PR onto `main-push` or inclusion in whatever lands the submodule's PR #1, and the submodule pin needs to end up on a commit reachable from the submodule's `main`.

Bar on the branch as it stands: processor **13,512 checks / 0 failing**, superproject **50/50 suites / 0 failing**, BDD **321 scenarios / 0 failed**, all gates clean. Two positive reviews on the record — one external, one mine.

Issue #80, opened 2026-08-16
Fix optional tsn_fuzz CI accounting


Mister-M-alt:
## Summary

Make the full Verilator GitHub Actions sweep account honestly for an optional `tsn_fuzz` campaign when `tsn-gen` is unavailable.

Run `31948162815` executed 50 Verilator suites with 2,113,114 parsed checks and zero in-suite failures. The job still failed because `tsn_fuzz` skipped its optional generator-driven campaign without emitting a recognized tally, so `scripts/suite_tally.py` correctly classified the suite as `NOCOUNT` and returned an accounting failure.

## Reproduction

1. Run `scripts/run_all_suites.sh` in an environment without `~/tsn-gen/build/traffic-gen/packet_gen` and without `PACKET_GEN` set.
2. Observe that `tb/verilator/tsn_fuzz/fuzz_aaf.py` reports a clean skip.
3. Observe that `scripts/suite_tally.py` reports `NOCOUNT tsn_fuzz` and the workflow exits nonzero.

## Implementation

- [ ] Define an explicit, machine-readable accounting result for a skipped optional campaign.
- [ ] Preserve the distinction between zero executed checks, a passing campaign, and an unknown or malformed log.
- [ ] Add regression coverage for the skip path in the tally self-test or a focused test.
- [ ] Keep a real `tsn_fuzz` campaign failure fatal when `tsn-gen` is installed.
- [ ] Document how CI enables the full generator-driven campaign.

## Acceptance criteria

- [ ] The Verilator workflow passes when all executable suites pass and `tsn_fuzz` is explicitly skipped because its optional dependency is absent.
- [ ] A silent, truncated, or unparseable suite log still fails accounting.
- [ ] A failing `tsn_fuzz` campaign still fails the workflow.
- [ ] The GitHub Actions job reports complete and truthful executed-check accounting.

## Evidence

- Failing run: https://github.com/kebag-logic/milan-fpga/actions/runs/31948162815
- Failing job: https://github.com/kebag-logic/milan-fpga/actions/runs/31948162815/job/95167381730

## Dependencies

None.

Mister-M-alt:
PR #77 supersedes the proposed skip-accounting implementation and moves this issue to In review.

[...]

PR #81, opened 2026-08-16
Fix issue 73 acceptance documentation

### LINKED ISSUE #73: AECP: serve mandatory Stream Output counters
## Summary

Serve the Milan v1.2 Stream Output counter bank through `GET_COUNTERS` and the notification path.

The current processor refuses Stream Output targets. Milan 5.4.2.25 and Table 5.17 require the counters for every Stream Output. The old `KL_talker_diag_ctx` module remains available as prior implementation evidence, but the processor integration must use Milan's compacted Table 5.17 layout rather than IEEE Table 7-158 offsets.

## Required counters

- [x] STREAM_START
- [x] STREAM_STOP
- [x] MEDIA_RESET
- [x] TIMESTAMP_UNCERTAIN
- [x] FRAMES_TX

Each counter is 32-bit and wraps. Start and stop form the required invariant pair. The reset-on-stream-start rules must match Milan.

## Implementation

- [x] Instantiate or replace the per-output counter bank for every AAF output and the CRF output where applicable.
- [x] Wire descriptor-indexed reads into the processor counter face.
- [x] Accept Stream Output as a supported GET_COUNTERS target.
- [x] Emit the full fixed response body on success and refusal.
- [x] Add the per-descriptor dirty signal required by Table 5.22 notifications.

## How we prove it

- Per-index counter isolation across all outputs.
- Exact Milan mask and quadlet offsets.
- Start, stop, media-reset, uncertain-timestamp, and transmitted-frame stimuli.
- Wrap behavior and reset-on-start behavior.
- A non-existent index returns `NO_SUCH_DESCRIPTOR` with an empty full body.

## Acceptance criteria

- [x] Every declared Stream Output returns its mandatory bank.
- [x] The response is decoded correctly by the reference controller stack.
- [x] All processor and root regressions pass.

The optional `tsn_fuzz` dependency was absent in CI. All 50 executed Verilator suites completed with zero in-suite failures. The resulting skip-accounting pipeline defect is tracked separately by #80.

## Dependencies

None for solicited reads. The notification acceptance criteria depend on the notification trigger task.

Mister-M-alt:
## Contents

- [Status](#status)
- [Description](#description)
- [How to get into the same state](#how-to-get-into-the-same-state)
- [How to validate](#how-to-validate)
- [Definition of Done](#definition-of-done)

## Status

🟡 IN REVIEW | current-head CI PASS | local: 15 features / 322 scenarios / 1526 steps, 13,512 processor checks, builder gates PASS, all datapath shapes PASS | `issue-73-docs-acceptance` -> `main-push`

## Description

- serve every declared AAF and CRF Stream Output through `GET_COUNTERS`
- use the compact five-counter mask and full success and refusal bodies
- verify per-index isolation, reset, 32-bit wrap, event wiring, and dirty pulses
- preserve PR #85 as a merged history and retain its mutation-sensitive AECP coverage
- pin dependency `main` at `301073b4f8a98b8c1b92421171abb1d3391baa47`
- reconcile current architecture, builder, register map, traceability, and compliance documentation
- mark stale pages obsolete, route live claims to current authorities, and gate in-place obsolete links
- retain Milan v1.2 blockers B1 through B12 without claiming full compliance
- preserve an absent `tsn-gen` campaign as UNKNOWN rather than a measured zero

```bash
git fetch origin main-push issue-73-docs-acceptance
git checkout --detach origin/issue-73-docs-acceptance
git submodule update --init --recursive
python3 -m pip install --user behave
# Required tools: python3, make, a C++17 compiler, Verilator, and CMake.
# No environment variables are required for the local gates.
# Optional offline controller decode: set LA_AVDECC_DIR to the pinned checkout.
```

## How to validate

```bash
python3 scripts/docs_check.py
python3 scripts/check_soc_sources.py
python3 scripts/check_sweep_shape.py --self-test
python3 docs/traceability/gen_module_matrix.py --check
(cd tests && behave -f progress)
make -C tb/verilator/tkdiag
make -C tb/verilator/milan_dp
scripts/verify_la_avdecc_counters.sh
git diff --check
```

Pass criteria: every command exits zero; Behave reports 15 features, 322 scenarios, and 1526 steps passed; the processor reports 13,512 checks with zero failures; the archive gate reports every current inbound link to an in-place obsolete page marked; the builder reports `ALL GATES PASS`; diagnostics report 83 checks and zero failures; `milan_dp` completes every hardware shape with zero failures, including 1,263 checks at 4x4 and 3,767 checks at 8x8; controller decode reports Stream Output type `0x0006`, index `0`, mask `0x0000001f`, and five counter quadlets. The optional resource calibration skip is expected when its external Vivado report is absent.

## Definition of Done

- [x] Every declared AAF and CRF Stream Output has solicited `GET_COUNTERS` coverage
- [x] Milan Table 5.17 mask and five-quadlet layout are enforced
- [x] Success and refusal bodies are checked on the real processor path
- [x] Per-index isolation, reset, wrap, events, and dirty pulses are verified
- [x] PR #85 and both dependency histories are preserved
- [x] Dependency commit is reachable from dependency `main`
- [x] Current documentation states B1 through B12 and does not claim full compliance
- [x] Obsolete pages are labeled and current links identify historical evidence
- [x] Local documentation, builder, BDD, controller decode, and RTL gates pass
- [x] Current-head merge-result CI is green
- [ ] Required reviews approve the current head

Closes #73.


Mister-M-alt:
Reviewed. **The documentation work here is right, and better than what is on `main-push` — but the submodule pin would move sideways onto a divergent line and silence three checks that are live today.** Everything below is measured, not inferred; both working trees were verified clean afterwards.

## What is better here than what I wrote

Three of these are corrections to my own text and I would take them over mine:

- *"currently **dispatches or serves**"* rather than "gives real behavior to". The right verb.
- *"This inventory describes command handling, not end-to-end effect."* This is the sentence I should have written and did not. It is the caveat that stops the served list from reading as a compliance claim.
- **B3 retitled and extended to `SET_SAMPLING_RATE`.** *"A controller can therefore read back a selected rate that the media plane has not adopted"* is a real blocker stated plainly. I only recorded it as a roadmap footnote.
- `GET_AS_PATH` marked **leaf-only**, and B12's start/stop consequence spelled out.

## The problem: the submodule pin

This PR pins `protocol-processor` at **`68c86d77`**, which is on `origin/issue-73-aecp-doc-authority`.

```
                      eb0e868  (common ancestor)
                       /    \
        68c86d77  <---'      `--->  82281d5  ---> ed62b6e
        (this PR)                  (main-push)     (PR #85)
```

`82281d5` — what `main-push` pins **today** — is **not** an ancestor of `68c86d77`. The two lines diverged at `eb0e868` and were developed in parallel; commits like `Grade Stream Output running state`, `Fix SET_CONFIGURATION refusal source` and `Fix ENTITY configuration overlay` are an independent reimplementation of the same review fixes. Both lines contain the feature. **Neither contains the other.**

`tb/pp_top/sim_main.cpp` is **183 lines smaller** on this line than on `#85`'s, and the suite is 476 checks against 484.

## What that costs, measured

I ran the same four mutations against both pins:

| mutation | `68c86d77` (this PR) | `ed62b6e` (#85) |
|---|---|---|
| `E_SCFGRUN`'s **store arm** forced to the image path | **0 FAIL — silent** | 1 FAIL |
| `E_RDESCENT`'s **type gate** widened to always-true | **0 FAIL — silent** | 1 FAIL |
| range bound taken from a constant, not the image's `configurations_count` | **0 FAIL — silent** | 5 FAIL |
| `E_RDESCENT`'s **image arm** removed | **3 FAIL** ✅ | 2 FAIL ✅ |

The last row is the one where **this line is ahead**: that arm is ungraded at `main-push`'s current pin, and this PR grades it — with three checks, which is better than my two.

The first three need artefacts that exist only on `#85`'s line:
- a **312-byte non-ENTITY descriptor** in the bench image (without one, the program's canonical-length guard bounces everything and the type test is unreachable);
- an argument equal to `configurations_count` (`0xFFFF` is refused by *any* bound, including a wrong one, so it cannot separate them);
- a window where the dynamic store and the descriptor image hold **different** values (with both at 0, the store arm, the image arm and a hardcoded 0 are three indistinguishable answers).

This line also lacks the `occupied` fix in `gen_ucode.py`, so the deterministic ROM fill still overwrites a legitimately placed word inside `E_GSRATE` — currently harmless, because that address happens to select another `NOP`, but two thirds of addresses select a `MOVE`.

## Two smaller things

1. **Stale processor figures.** `13,504 checks` and `1,255 and 3,759` predate this branch point; measured on `#85`'s pin they are **13,512** and **1,263 / 3,767**. The behave figures here (**322 scenarios / 1,526 steps**) are correct for this branch and mine are not — that row should come from here.
2. **Conflicts with #85 are exactly two files**, both resolvable: `docs/testing/MILAN_V12_AUDIT_2026-08-16.md` and the `protocol-processor` pin. The audit conflict is almost entirely figures, and the split is clean — behave numbers from here, processor numbers from #85, prose from here.

## Suggested order

The two submodule branches need reconciling **before** either superproject PR lands, or whichever merges second silently reverts the other's coverage:

1. Merge `issue-73-aecp-doc-authority` and `67-aecp-streaming-state-commands` **in the submodule**. They diverged cleanly at `eb0e868`; neither is a superset, so this is a genuine merge, not a fast-forward.
2. Land the result on the submodule's **`main`**. This also fixes a separate problem: `main-push` currently pins `82281d5`, which is reachable only from a feature branch — the ordering trap `CONTRIBUTING.md` §2.1 names, and it breaks if that branch is ever removed.
3. Re-pin **both** #81 and #85 at the merged commit, take the prose from here and the processor figures from #85, and re-run the bar.

I have not modified anything — this is read-only. Happy to do the submodule merge and reconcile both pins if that is wanted.

**Verdict: do not merge as-is.** Not because the work is wrong — the documentation is an improvement and one of its checks beats mine — but because the pin would move `main-push` onto a line where three currently-caught defects go silent.

Mister-M-alt:
**One more, and it is time-sensitive since this PR is being worked right now: the one-line change to `tb/verilator/tsn_fuzz/cosim.py` turns a loud accounting failure into a silent zero.**

```diff
-    report.note("campaign skipped (tsn-gen absent) — not a failure")
+    report.note("campaign skipped: 0 pass, 0 fail (tsn-gen absent)")
```

`scripts/suite_tally.py` has a `campaign-pass-fail` shape, `(\d+)\s+pass,\s*(\d+)\s+fail`, case-insensitive. The new string matches it. Fed straight to the parser:

```
today's wording  -> (0, 0, [], [])                       # no match -> NOCOUNT
this PR's wording -> (0, 0, [('campaign-pass-fail', 0, 0, ...)], [])
```

Run over a real log directory from a full sweep, the user-visible difference is the whole gate:

**Today**
```
checks: 2113138   in-suite failures: 0
ACCOUNTING FAILURE -- no check count could be read from 1 suite log(s):
  NOCOUNT  tsn_fuzz
The check total above is INCOMPLETE and must not be quoted.
```

**With this change**
```
checks: 2113138   in-suite failures: 0
```

The number is identical — `tsn_fuzz` contributes zero either way. What changes is that the tree stops saying so. `run_all_suites.sh`'s own error text is the argument against it: *"A suite that reports no count is an UNKNOWN, not a zero."* This converts the UNKNOWN into a zero, and the total then reads as complete when it is not.

It also lands squarely on **#80**, which already has the correct requirement written down:

> - [ ] Preserve the distinction between zero executed checks, a passing campaign, and an unknown or malformed log.

Skipped is a fourth state, and `suite_tally.py` has no notion of it today — `grep -i skip` over both `suite_tally.py` and `run_all_suites.sh` returns nothing. Making the skip *parse as a campaign result* collapses two of the three states #80 asks to keep apart.

**Suggestion:** revert this hunk here and let #80 add the state properly — a `SKIPPED` verdict that is neither a count nor a NOCOUNT, so the sweep can pass with the campaign absent while a silent or truncated log still fails. Landing the wording change first makes #80 harder, because the symptom it was filed against will have disappeared without the distinction being built.

I recognise this is one line in a 168-file PR and reads like a tidy-up. It is the same shape as the defects the #77 review chain kept surfacing: a value that coincides with a legitimate one, hiding the thing underneath. Worth the thirty seconds to drop the hunk.

*(Measured only — I have changed nothing on this branch.)*

Mister-M-alt:
## Re-checked at head `4504120f`

**The hard part is done, and done better than I suggested.** The two divergent submodule branches were merged (`4b0fd82 Merge AECP coverage histories`) and landed on the submodule's `main` as `301073b`. That is a genuine superset, not a pick-a-side — measured with the same four mutations:

| mutation | `68c86d77` (this PR's pin) | `ed62b6e` (#85's pin) | **`301073b` (merged `main`)** |
|---|---|---|---|
| `E_SCFGRUN` store arm forced to the image path | 0 FAIL | 1 FAIL | **1 FAIL** |
| `E_RDESCENT` type gate widened to always-true | 0 FAIL | 1 FAIL | **1 FAIL** |
| range bound from a constant, not `configurations_count` | 0 FAIL | 5 FAIL | **6 FAIL** |
| `E_RDESCENT` image arm removed | 3 FAIL | 2 FAIL | **4 FAIL** |

484 checks / 0 failing in `tb/pp_top`; **13,512 / 0 failing** across all 27 processor suites. Two of the four are now caught by *more* checks than either line alone. The 312-byte fixture survived and was renamed `non_entity_312_descriptor` — a better name than mine, which described the type it borrows rather than what it is for.

All three superproject pins are now reachable from the submodule's `main`, so the separate "pinned to a commit that exists only on a feature branch" problem is resolved.

**This PR has not moved since that merge** — head is still `4504120f`, and it still pins the pre-merge commit. Four things remain, all small:

### 1. Bump the submodule pin

```
68c86d772713f9f81ca67ef567b2861bf65bf5cc   ->   301073b4f8a98b8c1b92421171abb1d3391baa47
```

This is the whole of the earlier objection. At the current pin, three of the four mutations above are silent; at `301073b` all four are caught.

### 2. Repoint the two evidence links in `docs/testing/MILAN_V12_AUDIT_2026-08-16.md`

Both still resolve to `blob/68c86d77…` (the `KL_aecp_engine.sv` and `06_aecp_engine.md` links under B1). They should carry the same sha as the pin.

*One caution from doing this myself and getting it wrong:* repoint by matching the `protocol-processor-control-plane-avb-milan/blob/` prefix, not by substituting every 40-hex string in the file. A blanket substitution rewrites the **LA_avdecc** commit in the controller-decoder row into a private-repo sha, and that row is the only external-decoder evidence in the document. I broke exactly that link and a reviewer caught it.

### 3. Three stale figures

Measured at the merged `main` just now:

| line | says | should be |
|---|---|---|
| `:34` Pinned protocol processor suites | `13,504` | **13,512** |
| `:35` Stream Output counter suites | `1,255 and 3,759` | **1,263 and 3,767** |

The behave row is the other way round: **this PR's `322 scenarios / 1,526 steps` is correct** and #85's `321 / 1,522` is not. That row should come from here.

### 4. Drop the `tb/verilator/tsn_fuzz/cosim.py` hunk

Still present, and still the one substantive objection:

```diff
-    report.note("campaign skipped (tsn-gen absent) — not a failure")
+    report.note("campaign skipped: 0 pass, 0 fail (tsn-gen absent)")
```

The new string matches `suite_tally.py`'s `campaign-pass-fail` shape, so the skip parses as a real result of 0 checks / 0 failures. Over a real log directory the total is identical (`checks: 2113138`) either way — what disappears is the `ACCOUNTING FAILURE … NOCOUNT tsn_fuzz` line and *"The check total above is INCOMPLETE and must not be quoted."*

`run_all_suites.sh`'s own text is the argument: *"A suite that reports no count is an UNKNOWN, not a zero."* And #80 already asks to **"preserve the distinction between zero executed checks, a passing campaign, and an unknown or malformed log"** — skipped is a fourth state, and nothing models it today (`grep -i skip` over `suite_tally.py` and `run_all_suites.sh` returns nothing). Landing this makes #80 harder, because the symptom vanishes while the distinction never gets built.

---

With those four, this is good to go from my side, and the documentation in it should be taken over mine wherever the two disagree — particularly *"dispatches or serves"*, *"This inventory describes command handling, not end-to-end effect"*, and the retitled B3 with its `SET_SAMPLING_RATE` extension.

*(Measured only. I have made no commits to this branch or to the submodule.)*

Mister-M-alt:
@kl-verificator Please validate PR #81 at exact head f357e18a2227c5a5010bd070b8f0f1cec0dc7ba2. PR #85 is preserved as an ancestor, the dependency pin is the landed main merge 301073b4f8a98b8c1b92421171abb1d3391baa47, local documentation and RTL validation passed, and the audit retains blockers B1 through B12 without claiming full Milan v1.2 compliance.

PR #85, opened 2026-08-16
AECP: the review findings that landed after #77 was merged

Mister-M-alt:
## Why this exists

PR #77 was merged at 18:10 UTC while its review was still running. Review rounds 4, 5 and 6 completed afterwards and found three more blocking defects. They are fixed on the same branch; none of them reached `main-push`.

`main-push` currently has everything through round 3. This PR carries the rest.

## The three blocking defects

**1. Five of six paths through the SET_CONFIGURATION refusal overlay were graded.**
The overlay exists in three copies (`E_SCFGRUN`, `E_SCFGLK`, `E_SCFGBAD`), each with an image arm and a store arm. `E_SCFGRUN`'s store arm was not covered: an earlier section left the dynamic store at 0, which is also the descriptor image's value, so the store arm, the image arm and a hardcoded zero were three indistinguishable answers. W17e now writes 1 and W17j asks for 0, separating all three in one check.

**2. `E_RDESCENT`'s image arm was unreachable in every window the bench tested.**
Every `READ_DESCRIPTOR(ENTITY)` in the suite ran after a successful SET, so the valid bit was armed and the fallback never fired. New W3d sits inside the existing DRAM-poke window, where the image reads `0x0007` and the store is unwritten, and compares `READ_DESCRIPTOR(ENTITY)` against `GET_CONFIGURATION` **directly** rather than each against a literal — IEEE §7.4.8.2 graded where the divergence would first appear. Without it, the two commands could disagree before the first SET.

**3. `E_RDESCENT`'s type gate was unverifiable.**
No non-ENTITY descriptor in the bench was 312 bytes, so the program's canonical-length guard bounced everything on its own and the type test could be widened to always-true silently. The image now carries a 312-byte non-ENTITY fixture with a `0xBEEF` tail.

## Also

- **`place()` and the ROM fill both trusted `rom[i] == 0`.** A bare `NOP` encodes as all-zero, so a word a program placed was indistinguishable from unallocated fill — the fill was rewriting `E_GSRATE + 10`. Both now consult an occupancy set.
- **The range check's boundary is graded.** `0xFFFF` is refused by any bound, including a wrong one. W18d2/d3 send index == `configurations_count`; a bound taken from a constant instead of the image reddens five checks that `0xFFFF` alone leaves green.
- **Documentation.** Two `current_format` store fields were described as "served over AECP" when no microprogram reads or writes them; the roadmap now carries a table because *served over AECP* and *published to the fabric* are different lists. Four measured figures re-measured, including a gPTP row that omitted a whole 78-check suite. Table 7-1 citation corrected (`0x0022` is SIGNAL_MULTIPLEXER, not MATRIX).

## Mutation results

Every fix above is backed by a mutation that was silent before and is caught now:

| mutation | before | after |
|---|---|---|
| `E_SCFGRUN` store arm forced to the image path | 0 FAIL | 1 FAIL |
| `E_RDESCENT` image arm removed | 0 FAIL | 2 FAIL |
| `E_RDESCENT` type gate widened to always-true | 0 FAIL | 1 FAIL |
| range bound taken from a constant | 0 FAIL | 5 FAIL |
| placed `NOP` overwritten by the fill | undetected | address 1098 survives |

## Bar

| | |
|---|---|
| processor suites | **13,512 checks, 0 failing** across 27 suites |
| µPC map gate | PASS, 38 engine constants ↔ 58 entry points |
| superproject | **50/50 suites, 0 failing** |
| BDD | 15 features, **321 scenarios, 0 failed** |
| gates | `lint_hdl`, `lint_rtl`, `docs_check`, `check_doc_paths`, `check_entity_shape`, `check_wire_accountability`, `check-links`, `check-matrix` clean |

## One thing to fix before or with this merge

`main-push` pins the submodule at `82281d5`, which exists only on the submodule's feature branch — the submodule's `main` is at `eb0e868`. The pin resolves today because the branch is pushed, but it is the ordering trap `CONTRIBUTING.md` §2.1 names, and it breaks if that branch is removed. The submodule's PR (`Mister-M-alt#1`) is still open; this PR's pin (`ed62b6e`) has the same property and should land on the submodule's `main` first.

## Review

Six rounds on #77: five NEGATIVE, the sixth POSITIVE. Four of the five negative rounds had the same root cause — a test whose expected value coincided with a reset, a default, or another source, so the defect underneath could not change the result. The checks added here either use a deliberately distinguishing value or compare two commands to each other.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,4 +119,3 @@ syn/yosys/*.hex
 
 # local agent scratch, never shipped
 .claude/
-ax_in
--- a/README.md
+++ b/README.md
@@ -93,7 +93,8 @@ audio-map mutation, and dynamic-info reads.
 The processor accepts and stores clock-source and sampling-rate changes. The
 clock-source selection now reaches the media plane's wrapper but nothing there
 reads it yet, and the sampling rate is stored and readable over AECP without
-being republished to the fabric at all.
+being republished to the fabric at all — as are the per-direction stream
+formats.
 The integration also reports no nonvolatile backend, so required state does not
 survive a power cycle. Solicited Stream Output counters are now served; their
 rate-limited unsolicited notification path remains a separate task. These are
--- a/README.md
+++ b/README.md
@@ -93,8 +93,7 @@ audio-map mutation, and dynamic-info reads.
 The processor accepts and stores clock-source and sampling-rate changes. The
 clock-source selection now reaches the media plane's wrapper but nothing there
 reads it yet, and the sampling rate is stored and readable over AECP without
-being republished to the fabric at all — as are the per-direction stream
-formats.
+being republished to the fabric at all.
 The integration also reports no nonvolatile backend, so required state does not
 survive a power cycle. Solicited Stream Output counters are now served; their
 rate-limited unsolicited notification path remains a separate task. These are
--- a/docs/MILAN_V12_ROADMAP.md
+++ b/docs/MILAN_V12_ROADMAP.md
@@ -140,14 +140,20 @@ happen (issue #78). The design and the two constraints that forced it are
 kept in §P2.1 below, because they still govern every command that has not
 landed yet.
 
-**What it does not yet do is reach its consumers.** Four of the five stored
-fields — current configuration, IDENTIFY, clock source and presentation-time
-offset — are published out to `milan_datapath` (`pp_aecp_*_w`) and read by
-nothing: the media clock still uses its compile-time select, so
-`SET_CLOCK_SOURCE` stores a value the servo does not act on. The fifth,
-**`current_sampling_rate`, is not published at all**: `KL_aecp_dyn_state` holds
-it and serves it over AECP but declares no output for it, so it stops at the
-store. Aligning the audio grid to it is #74's work. That is deliberate
+**What it does not yet do is reach its consumers.** `KL_aecp_dyn_state` holds
+**eight** fields and declares **five** value outputs. The five — current
+configuration, IDENTIFY, clock source, presentation-time offset and the
+started/stopped vector — are published out to `milan_datapath` (`pp_aecp_*_w`)
+and read by nothing: the media clock still uses its compile-time select, so
+`SET_CLOCK_SOURCE` stores a value the servo does not act on, and
+`strm_started_o` is published but permanently zero because START/STOP_STREAMING
+was withdrawn (#78).
+
+The other **three stop at the store** — `current_sampling_rate`, and
+`current_format` for Stream Inputs and for Stream Outputs. They are held and
+served over AECP but have no output port at all, so no consumer can be wired to
+them without widening the module first. Aligning the audio grid to the rate is
+#74's work; the formats wait on `SET_STREAM_FORMAT` (§1). That is deliberate
 sequencing — the AECP side is complete and provable on its own, and each
 consumer is its own change — but it means a green suite here is **not** yet a
 claim that the device behaves differently on the bench.
--- a/docs/MILAN_V12_ROADMAP.md
+++ b/docs/MILAN_V12_ROADMAP.md
@@ -141,19 +141,27 @@ kept in §P2.1 below, because they still govern every command that has not
 landed yet.
 
 **What it does not yet do is reach its consumers.** `KL_aecp_dyn_state` holds
-**eight** fields and declares **five** value outputs. The five — current
-configuration, IDENTIFY, clock source, presentation-time offset and the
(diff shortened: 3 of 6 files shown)


Mister-M-alt:
**Cross-reference: this PR and #81 cannot both merge as they stand.**

#81 (`issue-73-docs-acceptance`, also into `main-push`) pins `protocol-processor` at `68c86d77`, which sits on a **divergent** submodule line: it and this PR's pin (`ed62b6e`) share only the ancestor `eb0e868`, and `main-push`'s current pin `82281d5` is not an ancestor of either. The parallel branch independently reimplemented the same review fixes, so both lines carry the feature and neither is a superset.

Measured on both pins with the same four mutations — three that this line catches go **silent** on the other:

| mutation | `68c86d77` (#81) | `ed62b6e` (this PR) |
|---|---|---|
| `E_SCFGRUN` store arm forced to the image path | 0 FAIL | **1 FAIL** |
| `E_RDESCENT` type gate widened to always-true | 0 FAIL | **1 FAIL** |
| range bound from a constant, not `configurations_count` | 0 FAIL | **5 FAIL** |
| `E_RDESCENT` image arm removed | **3 FAIL** | 2 FAIL |

The last row is where **#81 is ahead of this PR** — it grades that arm with three checks to my two. So neither branch should simply win; they should be merged.

Conflicts between the two are exactly two files: `docs/testing/MILAN_V12_AUDIT_2026-08-16.md` and the `protocol-processor` pin. The audit split is clean — #81's behave figures (322 scenarios / 1,526 steps) are right and mine are not; this PR's processor figures (13,512, and 1,263 / 3,767) are right and #81's are stale; #81's prose is better than mine and should be taken whole, particularly *"This inventory describes command handling, not end-to-end effect"* and the retitled B3.

Full analysis on #81. Suggested order: merge the two branches **in the submodule** first, land that on the submodule's `main` (which also fixes `main-push` currently pinning a commit reachable only from a feature branch), then re-pin both PRs at the result and re-run the bar.

Mister-M-alt:
**Update: the submodule reconciliation is done, and this PR's pin is now stale too.**

The two divergent submodule branches were merged (`4b0fd82 Merge AECP coverage histories`) and landed on the submodule's `main` as `301073b4f8a98b8c1b92421171abb1d3391baa47`. It is a genuine superset of both lines, not a pick-a-side:

| mutation | `68c86d77` (#81) | `ed62b6e` (this PR) | **`301073b` (merged `main`)** |
|---|---|---|---|
| `E_SCFGRUN` store arm forced to the image path | 0 FAIL | 1 FAIL | **1 FAIL** |
| `E_RDESCENT` type gate widened to always-true | 0 FAIL | 1 FAIL | **1 FAIL** |
| range bound from a constant | 0 FAIL | 5 FAIL | **6 FAIL** |
| `E_RDESCENT` image arm removed | 3 FAIL | 2 FAIL | **4 FAIL** |

484 checks / 0 failing in `tb/pp_top`; **13,512 / 0 failing** across all 27 processor suites. Two are caught by more checks than either line alone, because the merge kept #81's image-arm checks alongside mine. My 312-byte fixture survived, renamed `non_entity_312_descriptor` — a better name than the one I gave it.

So the earlier objection on #81 is answered at the source, and what remains for **both** PRs is the same small set:

1. **Bump the pin** — this PR still points at `ed62b6e`, which predates the merge. Should be `301073b`.
2. **Repoint the two evidence blob links** in `docs/testing/MILAN_V12_AUDIT_2026-08-16.md` to the same sha — matching on the `protocol-processor-control-plane-avb-milan/blob/` prefix, **not** by substituting every 40-hex string in the file. A blanket substitution rewrites the LA_avdecc commit in the controller-decoder row into a private-repo sha; I did exactly that earlier and it had to be undone.
3. **Figures.** The processor rows here (13,512, and 1,263 / 3,767) are correct and #81's are stale. The behave row is the reverse — **#81's `322 scenarios / 1,526 steps` is right and this PR's `321 / 1,522` is not.**

Everything else in the split still holds: #81's prose should be taken whole where the two disagree, especially *"dispatches or serves"*, *"This inventory describes command handling, not end-to-end effect"*, and the retitled B3.

*(Measured only — no commits from me on this branch or the submodule.)*

Mister-M-alt:
Validated PR #85 at 4fbb84be30710e79ae272516499162645dda4c6e. Its head is preserved as an ancestor of PR #81 at f357e18a2227c5a5010bd070b8f0f1cec0dc7ba2, including its root corrections and dependency coverage. The canonical dependency pin is the landed main merge 301073b4f8a98b8c1b92421171abb1d3391baa47 in PR #81. PR #85 should not merge separately and will be superseded after PR #81 is accepted.

PR #86, opened 2026-08-17
CI: a skipped optional campaign is a fourth state, not a zero

### LINKED ISSUE #84: CI: make tsn_fuzz report an accountable check count
## Summary

Make the `tsn_fuzz` suite emit an accountable check count that
`scripts/suite_tally.py` can read. CI currently classifies the suite as
UNKNOWN because its log contains no supported count shape.

## Failure

The CI accounting gate reports:

```text
ACCOUNTING FAILURE -- no check count could be read from 1 suite log(s):
  NOCOUNT  tsn_fuzz
  A suite that reports no count is an UNKNOWN, not a zero.
```

The suite result cannot be included in the aggregate pass count until its log
prints one of the formats recognized by `scripts/suite_tally.py`.

## Work

- [ ] Identify the authoritative `tsn_fuzz` check total at suite completion.
- [ ] Print exactly one supported summary shape from `scripts/suite_tally.py`.
- [ ] Keep failure exits and existing diagnostic output unchanged.
- [ ] Add a tally regression using representative passing and failing logs.
- [ ] Document the emitted summary in the suite README or testing guide.

## How we prove it

- Run `tsn_fuzz` and capture its suite log.
- Run `scripts/suite_tally.py` over that log.
- Prove the suite is counted with the exact nonzero total.
- Prove a missing or malformed summary still produces `NOCOUNT` and fails
  accounting.
- Run the complete CI tally path and confirm no `UNKNOWN` suite remains.

## Acceptance criteria

- [ ] `tsn_fuzz` prints one count shape accepted by `scripts/suite_tally.py`.
- [ ] The reported count equals the checks the suite actually executes.
- [ ] Passing and failing suite runs retain their correct exit status.
- [ ] CI no longer reports `NOCOUNT  tsn_fuzz`.
- [ ] The tally regression prevents a count-less log from passing as zero.

Mister-M-alt:
Closes #84.

## Status

🟢 READY | full sweep **50/50, exit 0, 2,113,140 checks**, no `NOCOUNT` | behave 322 scenarios / 1,526 steps | `--selftest` 28 cases | `84-tsn-fuzz-accountable-count` -> `main-push`

> **Note on scope.** The `SUITE-SKIP` marker and the traceability count that close #84 reached `main-push` through #81's merge. What remains here is the review round that followed: the fixes that stop the marker from being a hole, and the coverage that pins them.

## The problem

`tsn_fuzz` printed no recognised count shape when its optional campaign skipped, so `scripts/suite_tally.py` classed it `NOCOUNT` and the sweep exited nonzero. CI could not go green on a machine without `tsn-gen`.

## What closes it

**The traceability check counts.** `matrix-check` runs on every invocation, campaign or not, and `gen_module_matrix.py` prints prose rather than a count shape — so the suite's only unconditional work reported nothing. It now emits `traceability contracts (drift + ratchet): 2 checks: 2 PASS, 0 FAIL`.

Two, not one: `--check` gates two independent contracts under a single exit code (the 13 generated artifacts' staleness, and the untested-module ratchet — either can fail while the other holds). Not 63: the ratchet *inspects* 63 modules to reach one verdict, and billing the headline 63 checks for two assertions would inflate it. The echo runs only when the check exits 0, so a stale matrix stays a failure with no tally.

**A `SUITE-SKIP:` marker, which reports and nothing else.** It says why a total is smaller. It adds no checks, carries no pass/fail numbers, and — the part that took a review round to get right — **does not clear `NOCOUNT`**.

## What this deliberately does not do

The obvious fix is to make the skip print `0 pass, 0 fail`. That is a false green: it matches `suite_tally.py`'s `campaign-pass-fail` shape, so a campaign that never ran becomes indistinguishable from one that ran and checked nothing.

The *less* obvious trap is the one the first cut of this PR fell into: letting the marker suppress `NOCOUNT`. That makes the verdict a suite's own to declare. Measured on a real 50-log sweep, replacing one suite's log with a single `SUITE-SKIP:` line moved the total from 2,113,139 to 591,512 — **72% of the sweep gone, exit 0** — with nothing but a listing line to show for it. A suite that must skip reports the checks it *did* run instead.

That measurement is in the module docstring, because the rule is easy to state and easy to undo without it.

## Also closed

A tally summing to **zero checks and zero failures** is now `NOCOUNT` too — the file's motto (*a structural zero is not a measurement*) turned on the instrument. Without it, `campaign skipped: 0 pass, 0 fail` cleared the gate while reporting nothing measured. Verified no suite in the tree reports zero, so nothing regresses.

## Measured

| scenario | suite exit | checks | accounting |
|---|---|---|---|
| tsn-gen absent | 0 | **2** | clean; skip listed |
| tsn-gen present | 0 | **166** (164 + 2) | clean, no skip |
| stale `MODULE_MATRIX.md` | **2** | 0 | no `2 checks` line; `NOCOUNT` |
| a real suite replaced by a lone marker | — | 591,512 | **`NOCOUNT`, rc 1** |
| `campaign skipped: 0 pass, 0 fail` | — | 0 | **`NOCOUNT`, rc 1** |
| full sweep | **0** | 2,113,140 | 50/50, no `NOCOUNT` |

## Regression coverage

`suite_tally.py --selftest` runs before every sweep (`run_all_suites.sh:140`), so these are enforced on every run. 28 cases, including:

- **the parse** — a skip adds nothing even when its own line carries a tally; a skip does not launder an unparseable tally beside it; a marker that smuggles a tally is `UNPARSED`; prose (`campaign skipped`, `SKIP:`) is not the marker; the marker is anchored and case-sensitive.
- **the decision** — `is_nocount()` is a named function purely so `--selftest` can reach it, with cases for silence, a skip-only log, a skip beside a real tally, and an all-zero tally.
- **the call** — five cases run the whole tool over a temp log dir and assert the **exit code**, and they render the non-quiet table so its flag cannot drift from the verdict again. Reviews showed that mutating the call site re-opened the 72% hole, and that the `UNPARSED`→exit-code wiring was unpinned, both with every other case green.

Every one of these was added because a mutation survived without it. Ten deliberate breakages are caught that previously passed in silence.

## The cost of the floor, and what pays it

Giving the suite an unconditional two-check floor removes the backstop it used to have: before it, a campaign whose summary was reworded printed no shape the tally knew, fell to `NOCOUNT`, and stopped the sweep. With the floor, the same rewording would drop 164 checks from the headline with everything green.

So the `aaf` target now requires one of exactly two outcomes — a readable tally or a declared skip — and fails loudly on anything else. Verified: rewording the campaign summary to `164 ok / 0 bad` with `tsn-gen` present makes the suite exit 2 with a message naming the problem.

## The limit, stated

A suite that prints `SUITE-SKIP:` and then runs checks whose summary matches no shape has those checks dropped without a `NOCOUNT` — `UNPARSED` covers it when the summary looks tally-shaped. Written into the module docstring rather than left to be found.

## How to validate

```bash
python3 scripts/suite_tally.py --selftest
make -C tb/verilator/tsn_fuzz                          # skip path -> 2 checks
TSN_GEN_ROOT=<path> make -C tb/verilator/tsn_fuzz      # campaign path -> 166
./scripts/run_all_suites.sh                            # expect exit 0, no NOCOUNT
python3 scripts/gen_toc.py --check && python3 scripts/docs_check.py
(cd tests && behave)
```


Mister-M-alt:
Review returned NEGATIVE. All three blockers fixed, plus 4–10. The closing observation was the one that mattered most, so I'll start there.

> *Either half closes #84 on its own — the half that loosens the gate is the one that is not required.*

Correct, and I verified it before changing anything: with the `SUITE-SKIP` line stripped, the skip-path log still tallies and clears `NOCOUNT` on the matrix echo alone. So the marker was never needed to close #84, and it was buying a hole. **The marker no longer suppresses `NOCOUNT`.** It is now purely reporting — it says why a total is smaller and does nothing else.

**Finding 4 — 72% of the sweep behind a green run.** Reproduced exactly: replacing `shaper_core.log` with a lone marker took the total from 2,113,139 to 591,512 at exit 0. Now:

```
tally rc=1
checks: 591512   in-suite failures: 0
ACCOUNTING FAILURE -- no check count could be read from 1 suite log(s):
  NOCOUNT  shaper_core
The check total above is INCOMPLETE and must not be quoted.
```

The measurement is in the module docstring, because the reason this rule exists is less obvious than the rule. A suite that must skip reports the checks it *did* run — which is exactly what `tsn_fuzz` now does.

**Finding 5 — the justification was false, and it hid a regression.** You are right that CI exports `TSN_GEN_ROOT` (already on `main-push` before #84 was even filed), so `SUITE-SKIP` is never emitted there and "CI cannot pass without the optional tool" described developer machines. That claim is gone. More importantly the regression it enabled is gone with it: had `TSN_GEN_ROOT` ever stopped reaching the sweep, CI would have gone green with the campaign silently not run. It now fails.

**Findings 1, 2, 3 — the coverage did not pin what it claimed.** Also correct, and the diagnosis was the useful part: the SELFTEST tuple discarded the skip list, so any case whose only expectations were "0 checks, 0 failures, no unparsed" held whether or not the regex matched prose.

- Every tuple now carries an explicit `want_skips`, asserted.
- `skip-adds-nothing`'s fixture carries a tally *on the skip line*, so it can tell "excluded from the shapes" from "scanned like any other line".
- The classification is extracted into `is_nocount(matched, skipped)` — a named function purely so `--selftest` can reach it — with three new `nocount-*` cases exercising it directly. It takes `skipped` and ignores it, so the one place tempted to consult it says out loud that it does not.

Your four mutations, re-run: M1 (regex matches prose) → **3 FAIL**; M4 (delete the exclusion) → **2 FAIL**; M5 (skips suppress `NOCOUNT`) → **2 FAIL**; M6 (skip escapes `UNPARSED`) → **3 FAIL**. All were silent before.

**Finding 6** — the marker now still faces the `UNPARSED` net, with a case (`skip-line-is-not-a-hiding-place`) built from your fixture. Plus `skip-marker-must-start-the-line` for the anchoring and case-sensitivity you probed.

**Finding 7 — "singular" was wrong.** `--check` gates two independent contracts under one exit code (the 13 generated artifacts' staleness, and the untested-module ratchet — two separate `rc = 1` sites, either can fail while the other holds). The echo now reads `traceability contracts (drift + ratchet): 2 checks: 2 PASS, 0 FAIL`, and the comment explains why it is 2 and deliberately not 63. Confirmed again that the echo cannot run when the check fails: stale matrix → exit 2, zero `2 checks` lines, and now `NOCOUNT` as well.

**Finding 8** — `docs/testing/TESTING.md:195-198` and `:243-250` both updated; the first said the opposite of what the tool now does.

**Findings 9, 10** — the `TALLY_SHAPED` note is back beside `TALLY_SHAPED`; "two things" is now "three lines — two that count and one that deliberately does not".

**Re-measured**

| scenario | suite exit | checks | accounting |
|---|---|---|---|
| tsn-gen absent | 0 | **2** | clean; skip listed |
| tsn-gen present | 0 | **166** | clean, no skip |
| stale matrix | **2** | 0 | no `2 checks` line; `NOCOUNT` |
| lone marker replacing a real suite | — | 591,512 | **`NOCOUNT`, rc=1** |
| full sweep | **0** | 2,113,140 | 50/50, no NOCOUNT |

`--selftest` 19 cases PASS; `docs_check`, `check_doc_paths`, `check_archive`, `gen_toc --check` and `--verify-anchors` all clean. I also caught and reverted a regenerated `TEST_RESULTS.md` timestamp my `TSN_GEN_ROOT` run produced — not part of this change.

The shape of this round is worth naming: I added a mechanism that was not required, and it was the only part that could hurt. The fix was to delete its power, not to guard it better.

Mister-M-alt:
Re-review returned NEGATIVE on two lines of prose. Both fixed, plus 3–6. Also: this branch collided with #81 and I resolved it — details at the end, because they change what this PR now contains.

**Finding 1 (blocking) — `cosim.py`'s docstring still asserted the rejected invariant.** Exactly right, and the worst place for it to survive: the function that *emits* the marker, telling the next maintainer the marker is what keeps the suite out of `NOCOUNT`. Now says the opposite, with the reason: the traceability contracts do that, and giving the marker the power would let a suite hide a campaign behind a green sweep.

**Finding 2 (blocking) — the README's own TOC blurb still said "one".** Fixed; the body had been corrected and the summary had not.

**Finding 3 — the table flag was the last copy of the rejected predicate.** Deleted. `is_nocount()` is computed once per log and both the table and the verdict read it, so they cannot disagree again. Your example — the row printing `<- 1 SKIPPED` with no `NO COUNT` while the block below listed `NOCOUNT` — is gone.

**Finding 4 — `--selftest` did not exercise `main()`.** The sharpest one: mutating the *call site* re-opened the full 72% hole with every case still green. The self-test now runs the whole tool over a temporary log dir and asserts the **exit code**, which is what CI consumes:

- `e2e-lone-marker` → rc 1 (a suite whose only content is a marker still fails)
- `e2e-marker-plus-real-tally` → rc 0 (…and passes once it reports what it did run)
- `e2e-one-silent-suite-among-many` → rc 1

Your call-site mutation (`nc = (not matched) and not skipped`) now → **2 FAIL**, was silent.

**Finding 5 — `campaign skipped: 0 pass, 0 fail` still cleared the gate.** You marked it pre-existing and non-blocking; I fixed it, because it is precisely the wording #81 proposed and this PR argues against. `is_nocount` now also returns true when the tallies sum to **zero checks and zero failures** — the file's own motto turned on the instrument. Verified no suite in the tree reports zero before adding it, so nothing regresses. That fixture now gives `NOCOUNT`, rc 1, *"must not be quoted"*. Mutation (drop the guard) → **2 FAIL**.

**Finding 6 — the two stale figures.** `README.md:25`'s "3153 checks" (predates the 1722.1 campaign deletions) and `TESTING.md:226`'s "six summary shapes" (`SHAPES` has five) both corrected.

**The collision.** While I was working, the parallel agent pushed #81's full content onto this branch, and #81 has since merged to `main-push`. Two things worth recording:

- **They took the #81 feedback** — commit `8029814c Preserve unknown fuzz accounting` reverts the `0 pass, 0 fail` skip wording back to prose. That was the objection I raised on #81, and it is resolved at the source.
- **My earlier commits reached `main-push` through that merge** (`aac0b6a7`, `268147b7` are both ancestors of it), so the `SUITE-SKIP` marker and the traceability count are already there. This PR's diff is now the **4 files** of review fixes on top.

I resolved two rebase conflicts by hand: `cosim.py`'s docstring (merged both texts — mine adds the do-not-give-it-that-power warning, theirs the "smaller total" phrasing) and the README TOC, where I kept **their** separator convention (they changed it deliberately in `43a4c417`) with my wording for the one entry whose meaning changed.

**Re-measured after the merge**: `--selftest` 22 cases PASS; full sweep **50/50, exit 0, 2,113,140 checks**, no NOCOUNT; behave **322 scenarios / 1,526 steps, 0 failed**; `docs_check`, `check_doc_paths`, `check_archive`, `gen_toc --check` and `--verify-anchors` all clean. I also reverted a submodule-pin change the branch switch left behind.

The PR body is now stale in the ways you listed — it still describes the rejected design. I will rewrite it before merge rather than leave the merge record wrong.

Mister-M-alt:
Third review returned **POSITIVE**. All seven findings addressed anyway, including the five marked non-blocking — one of them turned out to be more than a wording fix.

**Finding 1 — `TESTING.md:200-201` was false, and the reason mattered.** You were right that the sentence contradicted the paragraph above it, and right about why: the two-check floor removed this suite's `NOCOUNT` backstop. A campaign whose summary was reworded would drop 164 checks into a green sweep. Your three fixtures reproduce exactly.

Rather than record the loss, I restored the backstop. The `aaf` target now requires one of exactly two outcomes — a readable tally, or a declared skip — and fails on anything else. Verified by rewording the campaign summary to `164 ok / 0 bad` with `tsn-gen` present:

```
tsn_fuzz: the campaign printed NEITHER a tally suite_tally.py can
  read NOR a declared skip. Its checks would be counted as zero
  with the sweep still green. Restore one of the two shapes.
make: *** [Makefile:103: aaf] Error 1     (suite rc=2)
```

The sentence is now true again *and* says why the reason changed — it used to be fatal by accident, and is now fatal on purpose.

You were also right to correct my review comment: *"had `TSN_GEN_ROOT` ever stopped reaching the sweep… it now fails"* was wrong. It exits 0 with 2 checks and a `SKIPPED` line. The marker's value is that it is not silent, not that it fails. That claim was only ever in a comment, and I am not going to leave it standing uncorrected.

**Finding 6 — the three self-test gaps.** All closed, and all three of your surviving mutations now caught:

| mutation | before | after |
|---|---|---|
| M8 table flag carries a private copy | survived | **2 FAIL** |
| M12/M15 `UNPARSED` no longer fatal | survived | **2 FAIL** |
| M14 empty logdir returns 0 | survived | **2 FAIL** |

The e2e cases now render the **non-quiet** table and assert that `NOCOUNT` in the verdict and `NO COUNT` in the row always agree — which is what makes M8 catchable at all.

**Finding 3 — the diagnostic told a developer to do what they had already done.** Fixed: the two causes are now distinguished per suite (`printed no tally line at all` vs `reported a tally, but it measured 0 checks and 0 failures`), with different advice for each.

**Finding 2** — the module docstring's headline said "unparseable or absent" and defined `NOCOUNT` as "no tally line at all". Both now cover the third way. **Finding 4** — "seventh" → "sixth". **Finding 5** — the PR body is rewritten; the case count is **28**, not 22.

**Finding 7 — the pre-existing README staleness.** Fixed since I was in the file: "Four co-simulation campaigns" → one, the `make legacy` line documenting a target that does not exist, "Table of the four paths" → one row, and the "2 tracked gaps" that belonged to the deleted AECP campaigns. The orphaned `legacy` comment above `clean:` is gone too.

**On the divergence you found** (`checks-colon-failures-colon` lacking the `(?<![\w.])` guard, so `subchecks: 419 failures: 0` is swallowed): I have left it. It is pre-existing, independent of this PR, and fixing a shape regex is the kind of change that wants its own before/after over all 50 logs rather than riding along here. Worth an issue — say the word and I will file it with your differential-sweep numbers.

**Re-measured**: `--selftest` 28 cases PASS; skip path 2 checks; campaign path 166; stale matrix exit 2 with no tally; reworded tally exit 2; full sweep **50/50, exit 0, 2,113,140**; `docs_check`, `check_doc_paths`, `check_archive`, `gen_toc --check` and `--verify-anchors` all clean. Reverted a regenerated `TEST_RESULTS.md` timestamp the `TSN_GEN_ROOT` run produced.

PR #90, opened 2026-08-17
Make a merge that lands past its branch say so

### LINKED ISSUE #88: The merge lane can strand work after a PR merges

Mister-M-alt:
Closes #88.

## Status

In review at `ee43ff2ba432a9148d6aabec18eb1c72f337b84e`.

## Description

Adds a post-merge containment command and makes the issue-to-merge lane explicit:

- Do not merge while any review round remains outstanding.
- Run the containment check when the branch stops moving.
- Re-run verification on the merge result.
- Check that every observable branch tip is contained in `main-push`.
- Treat a missing measurement or unresolved ref as UNKNOWN, never as success.
- Gate the containment self-test from `scripts/run_all_suites.sh`.
- Treat GitHub `verilator-suites` and `yosys-portability` as optional signals while requiring both equivalent commands locally.

## Correctness properties

- After fetch, divergent local and `origin/<branch>` tips are both checked, including when the caller spells the remote ref explicitly.
- Fetch refreshes every origin branch even when the configured refspec is narrow.
- Shallow history is deepened before evaluation; `--no-fetch` refuses a shallow repository.
- `--merged-prs` checks a same-repository live tip only when its first-parent history continues the frozen merge-time head OID and no post-merge GitHub ref event broke continuity.
- `--merged-prs` requests a bounded complete candidate set, refuses possible truncation, sorts by merge time, and only then selects the requested count.
- Fork heads, deleted or recreated branch names, post-merge force pushes, and discontinuous live tips become UNKNOWN.
- Patch-id equivalence is used only for linear histories because merge patch IDs omit unique resolution content.
- Linear replay evidence uses whitespace-preserving verbatim patch IDs plus exact touched-path postimages, never `git cherry` alone.
- Proof-producing Git comparisons disable external diff drivers, text conversion, and submodule-ignore configuration.
- Git replacement objects are disabled and active legacy graft files are refused.
- Branch and base operands must be exact ref names or full object IDs, never revision expressions; full object IDs remain immutable through fetch normalization.
- Full object IDs remain immutable when same-named local branches and tags exist.
- Explicit `refs/remotes/origin/` operands remain in the namespace refreshed by the origin fetch even when configured remote names overlap that prefix.
- The implicit current branch keeps its `refs/heads/` identity even when its legal name is 40 hexadecimal characters.
- A missing fetched branch or base remains fully qualified, so a same-named tag cannot replace it.
- Offline short names retain branch identity and cannot fall through to same-named tags.
- Configured remote shorthand retains remote identity even when the remote name contains slashes, its tracking ref is missing, or a slash-named local branch collides with it.
- Normal mode keeps an explicit local base origin-qualified after fetch, including when the origin base disappeared.
- Normal mode rejects non-origin remote-tracking refs that its origin-only fetch cannot refresh, including ambiguous shorthand and missing tracking refs.
- Squash detection compares every touched path with rename folding disabled and literal path handling enabled.
- A multi-commit branch with no net tree change is recognized as contained.
- Path enumeration and comparison failures become UNKNOWN.
- A fetched branch whose origin ref disappeared remains UNKNOWN even when a stale local branch exists.
- Repeated or conflicting command-line options are rejected.
- The suite runner invokes the containment self-test from the repository root regardless of caller directory.
- Self-test count relationships use controlled linear history, so the required post-merge run passes when the caller HEAD is a merge commit.

The documented conservative case remains: a multi-commit squash whose touched paths are later edited on the base can report STRANDED. This is a false alarm, not a false pass, and the differing paths are printed.

## Validation

- Candidate base `39d059b4d79ad227ce22b41f6ea081eb629b66a7` is an ancestor of the reviewed head; candidate merge tree `5b1934a0d9e8c6d9cc51c8b6458f042a9a9c72b3` equals the reviewed head tree.
- `python3 scripts/check_merge_containment.py --selftest`: 127 assertions, PASS.
- The same self-test from a synthetic two-parent merge HEAD: PASS.
- `python3 scripts/check_merge_containment.py --merged-prs 20`: 6 live branches contained, rc 0.
- Final-head local `scripts/run_all_suites.sh`: 50 of 50 suites, 2,113,140 checks, 0 failures, 0 timeouts.
- `tsn_fuzz` declared one zero-count field-campaign skip because `tsn-gen` is absent; the suite still reported an accountable result.
- Final-head local `syn/yosys/run.sh`: 44 of 44 tops passed.
- Yosys tied-input gate: PASS, 0 unjustified ties.
- Yosys observer-purity gate: PASS, 0 violations.
- Two independent cleared-context reviews of the exact head: VALIDATED.
- Python bytecode compilation: PASS.
- `bash -n scripts/run_all_suites.sh`: PASS.
- Documentation check: 0 findings across 204 files.
- Documentation paths: 1,086 cited paths resolve.
- Archive gate: PASS.
- TOC gate: PASS.
- Anchor verification: PASS.
- Added documentation contains no em dash.

## Regression coverage

The self-test executes these cases end to end:

- stale local branch behind its fetched remote;
- local branch carrying an unpushed commit beyond its contained remote, including explicit `origin/<branch>` spelling;
- merged PR with a frozen merge-time head OID;
- fork PR, recreated same-name branch, and post-merge ref-event refusal;
- full object IDs colliding with hex-named origin branches, local branches, and tags;
- an implicit 40-hex current branch and deleted branches colliding with tags;
- offline short branch and base names colliding with tags;
- merged PR results returned out of merge-time order and possible candidate truncation;
- revision suffixes on branch and base operands;
- shallow offline refusal and automatic deepening;
- failed ancestry plumbing;
- replacement refs and legacy grafts attempting to rewrite containment ancestry;
- ambiguous short names shared by a branch and tag;
- non-origin remote-tracking branch and base refs that normal mode cannot refresh, including slash-bearing remote names, ambiguous prefixes, and missing tracking refs;
- explicit origin refs colliding with configured slash-bearing remote prefixes;
- failed path enumeration and comparison plumbing;
- multi-commit squash followed by unrelated base advancement;
- multi-commit history with no net tree change;
- rename versus copy with the source path left behind;
- pathspec-shaped filename treated literally;
- linear patch-equivalent rebase followed by a same-path edit;
- whitespace-different patches that `git cherry` incorrectly equates;
- equal verbatim patch IDs applied to different repeated locations;
- textconv and external diff configuration hiding changed blobs;
- submodule-ignore configuration hiding a missing gitlink update;
- deleted remote branch with a stale contained local tip;
- deleted origin base with a stale explicit local base;
- merge commit carrying unique resolution content that `git cherry` omits;
- duplicate and conflicting command-line operands, including repeated `--no-fetch`;
- one-commit count checks isolated from a merge-result caller.

## Definition of done

- [x] The lane states when a PR may be merged.
- [x] The lane contains an exact post-merge command.
- [x] A live branch that advances after merge is detected without reading the review thread.
- [x] The containment self-test is part of the repository sweep.
- [x] Every reproduced false-success path has a committed regression.
- [x] The self-test passes from an ordinary commit and a merge result.
- [x] Fresh review is complete.
- [ ] Merge approval requirements are met.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,10 +48,12 @@ flowchart LR
     PR --> RV[In review<br/>multiple agents, CLEARED context]
     RV -->|findings| W
     W -->|fixed| RV
-    RV -->|2 positive:<br/>1 own + 1 EXTERNAL| M[merge to main-push]
-    M --> G[re-run the FULL bar<br/>on the merge result]
+    RV -->|2 positive:<br/>1 own + 1 EXTERNAL| G[validate candidate merge result<br/>with the FULL local bar]
     G -->|regression| W
-    G -->|clean| D[Done]
+    G -->|clean| M[merge to main-push]
+    M --> C[branch stops moving<br/>run containment]
+    C -->|finding| F[follow-up issue + PR]
+    C -->|clean| D[close issue manually<br/>Done]
 ```
 
 1. **Move the issue to *In progress*** on the project board before the first
@@ -60,8 +62,9 @@ flowchart LR
    built twice on 2026-08-16 because the issue was filed after the work
    started.
 2. **Cut the branch from the issue**: `gh issue develop <N> --base main-push`.
-   This links branch to issue on GitHub, so the PR closes the issue and the
-   board moves itself. A hand-named branch does neither.
+   This links the branch and issue on GitHub. Because `main-push` is not the
+   repository's default branch, merging its PR does not auto-close the issue.
+   Close the issue manually only after the post-merge containment check passes.
 3. **Do the work on that branch**, with the §3 verification bar met *on the
    branch* — a PR is not the place to discover the sweep is red.
 4. **Open the PR** against `main-push`, with the template below and the
@@ -108,12 +111,14 @@ flowchart LR
    `MERGED`, CI is green, and the branch still has commits ahead of the merge.
    Nothing in the lane compares those two facts unless step 7 does.
 
-   The issue closes itself; move the card to *Done* if it does not.
-7. **Re-run the whole verification bar ON THE MERGE RESULT, every time.** A
-   merge is a change nobody wrote and nobody reviewed, and *"Merge made by the
-   'ort' strategy"* is not evidence of anything. Gate the merged tree exactly
-   as §3 gates a hand-written one — full Verilator sweep, both repos' suites,
-   behave, the lint ratchet, yosys — before the merge button, not after.
+7. **Validate the candidate merge result, then prove containment.** A merge is
+   a change nobody wrote and nobody reviewed, and *"Merge made by the 'ort'
+   strategy"* is not evidence of anything. Before pressing merge, construct a
+   candidate from the latest `origin/main-push` and the reviewed PR head, then
+   gate that tree exactly as §3 gates a hand-written one: full Verilator sweep,
+   both repos' suites, behave, the lint ratchet, and Yosys. If the reviewed head
+   directly descends from the unchanged base, its tree is the candidate merge
+   tree; record both object IDs with the local gate results.
 
    This is not defensive box-ticking. On 2026-08-16 two lanes independently
    added the *same* six AECP settings-face pins to `KL_pp_shadow.sv` — one
@@ -124,7 +129,7 @@ flowchart LR
    split out, and restored an inventory row for an opcode the engine no longer
    decodes.
 
-   **Then check the merge actually took the branch:**
+   **After merge, check that it actually took the branch:**
 
    ```bash
    python3 scripts/check_merge_containment.py origin/<branch>
@@ -138,9 +143,15 @@ flowchart LR
 
    **Run it when the branch stops moving, not at the merge button.** Both
    incidents were *contained* at the instant they merged; the commits that
-   ended up stranded were pushed afterwards, by the review round that was still
-   running. A check run at merge time cannot see them. The moment that catches
-   it is the one where the card moves to *Done*.
+   ended up stranded were pushed afterwards as review activity continued. For
+   #86 a round was in flight. For #77 the stated bar had been met, but review
+   did not stop. A check run at merge time cannot see later pushes. The moment
+   that catches them is the one where the card moves to *Done*. Once
+   containment is clean, close the issue manually and move its project item to
+   *Done*.
+
+   `--no-fetch` disables Git ref refresh only. A `--merged-prs` sweep still
+   queries GitHub for the merged PR list and branch timeline evidence.
 
    Its `--selftest` runs inside `scripts/run_all_suites.sh` next to
    `suite_tally.py`'s, so the tool cannot rot into a green that means nothing
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,8 +87,24 @@ flowchart LR
    validated, with the gate results. The findings, the mutation tables and the
    clause arguments belong in the review itself and in the commit messages —
    a PR thread that reprints them is a PR thread nobody reads to the end.
-6. **Merge back into `main-push`** only once the findings are answered. The
-   issue closes itself; move the card to *Done* if it does not.
+6. **Merge back into `main-push`** only once the findings are answered, and
+   **not while a review round is in flight.** A round that has not reported is
+   a round outstanding; merging past it is merging unreviewed code with a
+   review thread attached.
+
+   This has happened twice, both times losing work rather than shipping a bug:
+
+   | PR | merged | state at that moment | cost |
+   |---|---|---|---|
+   | #77 | 2026-08-16 18:10 | round 3 of 6 running | rounds 4, 5 and 6 each returned NEGATIVE **after** the merge — an ungraded refusal arm where mutating the code was silent, and a test that could not fail. Re-landed as #85 |
+   | #86 | 2026-08-17 07:54 | a round running, which then found a hole in its own fix | three commits stranded on the branch; re-landed as #89, tracked by #87 |
+
+   Both were recoverable and neither was noticed by anything except a reviewer
+   checking by hand. The failure is silent by construction: `gh pr view` says
+   `MERGED`, CI is green, and the branch still has commits ahead of the merge.
+   Nothing in the lane compares those two facts unless step 7 does.
+
+   The issue closes itself; move the card to *Done* if it does not.
 7. **Re-run the whole verification bar ON THE MERGE RESULT, every time.** A
    merge is a change nobody wrote and nobody reviewed, and *"Merge made by the
    'ort' strategy"* is not evidence of anything. Gate the merged tree exactly
@@ -104,7 +120,33 @@ flowchart LR
    split out, and restored an inventory row for an opcode the engine no longer
    decodes.
 
-   Two merge-specific traps worth naming, both paid for:
+   **Then check the merge actually took the branch:**
+
+   ```bash
+   python3 scripts/check_merge_containment.py origin/<branch>
+   python3 scripts/check_merge_containment.py --merged-prs   # the last 20 PRs
+   ```
+
+   It exits non-zero and names the count when commits are left behind. Replayed
+   against the two merge points in step 6 it reports **4** and **3** stranded
+   commits respectively, which is what nobody was told at the time.
+
+   It is a script rather than a line in this file because a check that depends
+   on somebody remembering is not a check — the same reason
+   `scripts/suite_tally.py` exists. `--selftest` gates it, including the case
+   that a ref which cannot be resolved is an UNKNOWN and **fails**, never a
+   quiet pass.
+
+   **Do not hand-roll it with `git log A..B`.** In this environment that
+   returns **empty** when it follows another git command in the same shell
+   invocation — the proxy hook swallows the output — so it reports "nothing
+   diverged" for a branch that has. That is exactly how #89's description came
+   to claim a fast-forward that was not one. The script uses
+   `git rev-list --count` and `git merge-base --is-ancestor`, both of which
+   answer through a bare number or an exit code, so a swallowed stream cannot
+   be mistaken for agreement.
+
+   Three merge-specific traps worth naming, all paid for:
 
    - **Push the submodule before the superproject.** A superproject pin to a
      processor commit that only exists on a feature branch dangles the moment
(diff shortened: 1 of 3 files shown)


Mister-M-alt (commented):
PR #90 at a3f9edd2e19d3a74efb3e5726e0c8ae0eaebfd06 is not validated. The merged-PR sweep reports deleted branch refs as UNKNOWN even though GitHub retains each merged head OID, and fetching cannot restore deleted refs. The rev-list path also converts empty output to zero, contrary to the stated unknown-is-not-agreement rule. Newly added em dashes violate the documentation requirements. Changes requested.

Mister-M-alt:
Not merged. Two reviews said otherwise — an earlier `CHANGES_REQUESTED`, and this round's NEGATIVE — and the second found the script returning **rc 0 while work is stranded**, which is the exact failure it was written to catch. All findings fixed; re-review requested.

**F1 — answered from a stale remote-tracking ref.** The incident shape verbatim: merge lands, the review round pushes fixes from another worktree, the maintainer's clone still holds the old `origin/<branch>`, and the check says `contained`. It fetches now, with `--no-fetch` explicit rather than accidental.

**F2 — `--merged-prs` was permanently red on a healthy repo.** Two causes, both fixed:
- Deleted branches. Your suggestion was the head OID, and that alone does not work — the OID is not in the clone either, and `git fetch` will not retrieve an object nothing points at. Containment for those is now asked of GitHub's compare endpoint, which still knows.
- Scope. The sweep asked about every merged PR, including ~20 that targeted `main` and were never meant to be in `main-push`. It is scoped to the base branch now.

Result on this repo: **6 PRs, all `contained`, rc 0**. Previously 12 UNKNOWN and rc 1 with nothing wrong.

I also found a third cause while fixing it, which your fixtures would not have shown: `gh pr list --base` takes a branch *name*, and it was being handed `origin/main-push`, so it silently matched nothing and the sweep exited 0 on an empty list — a pass by vacancy. An empty sweep is now `RC_CANNOT_RUN`.

**F3 — squash/rebase false STRANDED.** Real: `allow_squash_merge` and `allow_rebase_merge` are both true here. Content equivalence is checked before anything is called stranded. Synthetic squash merge → `contained [content is identical … squash or rebase merge]`, rc 0; then a genuine commit on top → `STRANDED: 3 commit(s)`, rc 1.

**F4 — a typo'd flag turned stranded into green.** `--bse 43a4c417 origin/…` checked two branches named after the flag's argument and reported contained. Unknown options are rejected now.

**F5 — `gh` absent raised `FileNotFoundError` at the same rc as STRANDED.** Distinct exit codes: 0 contained, 1 finding, 2 cannot run.

**Mutation coverage.** You were right that the SELFTEST table re-implemented the predicate instead of calling it, so the advertised count had no coverage — the same "test that cannot fail" this repo's own rounds keep finding. The cases call `contained()` against real history now, including `stranded-count` asserting both the verdict and the number. The impossible `disagreement` row is gone.

**F6, F7, F8, F9 — wrong facts in a document about getting facts right.** All corrected:
- the two counts were reversed (**3** for #77, **4** for #86 — the fourth pushed during the #89 re-land);
- **round 6 of #77 returned POSITIVE**, not NEGATIVE. That error started in issue #88's body, which I wrote; I have corrected it there too so it does not propagate again;
- the `git log A..B` claim is gone. You could not reproduce it in 12 shapes, and it described a *local private* command proxy in a public contributor guide — where no external contributor has that tool. The design justification stands on its own without it: `rev-list --count` prints one integer and `merge-base --is-ancestor` answers only through an exit code, so neither can be half-read. The text now says the #89 cause was never pinned down, which is the honest version;
- the PR body's claim that #89 "says so" about the pending-checks merge was false. Removed.

**F15, and taking my own advice:** this PR is at `MERGEABLE/UNSTABLE` with `verilator-suites` and `yosys-portability` pending. Per the rule this PR adds, that is a decision to state rather than assume: the full sweep, behave and all doc gates were run locally and are green, and I am **not** merging on it — I am asking for the re-review first.

**Re-measured:** sweep **50/50, exit 0, 2,113,140**; behave 322 / 1,526; `--selftest` 5 cases; `docs_check`, `check_doc_paths`, `check_archive`, `gen_toc --check`, `--verify-anchors` all clean.

Minor items 10–14 not yet done: `--merged-prs` ordering now uses `sort:updated-desc`, and the dead `_git("--version")` guard is gone; the ambiguous-tag precedence note and the TESTING.md mention are worth a follow-up rather than another round here.

Mister-M-alt (commented):
PR #90 at 0edeeda7a269ae00e91e65e03dcec57c1412586b is not validated. The command can return rc 0 for a stale current local branch after fetching only its remote-tracking ref, and the git cherry fallback can hide unique merge-resolution content because merge commits are omitted. A multi-commit squash also becomes falsely STRANDED after the base advances. The self-test does not cover these paths and can fail at an ordinary merge-result HEAD. Four prohibited em dashes remain in CONTRIBUTING.md. Changes requested.

Mister-M-alt:
NEGATIVE again, and rightly — two of the previous blockers I reported as fixed were only *narrowed*, and the code I added to fix a third had three defects of its own and no coverage. Not merged.

**F1 — squash detection was still broken.** You are right that `git diff --quiet base branch` holds only while base's tree is byte-identical, so one unrelated commit later every squash-merged branch reads stranded again. Your five-sequential-squash fixture reproduced exactly: only the newest passed.

It now compares **only the paths the branch touched** (`git diff --name-only $(merge-base) branch`, then `git diff --quiet base branch -- <those>`), which is the direction you suggested. Same fixture:

```
contained  pr1 … pr5   [every path this branch touched is identical in main-push]   rc 0
```

Genuine stranding still caught: squash-landed + one new commit → `STRANDED 3`, rc 1; and a branch that *modifies* a path that already landed → `STRANDED`, rc 1.

**F2, F3, F4 — I deleted `contained_via_api` rather than fix it.** All three of your findings were correct, and the fourth fact settles it: `delete_branch_on_merge` is **false** on this repo and all six swept PRs resolve locally, so that path was never reached. It was speculative code carrying a wrong premise (`files` is the *three-dot* diff, so the squash arm was dead), a `null | length == 0` hole that turned an unreadable answer into `contained`, and two uncaught exceptions at STRANDED's exit code.

Adding unexercised complexity to a gate is how the gate stops meaning anything. A deleted branch is now reported UNKNOWN, loudly, with the docstring saying that GitHub's compare endpoint is the answer when this repo starts deleting branches — and that it needs its own review then, rather than an unrun version sitting in the tree now.

**F5 — the self-test still passed against a broken implementation.** The gap was that it called `contained()` and never `main()`, so the whole aggregation and exit-code path was uncovered. It now runs the tool end to end against a throwaway repo built in `tempfile`, with **three** commits so a count hard-coded to 1 cannot pass. All seven of your surviving mutations now fail:

| mutation | before | after |
|---|---|---|
| M2 count hard-coded to 1 | survived | **3 FAIL** |
| M6 drop the `isdigit` guard | survived | **1 FAIL** |
| M7 `(False, 1, None)` | survived | **3 FAIL** |
| M9 `return RC_OK` | survived | **3 FAIL** |
| M10 drop `stranded += 1` | survived | **3 FAIL** |
| M11 print the count as `0` | survived | **3 FAIL** |
| M12 drop `unknown += 1` | survived | **2 FAIL** |

**F6 — ambiguous refs.** Fixed: a name that resolves to both a tag and a branch is UNKNOWN naming both, instead of silently taking git's precedence. Your fixture (tag at the old commit, branch 3 ahead) went `contained rc 0` → `UNKNOWN rc 1`.

**F8 — `--base` swallowing a following flag.** Fixed; a `--base` value starting with `-` is refused, rc 2.

**F7 and F9 — the count semantics and reverted merges.** Both accurate and both left as stated limits rather than papered over: the number is "commits that are not ancestors", which after a partial squash overstates what is actually missing, and a reverted merge still reads contained because ancestry is the question being asked. Worth a follow-up sentence in the docstring; not worth guessing at.

**Re-measured:** replays exact (**4** and **3**); `--merged-prs 20` → 6 PRs contained, rc 0; sweep **50/50, exit 0, 2,113,140**; behave 322 / 1,526; all doc gates clean.

This is the third round on a 250-line script, and every round has found the same class of thing: a check that looked right and could not fail. Requesting re-review rather than merging.

Mister-M-alt (commented):
PR #90 at 362e3e69e9f8c7e65fa8285957a572f87a36172e is not validated. Containment false-pass paths remain, and the documentation policy is not met. Changes requested.

Mister-M-alt (commented):
PR #90 at ee43ff2ba432a9148d6aabec18eb1c72f337b84e hardens post-merge containment and the review lane. Validated: local Verilator 50/50 (2,113,140 checks), local Yosys 44/44, containment self-test 127/127, and documentation gates pass.

PR #92, opened 2026-08-17
Started/stopped gets one source of truth, and the datapath finally reads it

### LINKED ISSUE #78: Started/stopped is about to have two sources of truth: the ACMP record has the lifecycle but no writer, the dynamic store has the writer but no lifecycle — and nothing reads either
## Summary

The current tree contains two candidate locations for Stream Input
started/stopped state, but only the ACMP listener binding record has the required
lifecycle. This ticket makes that record the single source of truth and connects
`START_STREAMING`, `STOP_STREAMING`, `GET_STREAM_INFO`, and the listener datapath
to it.

PR #77 and its review follow-ups landed the shared streaming predicate and
`SET_CONFIGURATION`. They deliberately withdrew `START_STREAMING` and
`STOP_STREAMING` pending this source-of-truth decision.

## Current state

**ACMP listener binding record**

- `protocol-processor/hdl/acmp/pp_acmp_pkg.sv` defines `f_started`.
- `KL_pp_acmp_listener.sv` clears `f_started` on unbind.
- `KL_acmp_nvm_shadow.sv` includes the field in the captured binding record.
- No AECP command writes the field today.

**AECP dynamic-state store**

- `KL_aecp_dyn_state.sv` still defines `SEL_START` and exports a started vector.
- No dispatched microprogram writes `SEL_START` on `main-push`, so the exported
  vector remains at its reset value.
- The dynamic row has no clear-on-unbind path and is not a suitable second copy
  of binding state.

**Consumers**

- `hdl/milan/milan_datapath.sv` publishes the unused dynamic-store vector but
  does not use it to admit or discard Stream Input data.
- `GET_STREAM_INFO` does not derive `STREAMING_WAIT` from the binding record.

## Source-of-truth decision

Use `f_started` in the ACMP listener binding record as the only writable
started/stopped state. Do not reactivate `SEL_START` as a second state copy.

This direction reuses the record that already clears on unbind and is already
included in the ACMP NVM shadow. The required new path is an AECP-controlled
write into that record plus a read path back to the AECP and datapath consumers.

Relevant requirements and command behavior are already recorded in issues #67
and #70 and in the comments beside `f_started`. In particular, an unbound or
already-started Stream Input treats `START_STREAMING` as a successful no-op.

## Implementation scope

- Add descriptor-indexed read and write access between the AECP engine and the
  ACMP listener record's `f_started` field.
- Implement `START_STREAMING` and `STOP_STREAMING` for Stream Inputs with the
  four-byte command body and `control_data_length` 16 used in issue #67.
- Return `NOT_SUPPORTED` for Stream Output targets.
- Preserve lock checks, descriptor existence checks, full response bodies, and
  the existing successful no-op behavior for an inapplicable state transition.
- Source `GET_STREAM_INFO.STREAMING_WAIT` from the same binding record: 0 for a
  bound and started input, 1 for a bound and stopped input.
- Gate Stream Input data handling from the same started bit so a stopped input
  discards arriving stream data and a started input can process it.
- Keep unbind as the lifecycle owner that clears the started bit.
- Remove, retire, or leave permanently unwritten the dynamic-store `SEL_START`
  row so it cannot become a second writable source.

## Acceptance criteria

- [ ] There is exactly one writable started/stopped bit per Stream Input, the
  `f_started` field in its ACMP binding record.
- [ ] `START_STREAMING` on a bound and stopped Stream Input sets that bit and
  returns the expected full response.
- [ ] `STOP_STREAMING` on a bound and started Stream Input clears that bit and
  returns the expected full response.
- [ ] Repeating the current state, or issuing the command while unbound, returns
  success without changing the record.
- [ ] Stream Output targets return `NOT_SUPPORTED`; nonexistent descriptors and
  lock conflicts retain their existing refusal precedence and response shape.
- [ ] Unbind clears the bit, and rebind starts from stopped without a stale
  dynamic-store value restoring it.
- [ ] `GET_STREAM_INFO.STREAMING_WAIT` and datapath admission both observe the
  same record value.
- [ ] No test or integration path consumes a writable `SEL_START` copy.

## Verification

**Processor desk tests**

- In `protocol-processor/tb/acmp_listener`, start a bound input, unbind it, and
  prove the record clears.
- In `protocol-processor/tb/pp_top`, cover bound start, stop, repeated no-op,
  unbound no-op, Stream Output refusal, lock conflict, and nonexistent index.
- In `protocol-processor/tb/pp_top`, start, unbind, rebind, then read the state
  through `GET_STREAM_INFO`; it must report stopped.

**Model-driven integration**

- In `tb/verilator/milan_dp`, prove a stopped sink discards arriving AVTPDUs and
  its Stream Input counters do not advance, while the started case admits data.
- Compare solicited `GET_STREAM_INFO` after each transition with the record value
  observed by the datapath.

Run the processor suite through `protocol-processor/scripts/run_suites.sh`, then
run the repository Verilator and contract gates documented in `CONTRIBUTING.md`.

## Dependencies and related work

- The shared bound and streaming predicates landed with PR #77 and its review
  follow-ups, so there is no implementation blocker for this ticket.
- Issue #69 owns unsolicited notifications triggered by state changes. This
  ticket must expose the transition cleanly but does not duplicate that trigger
  framework.
- Issue #70 owns the real nonvolatile backend and physical power-cycle proof.
  This ticket must use the existing ACMP NVM shadow field but does not claim that
  the backend is complete.

Mister-M-alt:
## Status

Ready for re-review. Closes #78. VERSION `0x0002_004F`.

## Description

Started/stopped gets **one** source of truth: `f_started` in the ACMP binding
record. Milan Section 5.3.8.7 calls the state "undefined when the Stream Input
is not bound", so it is a property of the binding, and only that record has the
lifecycle — cleared on unbind, projected by the NVM shadow, restored through the
boot preload.

Three things land together, because the issue's title is literally true and each
half is useless alone:

1. **`START_STREAMING` (0x0022) / `STOP_STREAMING` (0x0023) are served.**
   IEEE Figure 7-59 gives command and response one shape — `descriptor_type`
   @24, `descriptor_index` @26 — so every arm answers at **cdl 16**, refusals
   included. Milan Section 5.4.2.19/.20 narrow IEEE's "STREAM_INPUT or
   STREAM_OUTPUT" to Stream Inputs; a Stream Output, and every other type, gets
   `NOT_SUPPORTED`.
2. **The dynamic store's selector 6 is retired, not reused.** It was a fully
   writable second copy with no lifecycle. The commands reach the record through
   a **write-only request region** that stores nothing, so a second source
   cannot come back by accident. It costs no opcode.
3. **The datapath reads the bit.** `pp_aecp_strm_started_w` was declared,
   connected and read by *nothing*, so both behavioural shalls of Section
   5.3.8.7 were unimplemented.

### Where the discard sits, and why it matters

On the listener **accept pulse**, not on the classification entry. The clause
says a stopped input "shall discard the Stream AVTPDUs it **receives**" — so it
still receives, still classifies and still counts. Gating the classifier instead
(the first cut) made `KL_stream_table` emit a bind_fall/bind_rise pair around
every stop/start, and Milan Table 5.6 makes that edge the counter-**reset**
event: all ten Stream Input counters were wiped, plus a spurious MEDIA_UNLOCKED,
on a sink that never left its binding. The suite now grades FRAMES_RX across a
stop/start pair.

The CRF Media Clock Input has no classification entry at all (`KL_crf_rx` keys
on subtype), so it carries the same predicate directly. Every Stream Input the
entity advertises is covered, which is what "For each Stream Input" requires.

### The bind is what defines it

A bind carrying STREAMING_WAIT lands **stopped**; without it, **started**. That
is IEEE Section 7.4.35's own premise, and Milan Table 5.9 bit 28 agrees from the
other side — it reports 1 only for a sink that is "bound and stopped", while
ACMP reports the *saved* flag (Table 5.37), so the two answers would contradict
each other any other way. `f_sw` and `f_started` are deliberately kept apart for
that reason.

### Liveness

The request is a **posted write** into a one-deep holder, drained at top
priority. The AECP microCPU stalls its E stage on `!st_ready_i` with no
watchdog, so tying that to "the ACMP walker happens to be idle" would have let a
busy or wedged walker freeze the entire AECP plane — every READ_DESCRIPTOR,
LOCK and GET_* with it. Every other microCPU face is bounded; this one is now
too.

### Notification

A committed transition raises Milan Table 5.22's GET_STREAM_INFO trigger
("Started/stopped state (Stream Input only)"); a repeat that changes nothing
raises none. Issue #69 owns the rest of the trigger set — this one is here
because it did not exist to be triggered until the command landed.

## Clause corrections

- `docs/testing/PDU_GETTER_SETTER_VERIFICATION.md` justified not implementing
  these with Section 5.3.7.3 — the **Stream Output** clause ("excludes the
  possibility for a Stream *Output* to be stopped"). It cannot license skipping
  a Stream *Input* shall. That page is `[OBSOLETE]`-bannered so it stands as the
  record; the live roadmap and the audit's B12 are corrected.
- `docs/testing/TORTURE_CAMPAIGN.md` told a bench operator to expect
  `NOT_IMPLEMENTED` for a Stream Output START. It is `NOT_SUPPORTED`.

## Scope, stated

Section 5.3.8.7 has **three** shalls. The two behavioural ones are implemented
and graded. The third — "saved in a non-volatile memory and restored after a
power cycle" — is not: the record projects and restores the bit, but
`KL_pp_shadow` sets `NVM_BACKED_C = 1'b0` behind a blank-flash stub, so nothing
survives a power cycle (issue #70). The record layout version moves to `0x02` so
a binding saved by firmware that never wrote the bit is refused rather than
fast-connected into silence.

## How to validate

    cd protocol-processor && bash scripts/run_suites.sh
    python3 protocol-processor/scripts/check_upc_map.py
    bash scripts/run_all_suites.sh
    cd tests && behave

## DoD

Every acceptance criterion in #78 is met. Each new guard was mutated and shown
able to fail; the tables are in the review comments.


Mister-M-alt:
Validation is blocked at a0f1aa52881078f8a97c18763106eefca8cd83ea. STOP_STREAMING does not gate the advertised CRF Stream Input, so the command reports stopped while CRF PDUs continue driving clock recovery. For AAF inputs, feeding started state into KL_stream_table bound inputs also turns STOP/START into false bind_fall/bind_rise events even though the ACMP binding remains present, resetting downstream monitor and queue state. Add CRF discard coverage, preserve binding edges across STOP/START, and reconcile the active roadmap, audit, and KL_pp_shadow documentation before re-review.

Mister-M-alt:
**Review round 1 — three reviewers, cleared context, separate lenses. All three returned CHANGES-REQUESTED. They were right.**

Two found the same blocker independently.

### Blockers

**The discard was at the wrong layer.** Gating the classification table on the
started bit made `KL_stream_table` emit a bind_fall/bind_rise pair around every
stop/start. Milan Table 5.6 makes that edge the counter-**reset** event, so a
stop/start wiped all ten Stream Input counters and injected a spurious
MEDIA_UNLOCKED on a sink that never unbound. The clause had said where the gate
belonged: a stopped input "shall discard the Stream AVTPDUs it **receives**".
Moved to the accept pulse, downstream of the monitor that counts. The test had
asserted "NOT matched" — it now asserts the frame *is* matched and that
FRAMES_RX advances while stopped.

**The first un-watchdogged stall in the microCPU.** `st_ready_i` was tied to the
ACMP walker being idle; a wedged walker would have frozen the whole AECP plane.
Now a one-deep posted-write holder, drained at top priority.

**Three checks could not fail.** The descriptor index was never graded (every
request in the tree targeted sink 0, so `strm_set_index_o = 16'd0` survived);
the lock refusal's *effect* was measured against a bit already at 0; the
NVM-restore path into the mirrored vector was never exercised. Each now has a
value to fail on.

### Also fixed

| Finding | Resolution |
|---|---|
| Table 5.22 started/stopped notification never fired (`act_notify_o` dangling) | scoped `act_strt_chg_o` pulse; a real transition raises one, a repeat raises none |
| `GET_STREAM_INFO` bit 28 had no test at all | graded across STOP/START, plus BOUND still set while stopped |
| NVM restore from firmware that never wrote the bit would fast-connect into silence | `LAYOUT_VER_P` 0x01 → 0x02 (which then exposed a negative test the bump had invalidated) |
| CRF Media Clock Input reported stopped while still consuming | brought under the same predicate; no advertised Stream Input is exempt now |
| Docs claimed 5.3.8.7 implemented; it has **three** shalls | scoped to the two behavioural ones, with `NVM_BACKED_C = 1'b0` named |
| Region-nibble comment cited a weaker invariant than the one that holds | corrected to the store's own contract (a locate answers `st_rdata_o = 0`) |
| Out-of-range sink request dropped silently | counted (`dbg_strq_drop_o`) |
| Stale banner: "PERMANENTLY ZERO ... no microprogram writes SEL_START" | corrected; the port is live and load-bearing |
| Offline model's equivalence claim broader than the truth | `NOT MODELLED` note beside the inventory rows |

### Mutation results

Every new guard was mutated and shown able to fail. Reviewer-supplied mutations
that previously survived and now fail: `strm_set_index_o = 16'd0` (2), `WRITE_ST`
moved ahead of `CHECK_LOCK` (1), mirror skipping `X_PRELOAD` (1),
`STREAMING_WAIT` back to a constant `1'b0` (1), classifier gated instead of
accept (2), `f_started != strtL_val_r` dropped (1), `strm_set_ready_o` tied high
(1).

One mutation is *equivalent*, not a gap: `act_strt_chg_o <= strq_job_r` without
`cellmut_r` — the no-op path never reaches `X_WB`, so the term is redundant with
the control flow rather than untested.

### Bar

processor 14,088 checks / 0 failing · uPC gate PASS (42 ↔ 62) · behave 340
scenarios / 1,632 steps · lint ratchet PASS at exactly 100 · docs_check,
check_doc_paths, check_archive, check_entity_shape, check_wire_accountability
all PASS. Full superproject sweep re-running on the final tree; numbers to
follow.

Findings are not a sign-off — this needs a re-review before it merges.

Mister-M-alt:
**Full bar on the final tree.**

| gate | result |
|---|---|
| processor suite (`protocol-processor/scripts/run_suites.sh`) | **14,088 checks / 0 failing** |
| uPC map gate | PASS — 42 engine constants, 62 entry points, all agree |
| superproject sweep (`scripts/run_all_suites.sh`) | **50/50 suites, 2,113,185 checks, 0 in-suite failures**, exit 0 |
| behave contract tier | 340 scenarios / 1,632 steps, 0 failed |
| lint ratchet | PASS at exactly 100 (no new violations) |
| docs_check / check_doc_paths / check_archive / check_entity_shape / check_wire_accountability | all PASS |

The sweep declares one skip honestly rather than hiding it: `tsn_fuzz` (the
AAF/AVTP field campaign) contributes 0 because tsn-gen is not installed on this
machine.

Submodule pushed before the superproject pin, per CONTRIBUTING 2.1.

Mister-M-alt:
Review of a15deab57b58a9cd77511caf66c24433f095f2e1: not validated yet.

- The new CRF stopped-state gate suppresses KL_crf_rx frame_p_i, which is also the receiver and Table 5.6 monitor strobe. A stopped CRF input therefore stops counting received PDUs and format/sequence events instead of only discarding them from clock-recovery/media processing. Milan v1.2 Sections 5.3.8.7 and 5.3.8.10 require the received traffic to remain observable. Keep the CRF observation/counter path upstream and gate only its media effects, with a stopped-CRF regression.
- The current audit and roadmap still contain mutually contradictory started/stopped claims, including unimplemented/withdrawn, resolved/absent, persisted despite the blank backend, and an unconsumed datapath output. RTL and test comments also alternate between not classified and still matched. Reconcile these current authorities and pin the actual processor commit.

The focused ACMP listener, processor, NxN datapath, and CRF receiver suites pass, but they do not cover the stopped CRF counter behavior above.

Mister-M-alt:
**Review round 2 — two reviewers, cleared context (verification lens + clause/wire lens). Both CHANGES-REQUESTED. Both were right, and one of them caught a regression I had introduced and misdiagnosed.**

### The branch was RED, and it was my newest commit

Fixing the re-bind gap below, I made the Table 5.22 trigger fire on *any*
`f_started` change. That also fires on bind and unbind — where a notification is
already pushed by another term — so every bind put **two** unsolicited frames on
the wire, shifting the AECP response queue and cascading to **71 failures** in
`pp_top`. I had seen those failures and attributed them to a concurrent agent's
mutation. That was wrong; the reviewer measured them to the commit.

The clause scopes it: Section 5.3.8.7 calls started/stopped "undefined when the
Stream Input is not bound", so a bind (undefined → started) and an unbind
(started → undefined) are not started/stopped **changes**. The trigger now
requires the sink bound *before and after*, leaving exactly the two paths that
move the bit under a live binding — the AECP request, and 5.5.3.5.6 step 2's
re-bind with STREAMING_WAIT flipped.

**Nothing would have caught it.** `G7c`–`G7g` now assert a bind pushes exactly
one notification and a STOP pushes exactly one. Re-run against the regression
they produce **one named failure** instead of a 71-failure cascade.

### Blockers and defects fixed

| Finding | Resolution |
|---|---|
| Trigger fired on bind/unbind → duplicate notification, branch red | scoped to a change under a live binding; `G7c`–`G7g` guard it |
| Re-bind (5.5.3.5.6 step 2) flipped started with **no** push | trigger keys on the change, not the walk kind; `S1k`/`S1l`/`S1m` grade it |
| `pre_ready_o` did not reflect the new arbitration arm — an NVM preload could be **silently dropped** | `!strq_pend_r` added |
| `CRF_SNK_IDX_C` aliases to 0 with `crf_sink: false` — a STOP on AAF sink 0 would silence the media clock | guarded on the sink existing |
| VERSION `0x004F` had **no changelog entry** (prose lost when a review agent reverted the file and restored only the value), and the surviving `0x004D` text still said these commands were withdrawn | entry written into the RTL comment and mirrored verbatim into REGISTER_MAP; `0x004D` marked superseded |
| Out-of-range request arm never driven by any test | `S1n` drives it and grades the counter |

### Claims corrected rather than defended

- **IEEE 7.5.2's unsolicited *response* for these opcodes is not sent.** Table
  5.22's GET_STREAM_INFO push is. My comment described the two duties as one.
  Corrected, deferred to #69 with the SET_* family, and recorded in the roadmap
  instead of marking the row done.
- **The holder defers the stall, it does not bound it.** One deep; a second
  request while one is pending still waits. Stated in the RTL.
- **`dbg_strq_drop_o` is observable at the port and in simulation, not to a
  controller.** Plumbing it to a CSR would be a register-map change for a
  counter whose only legitimate value is zero.
- **`W21cc` cannot prove the engine honours `strm_set_ready_i`** — unreachable
  from the wire, because the holder drains at top priority. The block now states
  what it does prove; the listener-side property is graded directly in `S1c2`.

### Bar

processor **14,107 / 0** · uPC gate PASS (42 ↔ 62) · superproject sweep
**50/50, 2,113,185 checks, 0 failures** · behave **340 scenarios** · lint PASS at
exactly 100 · all five doc/shape/wire gates PASS.

Round 2 found blockers, so round 2 is not the sign-off. Requesting round 3.

Mister-M-alt:
Review of 75a9425123062c44827b3b9860dbbe0e3b329a8c: not validated.

- `milan_datapath.sv:4368` still suppresses `KL_crf_rx.frame_p_i` while the CRF Stream Input is stopped.
- That strobe remains the receiver and monitor observation input used for the Milan Table 5.6 event counters and `FRAMES_RX` in `KL_crf_rx.sv`. A stopped Stream Input must suppress media and clock-recovery effects without making received CRF PDUs disappear from its counters.
- The new `crf_snk_stopped_w` guard only protects shapes without a CRF sink. It does not separate PDU observation from clock recovery, and this update adds no stopped-CRF counter regression.

The CRF receive-counter blocker therefore remains.

Mister-M-alt:
**Review round 3 — CHANGES-REQUESTED, and the blocker was again mine.**

### The blocker: I fixed one acceptor and called it a class

The holder arm went in at the **top** of the record walker's X_IDLE priority
chain. Round 2 caught that `pre_ready_o` didn't exclude it; I fixed that one and
wrote a comment saying "every other higher-priority arm was already in this
term". That closed the case in my head instead of checking it. **Three of the
four acceptors were still unguarded.**

`txn_ready_o` is the expensive one. It feeds the dispatch FIFO's pop side, so an
ACMP command at the queue head during the one drain cycle after any START/STOP
is **popped and never serviced** — the controller times out, and since only
`X_CONSUME` frees an RX slot, that slot **leaks permanently**. Four slots,
shared with ADP, MAAP, the talker and AECP; four such events wedge the control
plane until reset.

All four now carry the term. Each is mutation-proven:

| ungated | catches |
|---|---|
| `txn_ready_o` | RV3 |
| `evt_tk_ready_o` | RV5 |
| `pend_clr_pop_w` | RV4 (`collided 0 vs alone 1`) |

RV4 took three attempts and the failures are worth recording: the first
assertion was a loose three-way OR that could not fail; the second had a control
leg but the wrong timing; only putting the expiry and the request on the **same
clock edge** reproduced it. The defect lives in a single cycle.

### A comment of mine denied a path that exists

The trigger comment claimed it "leaves exactly two paths" that move the bit
under a live binding. A **BIND_NEW onto an already-bound sink** is a third: its
ROM cell is `A1 A11 A9 A2 A3 A4 A5` — no A10, so the binding never drops while
A2 rewrites the bit, and A4 already pushes. That put a second GET_STREAM_INFO
frame on the wire for one event. Excluded via `bind_act_r`, guarded by RV8.

### The changelog described the reverted design

Both places, both false since round 2: the "frames become FOREIGN at the parser"
claim (the classifier is not gated — that was the round-2 fix) and the CRF
"KNOWN LIMIT" (the CRF sink has been covered since `f566549c`, which landed
*before* the changelog was written). Corrected, the one-deep holder qualifier
added, and the RTL comment and REGISTER_MAP entry are now verified
**byte-identical** (3,889 chars each) so they cannot drift again.

Also fixed: two stale comments in `milan_datapath.sv` — one still describing the
classifier-gating approach, one still calling the shadow's AECP settings face
"NOT YET CONSUMED" when `aecp_strm_started_o` now has three live consumers.

### Confirmed fixed by the reviewer, with mutations

Notification scoping (all four mutations of the guard fail a test), the CRF
shape guard, `S1n`'s out-of-range drop, and the two changelog texts matching.

### Bar — on the MERGE RESULT

`main-push` had moved 22 commits ahead (PR #90 landed the containment tooling
and a change to `run_all_suites.sh` itself), so the branch bar was not the merge
bar. Absorbed it — clean, no conflicts — and re-ran:

- containment self-test (new, from #90): **PASS**
- processor **14,129 / 0 failing** · uPC gate PASS (42 ↔ 62)
- behave **340 scenarios** · lint PASS at exactly 100
- all five doc/shape/wire gates PASS
- full superproject sweep: running on the merge result; numbers to follow

Round 3 found blockers, so it is not the sign-off. Round 4 next.

Mister-M-alt:
**Bar on the MERGE RESULT** (branch has absorbed `origin/main-push`; 0 commits behind).

`main-push` had moved 22 commits ahead while this PR was in review — PR #90
landed the containment tooling and, notably, a change to `run_all_suites.sh`
itself. So the branch bar was not the merge bar. Absorbed it (clean, no
conflicts) and re-ran everything:

| gate | result |
|---|---|
| `check_merge_containment.py --selftest` (new, from #90) | **PASS** |
| processor suite | **14,129 checks / 0 failing** |
| uPC map gate | PASS — 42 engine constants, 62 entry points |
| superproject sweep | **50/50 suites, 2,113,185 checks, 0 in-suite failures**, exit 0 |
| behave contract tier | 340 scenarios / 1,640 steps |
| lint ratchet | PASS at exactly 100 (no new violations) |
| docs_check / check_doc_paths / check_archive / check_entity_shape / check_wire_accountability | all PASS |

The sweep declares its one skip rather than hiding it: `tsn_fuzz` contributes 0
because tsn-gen is not installed here.

Round 4 review is running against this merged tree, so it reviews what would
actually merge rather than the branch in isolation.

Mister-M-alt:
**Review round 4 — first round with NO functional defect found in the RTL.**

All three round-3 fixes hold under independent mutation:

| term removed | test that fails |
|---|---|
| `!strq_pend_r` from `txn_ready_o` | RV3 |
| `!strq_pend_r` from `evt_tk_ready_o` | RV5 |
| `!strq_pend_r` from `pre_ready_o` | RV6 |
| `!strq_pend_r` from `pend_clr_pop_w` | RV4 (`collided 0 vs alone 1`) |
| `!bind_act_r` from the trigger | RV8b |

The reviewer also enumerated every ROM cell that touches `f_started` and
confirmed **no Table 5.22 push can be missed**: each cell either raises
`act_strt_chg_o` or already pushes the same per-sink level from its own term.
And it independently re-derived the `X_WB` staleness argument — three entries,
thirteen `cellmut_r` sites, both loaders — and found the comment's count exact.

### What was asked for, and done

**The submodule's architecture of record still said this command was
unimplemented** and pointed readers at an abandoned branch —
`docs/architecture/06_aecp_engine.md`, in three places: the F06.14 command
table, the Table 5.22 trigger inventory, and the served-command list. That file
is what the next person copies from to derive a dispatch table, so it is the
same "two copies of one fact drift silently" shape `check_upc_map.py` exists to
prevent, one layer up. All three corrected.

### What was asked for, and deliberately not done

The same finding listed five **superproject** docs. All five are
`[OBSOLETE + 2026-08-16]`-bannered, and this lane's stated rule — applied two
rounds ago to `PDU_GETTER_SETTER_VERIFICATION.md`, and enforced by
`scripts/check_archive.py` — is that obsolete pages stand as the historical
record. I edited four of them before checking the banners, then reverted.
Rewriting them would destroy the record and contradict the call already made.

### Claims scoped rather than defended

- **CRF sink.** The direction that matters for every shipping build — the gate
  does NOT fire in normal operation — is held by 20+ milan_dp checks that go red
  if the predicate is forced true. The **stop** direction is not exercised: no
  suite binds and then stops that sink, because the sim carrying the AECP
  transactor does not inject CRF receive traffic. The changelog now says the
  predicate *reaches* it, not that the stop has been *demonstrated*.
- **Persistence.** Two of Section 5.3.8.7's three shalls land here. The third
  ("saved in a non-volatile memory and restored after a power cycle") does not:
  `NVM_BACKED_C = 1'b0` behind a blank-flash stub (issue #70). Now stated in the
  changelog, which is the artifact a reader consults first.
- **`bnd_was_r`** is redundant — removing it leaves both suites green. Kept
  because it states the clause's condition directly rather than relying on the
  ROM's action lists to imply it, and labelled as belt-and-braces.
- **`lstn_strq_drop_w`** is unreadable above the module: no CSR, no probe. The
  arm it counts is driven and graded in `tb/acmp_listener` (S1n); the net is
  labelled with what it costs and where to bring it out if it ever fires.

### Bar

processor **14,129 / 0** · uPC gate PASS (42 ↔ 62) · sweep **50/50, 2,113,185
checks** · behave **340 scenarios** · lint PASS at exactly 100 · all five
doc/shape/wire gates PASS. Both changelog copies verified **byte-identical**.

Round 5 running on this tree.

PR #93, opened 2026-08-17
Complete ADD/REMOVE_AUDIO_MAPPINGS acceptance

### LINKED ISSUE #68: AECP: ADD_AUDIO_MAPPINGS / REMOVE_AUDIO_MAPPINGS unimplemented — dynamic input mappings are mandatory for a Milan listener and the validation is all-or-nothing

Mister-M-alt:
Closes #68

## Summary

- completes atomic ADD_AUDIO_MAPPINGS and REMOVE_AUDIO_MAPPINGS behavior for dynamic Stream Port Input and Output maps
- enforces descriptor, geometry, conflict, duplicate, running-output, and full-command rollback rules
- accepts every output cluster published by the generated model, including fabric-disabled source templates
- retains protocol ownership separately from CMAP media enable so disabled mappings remain visible and conflict-safe without enabling an unavailable source
- serializes live ACMP and AECP dispatch heads through round-robin scoreboard admission until matching RX-slot retirement
- reserves every target Stream Output after the phase-1 recheck so ACMP, SRP, and local bypass starts cannot race phase-5 mapping writes
- exports the authoritative entity-lock level and refuses local CSR mapping writes while LOCK_ENTITY is held
- preserves physical map RAM and protocol ownership stores throughout the locked interval
- makes every generated Stream Port Input dynamic per Milan v1.2 5.3.3.9
- rejects mixed static and dynamic Stream Port Output images because the capture selector is global
- enforces the IEEE command cdl boundary: 63 records accepted, 64 rejected atomically
- returns NOT_SUPPORTED for static Stream Port Output GET_AUDIO_MAP requests
- reconciles the current audit, channel-map guide, traceability, builder output, and obsolete gaps record

Root head: `9305332c2356d6905e49836edc44a7a3931fd504`

Processor dependency: https://github.com/Mister-M-alt/protocol-processor-control-plane-avb-milan/pull/2

Processor head: `848801d31ae8415cfc965fdd7429dba930eb27de`

## Validation

- all 27 protocol processor suites: 14,010 checks PASS
- protocol processor pp_top: 982/982 PASS
- protocol processor ACMP talker: 807/807 PASS
- protocol processor uCPU: 304/304 PASS
- protocol processor zero-tolerance HDL lint: PASS
- root pp_shadow: 273/273 PASS
- root 8x8 datapath integration: 4,311/4,311 PASS
- root nine-shape datapath matrix: 7,618 checks PASS
- root RTL lint ratchet: PASS at 99 findings
- Behave contracts: 15 features, 333 scenarios, 1,584 steps PASS
- builder gates: PASS
- documentation gate: 0 findings across 204 files
- path gate: 1,087 paths and 95 anchors PASS
- archive and TOC gates: PASS
- processor portability synthesis: PASS

## Remaining compliance boundary

Mapping nonvolatile save and restore is not claimed here. Issue #70 owns that backend and replay work and remains a release blocker for Milan v1.2 Section 5.3.10.1 and test-plan item es-5.1.


Mister-M-alt:
Review of b7d0bbcd60eb485af4004fff81056fe26afa829c: not validated yet.

- Milan v1.2 Section 5.3.3.9 requires every Stream Port Input to use dynamic mappings, but endstation_arty_current still emits ADP_DMAP_IN_MASK_C as zero and returns NOT_SUPPORTED.
- The builder advertises per-port static/dynamic Stream Port Output policies, while milan_datapath reduces each mask to a global OR. A mixed configuration therefore selects the unseeded dynamic capture map for static outputs and does not preserve cross-port ownership.
- IEEE 1722.1-2021 Section 9.2.2.6 caps command control_data_length at 524. For ADD/REMOVE_AUDIO_MAPPINGS, 20 + 8N makes 63 records the standards boundary. The 68-record test is not a legal command-boundary test.
- Current documentation still contains conflicting suite counts, static-only channel-map wording, and stale builder comments.

The focused processor, Behave, builder, and diff checks pass, but these compliance and contract issues block validation until corrected and re-reviewed.

Mister-M-alt:
PR #93 is validated at root head `9305332c2356d6905e49836edc44a7a3931fd504`, base `139204e06c546670aa3a9ea33582c638de77bfca`, with processor pin `848801d31ae8415cfc965fdd7429dba930eb27de`.

The issue #68 live mapping scope is validated: generated geometry, atomic ADD and REMOVE, GET visibility, disabled fabric templates, ownership conflicts, running-output exclusion, entity-lock exclusion, notifications, and command-size boundaries.

Local validation passes 14,010 processor checks, 7,618 root datapath checks, 333 Behave scenarios with 1,584 steps, the RTL lint ratchet, builder gates, and all documentation gates. Mapping persistence remains tracked separately by #70 and is not claimed by this PR.

Mister-M-alt:
**Rebased onto `main-push` and the two collisions fixed.** Branch was `CONFLICTING/DIRTY` since #92 merged; it is `MERGEABLE` and merges cleanly now.

**ROM collision.** `E_AMADD` spans 34 words, so at 1632 it overlapped **both** `E_STOP` (1632..1641) and `E_STRMNS` (1664..1668), which #92 landed. Moved to **1712**, `E_AMREMOVE` to **1760** — the contiguous free run above `E_STRMBAD`. `place()` asserts on overlap so this could never have shipped silently, but it would have failed ROM generation on merge. `check_upc_map.py` passes: 44 engine constants, 64 entry points, all agree.

**VERSION.** `0x004F` is taken by #92. Renumbered to **`0x0002_0050`** — `milan_csr.sv`, the four Verilator pins, README, roadmap status line, and the register-map entry. The changelog chain now reads `0x0050` (mappings) → `0x004F` (started/stopped) → `0x004E`, each entry once, and the RTL comment and `REGISTER_MAP.md` copies are verified **byte-identical** (539 chars).

**The conflict that needed judgement.** `tests/features/aecp_response_contract.feature`: each branch still carried the rows the *other* removed — this one had 34/35, `main-push` had 44/45. Both features are served, so the resolution takes **neither** side. That table's own comment warns never to leave a served opcode in it.

**Bar on the merge result:** processor **14,168 checks / 0 failing**; uPC gate PASS; superproject sweep **50/50, 2,113,729 checks, 0 failures**; behave 339 scenarios; lint **99 ≤ 99** (the ratchet you tightened composes with #92, which added no violations); all doc/shape/wire gates PASS.

README's served count follows both authorities — the engine's `OP_*_C` constants and the behave `SERVED` inventory both say 25 here.

Mister-M-alt:
PR #93 is validated at root head 75ba5da4089d3f46637fb5ca01b008d8d0c33b77 with processor pin 2cdf0a1e0161de68e585cc78a05dd94f6204d651. The atomic audio mapping implementation, integration behavior, active documentation, verification evidence, and dependency state are validated. Mandatory local suites and documentation gates pass.

PR #95, opened 2026-08-17
Implement Milan GET_DYNAMIC_INFO batching

### LINKED ISSUE #71: AECP: GET_DYNAMIC_INFO (0x004B) unimplemented — a Milan SHALL that test plan v1.9 never sends, and the one command the ITER/APPEND exemplars were written for
## Summary

`GET_DYNAMIC_INFO` (`0x004B`) is a Milan v1.2 SHALL (§5.4.2.29, added in Milan 1.1) and is implemented by PR #95.

It was the lowest-priority open SHALL because the end-station validation test plan v1.9 never sends `0x004B`. Grepping the complete plan finds no test item that issues it. PR #95 closes the standards debt even though the current compliance run does not exercise the command.

It is also the command with the most interesting constraints, and the µISA already
has every operation it needs.

## What it is

One command that packs several fixed-size getters and returns their answers in one
response. IEEE §7.4.76.1, Figures 7-93/7-94 — each element:

| Offset | Field | Bytes |
|---|---|---|
| +0 | `info_command_specific_data_length` | 2 |
| +2 | reserved | 2 |
| +4 | `info_status` | 1 |
| +5 | reserved | 1 |
| +6 | `info_command_type` | 2 |
| +8 | `info_command_specific_data` | *L* |

Response `cdl` = `12 + Σ(8 + Lᵢ)`.

Note `info_command_type` is a full **16 bits** here, unlike the 14-bit field in
the AECPDU header (IEEE 1722.1-2021 split off a `cr` bit; this field did not
follow).

## The rules that make it awkward

**Only fixed-size getters are allowed.** The whitelist:
`GET_CONFIGURATION`, `GET_STREAM_FORMAT`, `GET_VIDEO_FORMAT`, `GET_SENSOR_FORMAT`,
`GET_STREAM_INFO`, `GET_NAME`, `GET_ASSOCIATION_ID`, `GET_SAMPLING_RATE`,
`GET_CLOCK_SOURCE`, `GET_SIGNAL_SELECTOR`, `GET_COUNTERS`,
`GET_MEMORY_OBJECT_LENGTH`, `GET_STREAM_BACKUP`.

Explicitly **forbidden**: `GET_CONTROL`, `GET_MIXER`, `GET_MATRIX`,
`GET_AVB_INFO`, `GET_AS_PATH`, `GET_AUDIO_MAP`/`VIDEO_MAP`/`SENSOR_MAP` — the
variable-length ones. A command containing any of these is `BAD_ARGUMENTS` and
**nothing is processed**.

**Per-element status**, not one verdict: each element carries its own `info_status`
from Table 7-141, so a packed `GET_SAMPLING_RATE` on a missing Audio Unit reports
`NO_SUCH_DESCRIPTOR` in *its* element while its neighbours succeed.

**Skip on overflow**, silently: an element that would push the response past the
AECPDU limit is **omitted**, and the rest keep processing. Not truncated, not an
error.

**And the hard one:**

> If an implementation could return the status code IN_PROGRESS to any of the
> subcommands packed into the GET_DYNAMIC_INFO command, then the implementation
> shall not support the GET_DYNAMIC_INFO command.

This engine's policy is never to answer `IN_PROGRESS` (06 §8's deadline arm forces
a real answer inside 240 ms), so the rule is satisfiable — but it is a standing
constraint on everything else, and worth a comment in the µcode so a future
`IN_PROGRESS` arm does not silently make this command non-compliant.

```mermaid
flowchart TD
    C[GET_DYNAMIC_INFO, K elements] --> W{every info_command_type on the fixed-size whitelist?}
    W -->|no| BA[BAD_ARGUMENTS - nothing processed]
    W -->|yes| L[for each element]
    L --> F{would it overflow the AECPDU?}
    F -->|yes| SK[omit it silently, keep going]
    F -->|no| R[run the getter, write its own info_status]
    R --> L
    SK --> L
    L --> D[SUCCESS, cdl = 12 + sum of 8 + Li]
```

## Why the µISA already fits

Nothing new is needed:

| Need | Op | Already used by |
|---|---|---|
| iterate the element list | `ITER_OPEN` / `ITER_NEXT` | `E_GCTRS`'s zero-body arm, `E_GAMAP` |
| append with skip-on-overflow | `APPEND` | `E_OVF`, and it is the **only** op that honours the 524-byte cap |
| per-element status | `SET_STATUS` into the element header | — |
| reject the whole command | `CHECK_ARG` with the NOT_SUPPORTED/BAD_ARGUMENTS arm | `E_RDESC`'s configuration-index check |

`E_ITER` (128) and `E_OVF` (416) are exemplar programs placed since the beginning
and never dispatched to. This command is what they were written for.

The sub-getters it must call are all landed or in flight: `GET_CONFIGURATION`,
`GET_SAMPLING_RATE`, `GET_CLOCK_SOURCE`, `GET_STREAM_FORMAT`, `GET_STREAM_INFO`
and `GET_COUNTERS` are served today; `GET_NAME` is a separate open issue. The
sensible scope is **the subset this entity implements**, with every other whitelisted type answering `NOT_SUPPORTED` in its own element, as required by IEEE 1722.1-2021 §7.4.76.1.

## The interop hazard worth knowing

`info_command_specific_data` is *"the complete AECP Message Specific Data buffer"*,
i.e. everything from octet 24 of what the standalone response would have been. For
**`GET_STREAM_INFO`** that buffer is **56 octets under Milan** (the 80-byte Figure
5.1 form) but **84 under base IEEE 1722.1** (the 108-byte form). A packed-response
encoder that uses the IEEE length will be mis-parsed by a Milan controller and vice
versa. Use the Milan length — this device serves the Milan form.

## Blocks

Nothing in test plan v1.9. Milan §5.4.2.29 only.

## How we prove it

**Desk (`protocol-processor/tb/pp_top`, a new §X)**
- A two-element command (`GET_SAMPLING_RATE` + `GET_CLOCK_SOURCE`) byte-exact,
  including each element's length, `info_status` and its own payload.
- Per-element status: one element naming a descriptor that exists and one naming a
  descriptor that does not; the response is `SUCCESS` with
  `NO_SUCH_DESCRIPTOR` in the second element only.
- The whitelist: a command containing `GET_AUDIO_MAP` answers `BAD_ARGUMENTS`, and
  the mapping store is proven **untouched** — *"nothing processed"* means nothing.
- Skip-on-overflow: enough elements to exceed the cap; the response stops early,
  the omitted elements are simply absent, and the surviving ones are intact.
- The `GET_STREAM_INFO` element carries **56** bytes, not 84.

**Contract (`tests/`)**
- Move `0x004B` from the open-gap Examples table into `SERVED`.

Mister-M-alt:
## Status

Mandatory local validation passes. PR #93 is merged, and this branch now targets `main-push` directly.

## Exact revisions

- root head: `12d3f3fcafda8bc6540a12639b71d5160654607f`
- integrated base: `efa231abb79eba5e19f92c84851c84bf7a6af865`
- protocol processor: `91f3b7c132a0074f6d630a438d90604c230d380d`

## Description

- Implements Milan v1.2 section 5.4.2.29 and the IEEE 1722.1-2021 section 7.4.76 fixed-get subset.
- Rejects forbidden types, structurally invalid lists, and oversized commands at the command level.
- Reports a non-success command-side `info_status` as record-local `BAD_ARGUMENTS`.
- Reports status independently per record and copies nonzero command data for legal unsupported getters.
- Checks each getter's actual response cursor against its selected fixed response length.
- Prevents a four-byte copied record from emitting a second word.
- Starts aggregate response length empty, silently omits a result that would exceed 524 bytes, and continues with later records.
- Emits the Milan 56-byte `GET_STREAM_INFO` record body.
- Integrates the audio mapping implementation merged through PR #93.

## Local validation

- protocol processor: 14,193/14,193 checks
- protocol processor `pp_top`: 1,094/1,094 checks
- protocol processor Yosys: all targets pass, including the Xilinx engine target
- root Verilator: 50/50 suites, 2,113,729 accounted checks, zero in-suite failures
- root Yosys: 44/44 tops, tied-input gate pass, observer-purity gate pass
- Behave after integration: 15 features, 338 scenarios, 1,615 steps
- RTL lint: 99 findings, within the 99-warning ratchet
- entity shape: 121 checks
- wire accountability: 79 checks
- documentation: 0 findings across 204 Markdown files; path, archive, anchor, and TOC gates pass

## Reproduce

```bash
git fetch origin 71-get-dynamic-info
git switch --detach origin/71-get-dynamic-info
git submodule update --init --recursive

protocol-processor/scripts/run_suites.sh
protocol-processor/syn/yosys/run.sh
scripts/run_all_suites.sh
syn/yosys/run.sh
behave tests/features
python3 scripts/docs_check.py
```

## Definition of Done

- [x] Exact 13-command fixed-get whitelist
- [x] Whole-command structural validation before execution
- [x] Record-local handling of command-side status
- [x] Command and response length boundaries
- [x] Per-record status and unsupported-data copying
- [x] Getter cursor and short-copy guards
- [x] Overflow omission with continued processing
- [x] Milan `GET_STREAM_INFO` record length
- [x] Root contract and capability documentation
- [x] Mandatory local simulation, synthesis, structural, and documentation gates

Closes #71


Mister-M-alt:
PR #95 is validated at root head `f178b2f1e90d051d1201ebebc639a77b6b626802`, stacked base `9305332c2356d6905e49836edc44a7a3931fd504`, with processor pin `5126e3f3bb26abb7e17d3890bf0b2048a14b1b26`.

The implementation matches Milan v1.2 section 5.4.2.29 and IEEE 1722.1-2021 section 7.4.76: exact 13-command whitelist, complete prevalidation, full-byte command-status validation, independent record results, legal unsupported-data copying, exact length boundaries, overflow omission with continued processing, and the Milan `GET_STREAM_INFO` body.

Local validation passes 14,030 processor checks, 2,113,684 accounted root Verilator checks across 50 suites, 44 root Yosys targets, 332 Behave scenarios with 1,573 steps, and all lint, structural, wire-accountability, and documentation gates. Two independent exact-head reviews found no blockers.

Mister-M-alt:
**Review: two reviewers with cleared context (clause/wire-format, and RTL/test-adequacy), plus an integration pass from the #92 branch.**

Verdict: **CHANGES-REQUESTED**, but narrowly. No wire-visible defect was found. Every field offset, length and cdl computation checks out against the figures, and the clause reading is right on the point most likely to be wrong. What needs attention is that the two mechanisms this feature is built on are not falsifiable by the suite that ships with it.

## The clause reading is sound

Verified against the PDFs, not the code's restatement:

- **"Silently omit an over-long result and continue" is exactly what 7.4.76.1 mandates** — *"the entity shall not append this dynamic_info response to the response buffer and shall continue processing the remaining dynamic_info structures"*. Not truncate, not stop.
- Figure 7-94 offsets measured against the figure's own bit ruler: length @+0, reserved @+2, `info_status` @+4, reserved @+5, `info_command_type` @+6, data @+8. The RTL, the C++ `direc()` and the Python builder all agree.
- The allowed set is **exactly** the thirteen fixed-size codes of 7.4.76.2, in Table 7-140 order, with `0x004B` correctly excluded so a batch cannot nest.
- The 524 cap is applied to the right quantity (9.2.2.6 counts octets after `target_entity_id`), and `cdl_w = 12 + pld_r` closes the loop.
- The **56-byte** GET_STREAM_INFO record body is right: Milan 5.4.2.10 replaces IEEE Figure 7-40 with Figure 5.1, which ends at byte 79.
- Milan 5.4.2.29 adds no narrowing beyond that.

## Worth resolving before merge

**1. Whole-list rejection on a non-zero command `info_status` is over-scoped.** 7.4.76.2 names exactly one atomic-rejection trigger — a forbidden `info_command_type` — and otherwise says *"handle each element of the dynamic_infos array as if it were an independent command."* 7.4.76.1's *"Always SUCCESS for a Command"* states a field value; it prescribes no entity behaviour. The truncated-header and data-overrun rejections **are** forced, because you cannot locate the next record. A bad `info_status` leaves the array perfectly parseable — and this PR already answers the analogous case (a record whose declared length is too short for its getter) with a **per-record** `BAD_ARGUMENTS`. Two equally-parseable malformations get opposite scopes. Either narrow it to per-record, or re-word the comment as this engine's convention rather than a requirement.

**2. The per-getter length table is a hand-copied mirror with no cross-check.** `gdi_subcommand_decode` hardcodes 4/12/56/8/8/136 and `A_GHDR` commits that into the record header *before* the getter runs, while `A_GPATCH` advances the cursor from the µCPU's actual `resp_len_w`. Nothing compares them. Every reachable µprogram was hand-traced and they all match today — but one added `BUILD_FLD` in E_GSTRI silently misaligns every following record and puts stale response RAM on the wire, while the standalone GET_STREAM_INFO stays byte-perfect and every existing test stays green. This repo's own rule is *derive, never mirror a constant*. Emit the table from `gen_ucode.py`, or void the response when `g_sub_end_r != g_out_r + 8 + g_rec_rlen_r`.

**3. `E_GSRATE`'s hit arm writes four bytes past its own record.** `OP_COPY_BUF` writes unconditionally at `eseq_r==2`, putting a spurious word at the next record's header. It is harmless only because `A_GHDR` phase 0 rewrites those bytes with `wstrb 4'hF` — correctness by coincidence between two modules, and untested: W8b uses a *miss*, W8l sends empty data. Add a batch with `GET_SAMPLING_RATE(0x0002, 0)` (hit) followed by another record, and one where it is last.

## Tests that cannot fail

Concrete surviving mutations, each named by the reviewer:

- **W8d** sends four byte-identical GET_COUNTERS records, so skipping record 3 instead of 4 produces an identical frame. Use distinct `{type,index}` pairs — `ctr_mask`/`ctr_value` already return different bodies.
- **The `A_GCOPY` data path is graded only against all-zero command data**, so a *missing* write is caught (the RX memory is poisoned) but a *wrong source address* is not: changing `24 + g_rd_pos_r` to `24 + g_data_start_r` passes every fixture. Give the GET_NAME record a distinctive non-palindromic payload.
- **Batched GET_CONFIGURATION is graded against `CFGIX == 0`** — four zero bytes. `gen_ucode.py`'s own E_GCFG comment records this exact trap, and W3b answers it for the standalone command by patching the image to `0x0007`; that falsifier is not repeated for the batch.
- **The full-body wrong-target refusals** (`UPC_TIZ8NS_C`/`TIZ4NS_C`/`GCTRSNS_C`) are never sent inside a batch — the one combination where a non-zero patched status meets a fixed 12/8/136-byte body.

## Smaller

- The 524 limit is mirrored three times (`A_GSCAN`, `A_GDEC`, `ucpu_pkg::RESP_CAP_C`) and must agree to the byte; reference the constant.
- The omission threshold is applied to the *response*, which Milan 5.4.1 explicitly frees from 9.2.2.6 — legal, but it will omit records it is permitted to send.
- The `524` literal is not tied to `RESP_BUF_C`; a build with a smaller `DESC_LINE_BYTES_P` could run past the buffer with `cdl` still claiming the bytes. One `$error` would close it.
- A per-record gather-face fault collapses the whole batch to `ENTITY_MISBEHAVING` rather than failing that record.
- `disp_batch_i`'s `rb_we_o = !batch_r` effect is unobservable — BUILD_HDR's bytes are below `HDR_BYTES_C` and already dropped. Assert the drop count or remove the line and the doc claim.
- `NOT_SUPPORTED` vs `NOT_IMPLEMENTED` for a permitted-but-unimplemented getter: the literal 7.4.76 reading was taken, which means the same opcode answers differently inside and outside a batch. Defensible; worth recording as a decision.

## Integration — from the branch that conflicts with this one

**#92 has merged**, so `main-push` is now VERSION `0x004F`. This PR takes `0x0050` on the assumption #93 gets `0x004F` — that is no longer available. **#93 needs `0x0050` and this PR `0x0051`**, plus the four Verilator VERSION pins, README, roadmap and register-map entries in each.

This PR also still targets `68-aecp-audio-mappings`, so it is gated behind #93 and needs retargeting to `main-push` after it lands, as its own description says.

Two things I checked that are **fine**: the lint ratchets compose (#93 removes the `tctx_chans_r[...[6:4]]` WIDTHTRUNC in a region #92 does not touch, so the merge lands at 2 ≤ 2), and the `tests/steps/` additions are genuinely orthogonal despite both PRs touching the `SERVED` inventory — different opcodes, different helpers, different code paths.

## Verified correct, so it is on the record

No unbounded stall or deadlock; the backward `info_status` patch modifies exactly one octet and W8b would fail on a strobe-ignoring buffer; no inferred latch, missing reset, unique-case gap or out-of-range index; the pre-scan cannot read past the committed slot; `IN_PROGRESS` is unreachable; and W8b/W8c2/W8j2/W8j3/W8k/W8l/W8m/W8n are all real discriminating checks with per-record status graded independently of the aggregate.

Mister-M-alt:
**Status check — no change on this branch since the review (head still `f178b2f1`), but the ground under it moved.**

#92 merged into `main-push` (`56ee1e0f`), followed by its documentation follow-up #96 (`a339692e`). That has three consequences for this PR, one of which is a hard build failure rather than a merge nuisance.

### 1. BLOCKER for the stack — two microprograms claim the same ROM words

`main-push`'s processor pin now carries #78's entry points, and they overlap `ADD_AUDIO_MAPPINGS` byte for byte:

| | ROM words |
|---|---|
| `E_AMADD` (from #93, inherited here) | **1632..1665** (34 words) |
| `E_STOP` (on `main-push`) | **1632..1641** |
| `E_STRMNS` (on `main-push`) | **1664..1668** |

`E_AMADD` collides with both. `E_STRT` 1600..1609 and `E_STRMBAD` 1696..1700 are also now occupied, so the tail is no longer the open run it was when these were allocated.

The good news is that this cannot ship silently: `place()` asserts on overlap, so the merged tree fails at ROM generation, and `check_upc_map.py` independently gates the engine's `UPC_*_C` constants against the entry points. Two gates, both loud.

After the merge the free runs are **1335..1455** (121 words) and **1701..2047** (347). `E_AMADD` needs 34, so either fits — moving it and `E_AMREMOVE` above 1700 keeps the tail contiguous.

### 2. VERSION — `0x004F` is taken

`main-push` is at `0x0002_004F`. #93 claims the same, so **#93 needs `0x0050` and this PR `0x0051`**, along with `milan_csr.sv`'s changelog entry, `docs/reference/REGISTER_MAP.md` (which must stay byte-identical to the RTL comment), README, the roadmap, and the four Verilator VERSION pins.

### 3. #93 is now CONFLICTING/DIRTY

It went dirty the moment #92 landed. Superproject conflicts: `milan_csr.sv`, `REGISTER_MAP.md`, `MILAN_V12_ROADMAP.md`, `tb/verilator/milan_dp/sim_nxn.cpp`, `tests/features/aecp_response_contract.feature`, and the submodule pin. In the processor: `KL_aecp_engine.sv` and `gen_ucode.py`.

Two of those are worth calling out because the resolution is not "take one side":

- **`aecp_response_contract.feature`** — #92 removed rows 34/35 from the not-implemented table and #93 removes 44/45 from the same contiguous block. Textual adjacency, not a semantic clash: the resolution is to remove **both** pairs. Restoring either would put a served opcode back in the unimplemented table, and the table's own comment says never to leave a row there passing for the wrong reason.
- **the submodule pin** — both branches moved it, and the two processor branches have genuinely diverged. This needs the processor branches merged, not a pin chosen.

This PR still targets `68-aecp-audio-mappings`, so none of it is actionable here until #93 is resolved and this is retargeted to `main-push`.

The review findings from the earlier comment stand unchanged.

Mister-M-alt:
**Restacked on the fixed #93 and renumbered.**

**ROM collision** — inherited from #93 and fixed there: `E_AMADD` moved 1632 → **1712**, `E_AMREMOVE` 1680 → **1760**, off the words `E_STOP` and `E_STRMNS` occupy. GET_DYNAMIC_INFO itself adds no microprogram entry points (it is engine-side), so nothing else moved. `check_upc_map.py`: 44 constants, 64 entry points, agree.

**VERSION** — `0x0050` went to #93, so this is now **`0x0002_0051`**: `milan_csr.sv`, the four Verilator pins, README, roadmap status, register-map entry. Chain reads `0x0051` (GET_DYNAMIC_INFO) → `0x0050` (mappings) → `0x004F` (started/stopped) → `0x004E`, each once, with the RTL comment and `REGISTER_MAP.md` copies **byte-identical** (350 chars).

**The `tests/steps/` merge was done semantically, not as a marker sweep** — this repo's rule, and the reason is on the record: naive unions broke main twice. The three features' additions are genuinely orthogonal (different opcodes, helpers and code paths), so the resolution is a deliberate union. Verified afterwards that all five new opcodes survive in the `SERVED` inventory: `0x0022`, `0x0023`, `0x002C`, `0x002D`, `0x004B` — 26 rows, matching the engine's 26 `OP_*_C` constants exactly, which is what the behave inventory gate checks.

`aecp_response_contract.feature` took neither side of its conflict again: this branch carried rows 34/35, #93 carried row 75, and all three are served.

**Bar on the merge result:** processor **14,188 checks / 0 failing**; uPC gate PASS; superproject sweep **50/50, 2,113,729 checks, 0 failures**; behave 338 scenarios; lint **99 ≤ 99**; all five doc/shape/wire gates PASS.

Still targets `68-aecp-audio-mappings`, so it needs retargeting to `main-push` once #93 lands. The review findings from my earlier comment are unaffected by this rebase and still stand.

Mister-M-alt:
PR #95 is validated at root head `12d3f3fcafda8bc6540a12639b71d5160654607f`, base `efa231abb79eba5e19f92c84851c84bf7a6af865`, and protocol-processor pin `91f3b7c132a0074f6d630a438d90604c230d380d`.

The requested review corrections are complete: command-side status is record-local, getter response length is checked against the selected record shape, four-byte copies cannot spill into the following record, limit handling uses the shared 524-byte cap, and the regression cases use discriminating data and targets.

Mandatory local validation passes: processor 14,193 checks with 1,094 in `pp_top`; root Verilator 50/50 suites with 2,113,729 accounted checks; root Yosys 44/44 tops plus tied-input and observer-purity gates; Behave 15 features, 338 scenarios, and 1,615 steps; lint, structural, wire-accountability, and documentation gates all pass. Two independent exact-head reviews validate the functional and integration result.

Issue #98, opened 2026-08-18
Documentation consistency: align active claims and add drift detection


Mister-M-alt:
## Objective

Keep repeated implementation-status claims in active documentation consistent
with the RTL and firmware, including claims written in prose outside the marked
feature-status rows.

## Foundation already delivered

- [x] A machine-readable feature ledger exists.
- [x] Stable feature identifiers and supported status values are checked.
- [x] The feature-status checker rejects duplicate/unknown identifiers and
      disagreement inside marked rows.
- [x] The checker runs in the documentation workflow without network access.

## Open findings (reconciled on `dev`, 2026-08-23)

1. RTL and the canonical register map report VERSION `0x0002_0054`, while active
   prose still reports `0x0002_0053` in `README.md`,
   `docs/MILAN_V12_ROADMAP.md`, `docs/reference/REGISTER_MAP_CLASSES.md`, and
   `docs/testing/MILAN_V12_AUDIT_2026-08-16.md`. Repeated served-operation counts
   can drift for the same reason: they are prose outside the marked rows.
2. `docs/integration/BAREMETAL_FIRMWARE.md` still describes AEM identity/CRC as
   gating the PTP clock, while firmware and CSR reset make the PHC/fabric-gPTP
   path live independently before `load_aem_image()`.

The existing documentation gates are green despite both contradictions, so
correcting prose alone is insufficient.

## Work

- Correct all active VERSION, served-operation-count, and boot-order claims.
- Extend the consistency mechanism to the repeated prose values above, using a
  structured marker or generated fragment rather than another spelling scan.
- Keep historical and correctly archived pages out of the active comparison.
- Preserve explanatory prose, but source every repeated status/value from one
  canonical fact.

## Acceptance criteria

- [ ] Active documentation agrees on the current VERSION, served operation
      inventory, PTP/AEM boot order, CRF behavior, persistence, notifications,
      and started/stopped behavior.
- [ ] Changing one active VERSION or operation-count claim while leaving the
      canonical value unchanged makes the checker fail and names every
      conflicting file.
- [ ] Reversing the documented PTP/AEM boot order without changing firmware
      makes a focused check fail.
- [ ] Duplicate identifiers, unknown statuses, and missing ledger references
      remain rejected.
- [ ] Archived pages remain ignored only when they carry the required archive
      banner and successor link.
- [ ] All documentation, path, archive, contents, and wording gates pass.

Related: #94 covers wording/link hygiene; implementation tickets remain the
[...]

Issue #97, opened 2026-08-18
START/STOP_STREAMING: wait for record commit and preserve CRF observation


Mister-M-alt:
## Summary

PR #92 merged into `dev` at `56ee1e0fe296277ccca67225ff0ab2ffd0d0cff6`
and closed issue #78, but the merged START/STOP_STREAMING path still has two
functional correctness defects. The existing suites and CI are green because
they do not observe state at the response boundary and do not inject CRF traffic
while the CRF Stream Input is stopped.

This task fixes the merged implementation forward. It does not reopen the
single-source-of-truth decision: `f_started` in the ACMP binding record remains
the only writable started/stopped state.

Exact merged revisions:

- root merge: `56ee1e0fe296277ccca67225ff0ab2ffd0d0cff6`
- PR head: `2616f3a4c5aabdf767e3c5c7c3b15a41d10c4900`
- protocol processor: `2d375fbddcff32489ca00ec0bc400f4a2ca1427e`

## Defect 1: SUCCESS can precede the state commit

The AECP engine treats `strm_set_ready_i` as completion of `WRITE_ST`. The ACMP
listener currently asserts `strm_set_ready_o` whenever its one-entry holder is
free. That handshake only captures the request into `strq_*`; the record walker
applies and commits `f_started` later when it returns to `X_IDLE`.

The START/STOP microprogram proceeds directly from `WRITE_ST` to `SET_STATUS`,
builds the response, and sends `SUCCESS`. Under a busy walker, the response can
therefore leave while `GET_STREAM_INFO`, `strm_started_o`, and the datapath still
expose the old state.

The implementation needs separate acceptance and completion semantics:

- holder acceptance may remain decoupled from walker availability;
- command completion must acknowledge a committed transition or a confirmed
  no-op after the record has been examined;
- `SUCCESS` must not be emitted before that completion;
- the completion path must be bounded or fail safely so a broken walker cannot
  wedge the AECP plane indefinitely;
- the existing ACMP, timer, preload, and event arbitration guards must remain
  intact.

## Defect 2: stopping CRF suppresses observation, not only consumption

The root connects `KL_crf_rx.frame_p_i` to
`avtprx_parse_p && !crf_snk_stopped_w`. Inside `KL_crf_rx`, `frame_p_i` is the
source for frame matching, validation events, observation-interval flags,
FRAMES_RX, and receive timeout handling. A stopped CRF Stream Input therefore
makes received CRF traffic invisible to the diagnostic path.

[...]

Issue #74, opened 2026-08-16
Media clock: select CRF and align the audio grid


Mister-M-alt:
## Objective

Make the CRF Media Clock Input selectable and prove that the physical audio
frame grid remains aligned with the packet media grid.

Milan clauses: 5.3.11.1 and 7.2.2.

## Current state (reconciled 2026-08-23)

`SET_CLOCK_SOURCE` is no longer a blocker: the protocol processor accepts and
stores it, and the live `pp_aecp_clk_src_index_w` value reaches
`milan_datapath`. The remaining gap is the consumer. The root still fixes
`CRF_CLK_SELECTED_C` to zero, so the media NCO, MMCM servo, and restart gate
cannot follow the selected CLOCK_SOURCE.

The true-ratio simulation also measures the TDM frame clock about 10.6 ppm
below the exact 48 kHz packet grid and currently passes by proving that the
grids differ. The implementation must first record the phase/frequency contract
between `clk_tdm_i`, `clk_audio_i`, and `media_tick_p`; that design decision is
part of this ticket, not an external dependency.

## Work

- [ ] Replace `CRF_CLK_SELECTED_C` with the live CLOCK_DOMAIN source selection.
- [ ] Route that selection to both the MMCM servo gate and media NCO gate.
- [ ] Define the phase and frequency relationship between `clk_tdm_i`,
      `clk_audio_i`, and `media_tick_p` in INTERNAL and CRF modes.
- [ ] Remove the independent-divider drift or add a bounded phase-alignment
      mechanism.
- [ ] Gate media-clock-restart propagation and `mr` on the selected CRF source.
- [ ] Preserve clean INTERNAL free-running behavior.

## Proof and acceptance

- [ ] A true 391/1591 clock-ratio simulation runs both grids long enough to
      expose the old drift and proves bounded phase error with no sample slip.
- [ ] CRF selection drives a non-zero servo command through the exact live
      status slice; GET_CLOCK_SOURCE reads back the same selection.
- [ ] CRF lock loss enters the declared holdover/restart behavior without an
      undeclared frame discontinuity.
- [ ] INTERNAL mode remains free-running and does not consume CRF trim.
- [ ] A bench probe compares the TDM frame pin with the media-grid test point
      over a duration long enough to expose 10.6 ppm.
- [ ] No open `[GAP]` remains for source selection, grid alignment, or `mr`
      trigger reachability; clock, CRF, packetizer, and root regressions pass.

[...]

Issue #66, opened 2026-08-16
AECP: GET_NAME/SET_NAME unimplemented — the image emits no name table, offset 4 is not lane-aligned, and a written name would be invisible to READ_DESCRIPTOR


Mister-M-alt:
## Summary

`GET_NAME` (`0x0011`) and `SET_NAME` (`0x0010`) are Milan v1.2 SHALL commands and
this entity answers both with the conformant `NOT_IMPLEMENTED` echo. They are the
last two commands of the *enumeration and identity* set still missing, and they
block five items of the end-station validation test plan.

This is **not** a "wire up two microprograms" job, and the issue exists mostly to
say why. Three separate things are missing, and the first of them is a sizing
decision that needs a call before any RTL is written.

Landed context: the AECP dynamic-state store (`KL_aecp_dyn_state.sv`) and the
`SET_SAMPLING_RATE` / `SET_CLOCK_SOURCE` pair went in at VERSION `0x004C`, so the
pattern a name command follows — locate, check the lock, write, echo — already
exists and is tested. Names do not fit it, for the reasons below.

## Defect 1 (blocker): there is no name table in the descriptor image

`avdecc/gen_aemi_image.py:239-241`, deliberately:

> Names are left to the descriptors' own inline `object_name` fields, so no name
> table is emitted (`name_index` unset means "unnamed" to the packer). The store's
> name table is an optimisation for models that share strings.

So every descriptor's `nbase` reads `0xFFFF` ("unnamed") and the names live inline
at descriptor offset 4. `KL_aecp_desc_store` has a perfectly good writable
name-table overlay — it is the *only* writable region it has — and nothing puts
anything in it.

## Defect 2 (blocker): offset 4 is not lane-aligned, and the µISA has no shift

The state port reads a 64-bit **lane**: `st_addr[15:3]` picks it, and `COPY_BUF`
starts at the lane boundary. A 64-byte `object_name` that begins 4 bytes into lane
0 cannot be lifted by either mechanism.

This is the same constraint that decided which fields the `0x004B` read-side set
could serve at all:

> A field is servable from the descriptor image only if it **ends its 8-byte lane**
> (readable right-justified by `READ_ST` + `BUILD_FLD`) or **begins one** (liftable
> by `COPY_BUF`). Everything else needs a shift the µISA does not have.

## Defect 3 (correctness): a written name would be invisible to `READ_DESCRIPTOR`

This is the one that makes "just write the overlay" wrong rather than merely
[...]

PR #104, opened 2026-08-18
Gate Milan feature status consistency

Mister-M-alt:
[A0] Status: Ready for review

Description:
- Defines one JSON ledger for repeated Milan feature claims.
- Adds machine-checked status rows to active authority pages.
- Anchors the gateware version and served AEM inventory to source declarations.
- Records START/STOP as partial until issue #97 is fixed.
- Adds the offline checker and negative controls to the documentation gate.

how-to-reproduce:
- Change one marked implemented row to missing.
- Run python3 scripts/check_feature_status.py.
- Observe the feature identifier and every claiming file in the failure.

how-to-validate:
- python3 scripts/check_feature_status.py --self-test
- python3 scripts/docs_check.py
- python3 scripts/check_doc_paths.py
- python3 scripts/check_archive.py
- python3 scripts/gen_toc.py --verify-anchors
- python3 scripts/gen_toc.py --check
- GIT_DIR=/nonexistent python3 scripts/check_feature_status.py

DoD:
- [x] Stable feature identifiers and canonical statuses
- [x] Current active claims aligned
- [x] Archive and obsolete pages ignored
- [x] Conflict, duplicate, unknown-status, and unknown-reference controls
- [x] No added documentation em dashes
- [x] Existing documentation gates pass

Tracks #98.


Mister-M-alt:
[R1] Review round 1: NEGATIVE — two blockers, one accuracy set, then a re-review can be positive. Three cleared-context lenses (checker correctness, claim accuracy vs the tree, gate/workflow integration) read the diff cold.

**Blocker 1 — the checker has no fence state, and the false green is demonstrated** (`scripts/check_feature_status.py` `parse_document`). Reproduced both directions on a scratch copy:
- Delete the real marked block from a ledger-listed page and leave only a copy inside a ``` fence → exit 0. The fenced copy satisfies both the is-not-marked cross-check and status agreement, so a page can lose its live stamp without the gate noticing.
- Append a fenced *illustrative* row → exit 1 (`duplicate feature identifier` / `unknown feature identifier`), so an author cannot quote an example row in prose at all.
Fix shape: track ```/~~~ fences exactly as `docs_check.check_md` does, then add a fenced-row negative control to `--self-test` (today's 10 fixtures cover neither direction). An outer HTML comment likely has the same hole (plausible, not demonstrated).

**Blocker 2 — active pages that contradict the ledger today, invisible because they carry no markers:**
- `docs/ENDSTATION_BUILDER.md:738-740` states SET_NAME "still works and is gated" — the engine has no `OP_SET_NAME` decode, issue #66 is open, and the ledger itself lists SET_NAME under `missing_mandatory_aem_operations`. Reword or stamp.
- `docs/reference/REGISTER_MAP_CLASSES.md:70` states current VERSION `0x0002_0043` — stale vs `milan_csr.sv` `0x0002_0051`, i.e. a live version claim the `gateware.current-version` feature missed.
- `docs/reference/FR_NFR.md:94` declares the `OP_*_C` table "the authoritative served set" while the new MILAN_FEATURE_STATUS.md declares the JSON canonical — two competing authorities in active docs; its whole verdict table (117-140) is accurate today but unstamped. Related demonstrated hole: a status-shaped row placed *outside* a marked block is silently ignored (exit 0), so drifted claims that never had markers stay invisible — worth stating as a known limit if not closed.

**Verified good** (so the re-review can be fast): ledger facts all true against the tree — VERSION anchor matches `milan_csr.sv:110`; the 26-op served set matches `KL_aecp_engine.sv` and `aecp_engine_steps.py` both directions; partial/missing verdicts for #97/#70/#69/#66 all honored. All six doc gates pass at head; no em dash on any of the 861 added lines; commit style clean; docs.yml wiring sound including the no-git job (verified by running with git hidden); archive/obsolete exclusion works on every contradicting legacy page sampled.

Minor, non-blocking: unreadable tracked file crashes `active_markdown_documents` instead of producing a finding; `served_mvu_operations` is shape-checked but never cross-checked against a source; the long-gate-policy block sits mid Section 2.1 while the policy text lives ~160 lines later; `verification.long-gate-policy` is repo process, not a Milan feature — slight ledger scope creep; each active md is read twice.

Mister-M-alt:
[A0] Updated PR #104 to address the review findings: fenced and unmarked status rows are handled, source parsing is comment-safe, malformed ledger data reports cleanly, active-page coverage is broader, and the stale version, notification, clock, and SET_NAME claims are corrected. Validated with 16/16 checker controls, all documentation gates, local Verilator 50/50 with 2,113,729 checks, and local Yosys 44/44.

Mister-M-alt:
[R1] Review round 2 on 03130a23: POSITIVE. Both round-1 blockers are closed, reproduced from the failing side by two fresh cleared-context agents.

Blocker 1 (fence state): deleting the real marked block and leaving a fenced copy now fails loud (9x "is not marked in README.md", rc=1); a fenced illustrative row no longer false-reds (rc=0). Every HTML-comment variant is either handled or fails noisy -- no silent path found. The self-test's 16 controls now include both fence directions and the unmarked-row case. Blocker 2 (unmarked drifted row): the exact round-1 repro row in prose now yields "feature-status row is outside a marked block", rc=1.

Claim accuracy: all five round-1 items verified true against the tree at head -- ENDSTATION_BUILDER's SET_NAME wording (no OP_SET_NAME decode, #66 open, stamped missing row), REGISTER_MAP_CLASSES + milan_csr.md at 0x0002_0051 with machine-checked rows tied to the milan_csr.sv declaration, FR_NFR reworded to defer to the ledger with seven agreeing stamped rows, the clock claim matching the exported-but-unconsumed aecp_clk_src_index_o wiring (#74), and the notification claim matching the real ev_amap producer with the open halves marked partial (#69). Newly stamped pages spot-checked truthful; the ledger document lists make the check bidirectional.

Non-blocking residuals, for the record or a follow-up: (1) an unreadable tracked file still exits via a raw FileNotFoundError traceback rather than a clean finding (fail-loud, so not a blocker); (2) served_mvu_operations remains shape-checked only, though a cross-check source exists at aecp_engine_steps.py:210; (3) FENCE_RE toggles on any fence marker regardless of character, so a lone ~~~ literal inside a backtick fence desyncs the tracker -- demonstrated false red, and only silent in a page with no required block; (4) a status row outside a block with a typo'd status word is still invisible. All four fail safe or need contrived input.

From this session's side the re-review bar is met; checker 0 findings, self-test 16/16, all five doc gates rc=0, no-git run rc=0, no em dash on any added md line.

Mister-M-alt:
[A0] PR #104 now includes exact machine-checked AEM and MVU inventories, closes all cleared-context review findings, corrects remaining active clock, map, notification, and builder-pipeline claims, and marks the superseded media-clock finding obsolete. Validated with 18/18 checker controls and every documentation gate.

Mister-M-alt:
[R0] PR #104 at d10326d2 is validated. Two independent reviews and all required local gates pass.

Issue #110, opened 2026-08-18
gPTP plane: integrate the FPGA-gPTP processor and retire ptp4l


Mister-M-alt:
## Summary

Integrate the gPTP protocol processor (https://github.com/Mister-M-alt/FPGA-gPTP) into this repository as a fabric time-sync plane, first as a build option and then as the DEFAULT, retiring `ptp4l` + `phc2sys` + the `milan-statd`/`gptp2csr.sh` mirror chain from the softcore.

The donor repo is the gPTP sibling of the `protocol-processor` submodule: a micro-coded engine in the same integration shape (byte in / byte out, one clock domain), whose uCPU is a pure superset of the AECP one (adds `OP_ALU` and `OP_MULDIV` for 64-bit timestamp arithmetic). Its current state is a verified resource-validation skeleton: parser, TX slot, timer service, engine region map and both mutation-proven Verilator suites exist; the 802.1AS state machines of record (BTCA, asCapable ladder, receipt timeouts, Milan v1.2 Table 4.1/4.2 profile numbers) still have to land as ucode revisions on the same ROM.

Why move it into fabric (all measured on this project, see `docs/design/GM_LOSS_RECOVERY.md` and `docs/ARCHITECTURE_HW_SW_SPLIT.md`):

- The single-hart service budget starves `ptp4l` under load into false "multiple GM" verdicts; a fabric plane has no scheduler.
- GM-loss recovery through `ptp4l` is a step-vs-slew compromise (a running instance slews at about 26k ppm, 40 minutes for a 60 s offset; the operational rule is "restart once"). The fabric servo can implement the DLL-like fast GM reaction directly.
- The publication chain (`ptp4l` -> UDS/pmc -> `milan-statd` -> devmem -> CSR 0x624/0x628, 0x6E4, 0x778, the 0x730 group) becomes wires; the fail-stale mode of a restarted `ptp4l` disappears with the chain.
- A JTAG warm reload kills the TX timestamp path and drives `ptp4l` FAULTY (`tu=1`); a fabric plane restarts with the bitstream.
- It completes the fully-in-fabric direction the lwSRP and AECP/ACMP planes already took.

## Work

[...]

PR #113, opened 2026-08-19
The gPTP plane re-pins onto the real counter (issue #110, round 3a)

Mister-M-alt:
[A1] The parent-side beachhead for #110. The gptp-processor submodule was already in the tree at the bench-grandmaster tip (4490637) with nothing referencing it; this PR brings the pin forward across the two donor rounds that landed since -- ucode v4 (the asCapable ladder, the Milan Table 4.2 receipt timeouts, neighborRateRatio) and v5 (the step-vs-slew PHC servo) -- and gives the parent tree its first integration proof.

**Pin e390d65** (= donor PRs #1 and #2, each merged through a cleared-context review with reviewer-authored mutations; donor battery at the pin: engine 120 checks, ucpu 768, parser 31, sixteen planted mutations red, still zero RTL change since the priced skeleton).

**New suite tb/verilator/gptp_plane** (13 checks): the submodule's KL_gptp_engine married to the PARENT's real hdl/ieee8021as/ptp_timestamp/timestamp_counter.sv -- the exact pairing the datapath splice will instantiate. The donor's own suite proves the protocol against a MODEL of the counter; this one replaces the model with the counter itself (Q8.24, 8.0 ns increment) and proves closed-loop: boot beside the ticking counter, the asCapable ladder off two pdelay exchanges, adoption of a better announce into the publish bank, then a +100 ppm master 1 ms ahead gets ONE adjtime re-base near +1 ms and the PI addend slews the REAL counter into lock -- measured offset under 150 ns, the latched adjfine level at the +100 ppm ideal (13,421 Q8.24 units within 15%), and the counter's advance tracking the master's within 100 ns over the last four sync intervals. The wrap carries the one piece of fabric the splice will add beyond wires (the adjfine pulse-to-level latch); both integration-specific mutations (a dead latch, a dead adjtime strobe) turn the run red.

**Plumbing**: the .gitmodules URL moves from SSH to HTTPS so anonymous CI can fetch it (the protocol-processor precedent; the SSH form was uninitializable in CI), and the rtl.yml sweep job fetches gptp-processor so the new suite runs remotely. The suite's ucode hex is generated per make from the submodule (untracked: its content depends on the pin, so a tracked copy would silently diverge on the next bump).

**Gates run here**: the new suite 13/13; docs_check 0/205, TOC OK, doc paths OK, lint ratchet PASS, zero added em dashes. No hdl/ file changes, so no other suite's inputs moved and no VERSION bump is due; the remote sweep covers the full battery per the recorded long-gate policy.

**Not in this PR** (the splice round that follows, opening with the area verdict against the donor's OOC anchors since the part is LUT-bound): the 0x88F7 classifier steer, the control-cascade TX + txts return, the publish bank onto the CSR words and the tu bit, the elaboration option and default, and the rootfs retirement of ptp4l.

### DIFF
--- a/.github/workflows/rtl.yml
+++ b/.github/workflows/rtl.yml
@@ -28,7 +28,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Fetch the RTL submodules
-        run: git submodule update --init third_party/verilog-axis protocol-processor
+        run: git submodule update --init third_party/verilog-axis protocol-processor gptp-processor
 
       - name: Cache the pinned Verilator build
         id: cache-verilator
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,4 +12,4 @@
 	url = https://github.com/Mister-M-alt/protocol-processor-control-plane-avb-milan.git
 [submodule "gptp-processor"]
 	path = gptp-processor
-	url = git@github.com:Mister-M-alt/FPGA-gPTP.git
+	url = https://github.com/Mister-M-alt/FPGA-gPTP.git
--- a/docs/traceability/MODULE_MATRIX.md
+++ b/docs/traceability/MODULE_MATRIX.md
@@ -130,7 +130,7 @@ _gPTP timestamping / pdelay / sync_
 | ✅ `ptp_csr_sync` | `ieee8021as/ptp_timestamp/ptp_csr_sync.sv` | `hostplane` · `milan_dp` · `ptp_sync` · `ptp_ts` | — |
 | ✅ `ptp_ts_core` | `ieee8021as/ptp_timestamp/ptp_ts_core.sv` | `hostplane` · `milan_dp` · `ptp_ts` | — |
 | ✅ `ptp_ts_top` | `ieee8021as/ptp_timestamp/ptp_ts_top.sv` | `hostplane` · `milan_dp` · `ptp_ts` | — |
-| ✅ `timestamp_counter` | `ieee8021as/ptp_timestamp/timestamp_counter.sv` | `hostplane` · `milan_dp` · `ptp` · `ptp_ts` | — |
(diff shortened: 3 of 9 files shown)


Mister-M-alt:
[R1] Cleared-context review of the PR head: sound, with findings. Verified: exact 8-file scope, the gitlink = the donor main tip containing its two reviewed rounds; the suite 13/13 with the hex regenerating from the submodule; the counter port mapping correct incl. the one-cycle adjtime pulse contract; no tautologies (the mst_base floor terms cancel exactly, the addend check anchors on the bench constant, an unsteered counter misses the rate window 8x); the timescale self-consistent with the generator's gain formula; anonymous HTTPS clone works and every CI job's submodule needs are coherent; four reviewer-authored mutations red (wrong increment, negated adj, drifted master model, wrong-clock ucode). Findings: (F1, blocking) the traceability no-drift gate is red -- the new suite adds gptp_plane to timestamp_counter's matrix row and the PR did not regenerate MODULE_MATRIX.md + README-tests.md; (F2) tying phc_ns_i to a constant survives -- it is dispatch metadata in the current ucode, unobserved by the checks; (F3, note) the closed-loop checks are steady-state-anchored, so a halved interval survives while gross gain errors are caught; (F4) the wrap's adjfine latch was async-reset beside a sync-reset counter.

[A1] Answered in the head commit: F1 -- make matrix regenerated exactly the two files (the no-drift contract now passes 2/2 locally; the docs job goes green); F4 -- the latch is sync-reset, matching the counter it feeds. F2 is a REAL blind spot and is now stated as such in the suite README: no current ucode handler consumes phc_ns_i, so its first functional consumer (a donor-side ucode revision) and the observing check land with the splice round -- tracked on the #110 later-rounds list. F3 stands as noted: steady-state anchoring is the design of these checks; the transient shape is the donor engine suite's job. One hygiene remark for the record: the two em-dash lines on the delta are the matrix GENERATOR's own empty-notes cell format on regenerated rows -- machine output under the no-drift contract, not authored prose; a dash-policy change there belongs to the generator, not a hand edit. Suite re-run 13/13, matrix contract 2/2, docs_check 0/206, doc paths OK, zero authored em dashes.

Issue #115, opened 2026-08-19
gPTP plane: donor ucode completion (BTCA full compare, cease rule, source matching)


Mister-M-alt:
Subticket of #110. The 802.1AS remainder in the gptp-processor submodule, each a ROM-words round on the same engine (zero RTL change so far holds):

## Work
- [ ] Full BTCA: stepsRemoved and sourcePortIdentity tie-breaks (the two-node subset deliberately excluded them on a point-to-point bench).
- [ ] The multiple-responder cease rule of Milan v1.2 4.2.6.2.5 (stop Pdelay_Req on 3 successive multi-CID responses; resume on link toggle / port toggle / 5 min).
- [ ] Sync sourcePortIdentity + sequence matching between Sync and its Follow_Up.
- [ ] An asCapable fall while slave stops sync consumption (today the servo would keep steering on the stale last-measured pdelay -- PR #2's review note).
- [ ] A ucode consumer for phc_ns_i (the live-clock snapshot), which unlocks the parent-side observing check (PR #113 review F2).

## Proof
The donor engine suite grows per revision with mutation proof (house rule; battery at the current tip: engine 120 / ucpu 768 / parser 31, sixteen mutations red).

## Dependencies
#110 (parent). Feeds the splice round's observing check.

Mister-M-alt:
[A1] Round open, first PR up: Mister-M-alt/FPGA-gPTP#3 (ucode v6, the full-compare round). Landed in it: full BTCA against a stored best-vector record (parent updates including degradation with IMMEDIATE takeover, stepsRemoved and sourcePortIdentity tie-breaks, best reset on become, and the v5 sync-ok 1 Hz flicker retired -- cleared only on a GM change now), Sync/Follow_Up sequence+source pairing, the asCapable consumption gate, and the Sync originTimestamp carrying the live PHC via gather -- the phc_ns_i consumer that unlocks #114's observing check. Engine suite 120 to 141 checks, six new mutations red (twenty-two lifetime), zero RTL change (three write-only scratch slots reclaimed instead). Two real traps found while building, both fixed and documented: e_ult64's scratch registers must never be its operands, and the single-message-bank overwrite race (bank reads now hoisted to handler entry; the second bank is an engine revision listed for the scratch-widen PR). Remaining for this ticket: the Milan 4.2.6.2.5 cease rule with the scratch widen. Review running.

Mister-M-alt:
[...]

Issue #114, opened 2026-08-19
gPTP plane round 3b: the datapath splice, gated by the area verdict


Mister-M-alt:
Subticket of #110. The plane is in the tree at its proven tip (PR #113) but no shipping build instantiates it.

## Work
- [ ] Area verdict FIRST: OOC the plane at the shipping shape against the donor's docs/RESOURCE_VALIDATION.md anchors and the current utilization; the splice does not start until the budget is settled (the part is LUT-bound; the plane priced at ~2.9k LUT; the standing directive defaults memory-shaped state to the DRAM control-path base).
- [ ] Classifier steer of EtherType 0x88F7 with the sof-latched ingress timestamp into the plane (KL_pp_shadow pattern), TX onto the control cascade with the egress timestamp returned over txts_*.
- [ ] PHC wires: the adjfine pulse-to-level latch (proven in tb/verilator/gptp_plane's wrap) and the adjtime strobe onto timestamp_counter's knobs.
- [ ] Publish bank onto the CSR words milan-statd mirrors today (0x624/0x628, 0x6E4, 0x778, the 0x730 group) and plane ownership of the tu bit (no CLKV lease).
- [ ] Elaboration-time build option (static shape rule: a build parameter, not a runtime poke), default OFF in this round.
- [ ] The tb/verilator/gptp_plane suite grows the phc_ns_i-observing check (the PR #113 review's F2 blind spot) once a ucode consumer exists, and milan_dp gains a leg or checks covering the spliced faces.

## Acceptance
- [ ] A bitstream builds with the option ON inside the LUT budget and timing closes; all suites, yosys, behave and builder gates green in BOTH option states.

## Dependencies
#110 (parent). The donor ucode leftovers ticket feeds the phc_ns_i consumer.

Mister-M-alt:
[A1] Round open as draft PR #119 (the contribution flow's vehicle; it flips to ready when the round completes). On it so far: the pin advance to the donor v7 tip, the fabric wrapper KL_gptp_shadow (with the commit-pulse ingress-timestamp transport) and the observer-pure boundary stamper KL_gptp_txstamp, lint-clean and in the traceability matrix. In flight: the fresh full-Vivado area baseline of dev (the ticket's gate -- the plane's 3,046 LUT lands near the historical closure cliff, so the verdict is measured, not arithmetic), the wrapper's Verilator suite, and the option-gated datapath splice, default OFF.

Mister-M-alt:
[...]

PR #118, opened 2026-08-19
The gPTP pin advances to v6 and the blind spot closes (issue #115)

Mister-M-alt:
[A1] The parent-side companion of the #115 donor rounds: the gptp-processor pin advances to 71ee8de -- the v6 full-compare tip merged through review (donor PR #3: full BTCA against a stored best-vector record with parent updates and immediate takeover on degradation, Sync/Follow_Up sequence-and-source pairing, the asCapable consumption gate, the dispatch-identity guard against the single-bank race, and the Sync originTimestamp carrying the live PHC).

**The payoff this bump unlocks**: PR #113's review filed phc_ns_i as an F2 blind spot -- tying the engine's live-clock input to a constant survived every check because no ucode consumed it. v6's gather consumer makes it observable, and tb/verilator/gptp_plane grows phase 5: announce silence rides the receipt timeout to grandmaster (pdelay keeps asCapable alive), and the transmitted Sync's originTimestamp must be the live timestamp_counter value. The formerly-surviving mutation now fails the run (verified both ways in this branch); the suite goes 13 to 15 checks.

**Deliberately NOT in this PR**: the donor's in-review engine v2 (the second message bank + the cease rule, donor PR #4) -- that pin advance ships with round 3b (#114), whose area verdict prices against the re-measured 3,046 LUT engine.

Gates here: the suite 15/15 with the mutation red-then-restored, docs_check 0/206, matrix no-drift 2/2, doc paths OK, zero added em dashes. No hdl/ changes; the remote sweep covers the battery per the long-gate policy.

### DIFF
--- a/tb/verilator/gptp_plane/README.md
+++ b/tb/verilator/gptp_plane/README.md
@@ -20,13 +20,12 @@ What it proves:
 | 2 | a live auto-answering peer raises asCapable at the second exchange |
 | 3 | a better announce is adopted; the publish bank carries the GM identity and role flags |
 | 4 | closed loop vs a +100 ppm master 1 ms ahead in counter time: ONE adjtime re-base near +1 ms (the correction negates the offset), the latched adjfine level lands at the +100 ppm ideal (13,421 Q8.24 units, within 15%), the measured offset locks under 150 ns, and the REAL counter's advance tracks the master's within 100 ns over the last four sync intervals |
+| 5 | announce silence rides to grandmaster and the transmitted Sync's originTimestamp is the live counter value -- phc_ns_i observed on the wire (a constant-tied mis-wire fails) |
 
-Known blind spot, on purpose: `phc_ns_i` (the engine's live-clock
-snapshot input) is dispatch metadata in the current µcode -- no handler
-consumes it yet -- so a mis-wire of that one input is invisible to these
-checks. The loop closes through the ingress/egress timestamps. Its first
-functional consumer is a donor-side µcode revision; the splice round
-carries the observing check.
+The `phc_ns_i` blind spot from PR #113's review is CLOSED: the
+submodule's v6 ucode consumes the live-clock input for the Sync
+originTimestamp, and phase 5 observes it on the wire -- tying
+`phc_ns_i` to a constant now fails the run.
 
 Timescale note: the bench clock is 2 MHz while the counter keeps its
 8.0 ns/tick shape, so counter time runs 62.5x slower than the bench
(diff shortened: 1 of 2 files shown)


Mister-M-alt:
[R1] Cleared-context review: VALIDATED. Verified: exact three-file scope; the gitlink is precisely the donor main tip (the v7 second-bank work confirmed NOT in this pin -- single 32-word bank at the pinned engine); suite 15/15; the reviewer re-ran the constant-tie mutation independently (red on the origin check, every other check surviving -- consistent with the pre-v6 total blind spot) and restored to green; phase 5 judged non-tautological (wire-parsed origin, per-build gather in the donor ucode, timeout counted as a failed check, 2.5x timer margin, servo state cancels); donor engine suite 149/149 + lint at the pin; hygiene clean. One low nit: the origin parse indexed bytes 48..57 with only a >14 size guarantee -- undefined behavior on an already-broken DUT emitting a truncated Sync.

[A1] The nit is taken in the head commit: a length check precedes the parse, so a truncated Sync fails as a counted check instead of undefined behavior. Suite 15 -> 16 checks, re-run green.

Mister-M-alt:
[R0] Reviewed and validated final merged SHA 9b8b84a4a53a6e92859257e00c59c5011c4c1ae7. The final length guard makes truncated Sync frames fail deterministically before timestamp parsing. The gptp_plane suite passes 16/16, and tying phc_ns_i to zero fails only the live originTimestamp check at 15/16. The pinned donor engine, uCPU, and parser suites pass 149/149, 768/768, and 31/31. Documentation, feature-status, traceability, cited-path, archive, and diff hygiene gates pass.

Mister-M-alt:
[R0] Reviewed final head 9b8b84a4a53a6e92859257e00c59c5011c4c1ae7. I verified the parent diff, issue #115 and donor PR #3 provenance (gitlink 71ee8de), gptp_plane 16/16, donor uCPU 768/768, parser 31/31, engine 149/149 plus lint, and the docs/no-drift/path/TOC gates. Negative controls were red: phc_ns_i=0 failed the wire-observed origin check, and a 57-byte Sync failed the new length guard without parsing. I validate the PR.

Mister-M-alt:
[R0] PASS at exact final head `9b8b84a4a53a6e92859257e00c59c5011c4c1ae7`. Validated the #115 v6 slice: the gitlink is donor merge `71ee8deb6510b6f32df478a6df50cac213a258bd`; BTCA ordering, parent degradation/immediate takeover, best-record reset, Sync/Follow_Up sequence plus single-port source pairing, the asCapable consumption gate, dispatch guard, and live-PHC Sync origin all hold. Donor engine 149/149 and parent gptp_plane 16/16 pass. A constant `phc_ns_i` mutation fails the origin check (15/16), and a forced 57-byte Sync fails the new length check without parsing. docs_check 0/206, matrix, cited paths, and diff hygiene pass. The cease-rule/v7 work remains intentionally outside this pin. No blockers.

PR #11, opened 2026-05-24
04 adp discovery controller

cemaldogann:
ADP Discovery Controller implementation - 5.6.4 Milan Specs 

#4

PR #119, opened 2026-08-19
The gPTP plane splices into the fabric (issue #114, round 3b)

Mister-M-alt:
[A1] Round 3b of epic #110, opened as a draft per the contribution flow -- the round is IN PROGRESS and lands commit by commit on this PR. Scope: the datapath splice that makes the plane real fabric, gated by the ticket's area verdict.

**Landed so far on this branch:**
- The submodule pin advances to 441a833 (the donor's engine v2 + ucode v7 tip: the second message bank, 64-word scratch, and the Milan 4.2.6.2.5 cease rule, merged through review with thirty-one mutations red).
- hdl/ieee8021as/gptp_plane/KL_gptp_shadow.sv -- the fabric wrapper in the KL_pp_shadow integration idiom: an input-only 0x88F7 tap (tvalid && tready qualified, the gh #65 rule) into a frame FIFO with counted drops, a byte serializer feeding the engine at 1 B/clk, the adjfine pulse-to-level latch, and a TX gearbox onto one frame-held wide control-lane leg. The ingress-timestamp transport is the part a register latch gets wrong: the arrival stamp rides a side FIFO pushed on the frame FIFO's status_good_frame commit pulse -- exactly once per DELIVERED frame, so the drop machinery can never desync timestamps from frames.
- hdl/ieee8021as/gptp_plane/KL_gptp_txstamp.sv -- an observer-pure egress stamper at the MAC boundary (the control lane does not traverse ptp_ts_top's TX stamper; only the shaped data path does). It latches the PHC at the first beat of armed 0x88F7 frames and returns {ts, sequenceId} for the engine's seq-matched pending exchange; armed counting pairs it to exactly the plane's frames. Tap purity holds. The one-gPTP-stack-per-port exclusion for A/B bring-up is documented in the banner.
- Both modules lint clean under the repo ratchet (99 <= 99, no new violations) and appear in the regenerated traceability matrix (currently the two honest untested rows -- their suites are the next commits).

**In flight:**
- The AREA VERDICT: a full Vivado baseline of the current dev tip is building now (the last recorded placement was 50,040 LUT / 78.9% on 08-13, three feature rounds ago; the plane measures 3,046 LUT at the donor's OOC instrument, which lands the option-ON build near the historical 84.77% closure cliff -- hence a fresh measurement, not arithmetic). The verdict lands in this PR before the option is offered ON anywhere.
[...]

PR #121, opened 2026-08-19
Complete AECP notification triggers and monitoring

### LINKED ISSUE #69: AECP: the unsolicited-notification trigger set is empty, and the departing-controller monitor (Milan 5.4.5.3) has no originator — registry lands, nothing fires

Mister-M-alt:
[A0] Completes the AECP unsolicited-notification trigger set and departing-controller monitoring required by issue #69.

This change integrates the reviewed protocol-processor implementation from [protocol-processor PR #11](https://github.com/Mister-M-alt/protocol-processor-control-plane-avb-milan/pull/11), updates the pinned submodule revision, extends the root integration and compliant tests, and records the resulting verification evidence.

Local validation:

- All 51 Verilator suites passed with 2,117,848 checks and zero failures.
- Yosys portability passed all 44 tops with zero failures.
- Documentation, traceability, source-list, archive, link, entity-shape, binding, wire-accountability, and builder gates passed.
- `milan_dp`, `csr`, and `pp_shadow` focused integration suites passed.
- No added documentation em dashes or internal compliant-test identifiers were found.

Vivado and bitstream generation were not run per the current project instruction. The optional remote Verilator and Yosys jobs have both been run locally as required.

Closes #69


Mister-M-alt:
[A0] Validation evidence for base 00b586a2802b640c66b9f89d556d95758fcfc5e7 and head 06df037c1fbb60f56e984ab1205c193ce7909df7: all 51 local Verilator suites passed with 2,117,848 checks and zero failures; local Yosys portability passed 44 of 44 tops; focused milan_dp, csr, and pp_shadow integration suites passed; documentation, traceability, source-list, archive, link, entity-shape, binding, wire-accountability, and isolated builder gates passed. The pinned protocol-processor head was independently validated before merge. No Vivado or bitstream generation was run.

Mister-M-alt:
[A0] Candidate merge validation addendum: protocol-processor passed 14,645 checks with zero failures, including its microcode map gate; BDD conformance passed 15 features, 334 scenarios, and 1,571 steps with no failures or skips; root RTL lint passed at the 99-warning ratchet. The reviewed head remains 06df037c1fbb60f56e984ab1205c193ce7909df7.

Mister-M-alt:
[R1] Independent cleared-context review at head 06df037c: POSITIVE, VALIDATED with findings all low/informational -- nothing blocking. Verified: the gitlink is exactly the PP main tip carrying PR #11's three commits and nothing else; the parent wiring is a lossless round-robin descriptor-event arbiter feeding the PP's pass-through ctr_change ports (single clock, no CDC, no new AXIS taps); unsolicited TX rides the packed PP lane through ctl_tx_mux with the gasket applying; NO new CSRs and the 0x0055 VERSION story is consistent across milan_csr, the REGISTER_MAP changelog, the ledger and all four full-value sim pins. Runs reproduced: the full sweep 51/51 at exactly the claimed 2,117,848 checks (tsn-gen tier accounted), pp_shadow 273, csr, all milan_dp legs, behave 334/1,571 (the README tally was honestly re-run after going stale on dev), and every docs/lint/matrix/wire gate green with zero added em dashes. The Milan numbers check out against the v1.2 PDF: 5.4.5.2 requester exclusion and no-op silence, the Table 5.22 one-per-second-per-descriptor limiter, 5.4.5.3's randomized 30-60 s with one 250 ms retry, any-status rearm and targeted DEREGISTER. Two reviewer-authored mutations red at exactly the pinned checks (dropped input-dirty accumulation; lossy clear-all on output pick).

Worth applause: all four residuals from my PP-side [R1] rounds were FIXED at the merged head rather than waived -- the parked per-entry cancels retire the false-deregistration collision, the abortable A_START handles the grant+1 canceled frame, KL_pp_release_merge closes the release race, and the lock-expiry exclusion is now a deliberate, documented all-controllers broadcast -- each with dedicated PP tests in the audit.

Findings, none blocking: (1) LOW -- milan_datapath.sv:1838's clkv_tu_w declaration lost its indentation (style only); (2) INFO -- a dirty pulse landing the same cycle its descriptor is picked coalesces into that pick (benign for current-state notification bodies; a comment would help); (3) INFO -- syn/ooc/pp_shadow_ooc.tcl gained KL_pp_release_merge but not KL_aecp_ca_originator (consistent with that instrument's pre-existing file set; unverifiable while Vivado is paused); (4) INFO -- behave does not grow for #69 (parent coverage lives in milan_dp + the PP suites), stated here so the coverage story is explicit. Merge-ready from this reviewer's side; note the open PR #119 also touches milan_datapath, so whichever lands second rebases.

Mister-M-alt:
For now Task is on hold

Mister-M-alt (commented):
[R0] FINAL REVIEW — **NEGATIVE / DO NOT MERGE THIS CANDIDATE** — exact head `06df037c1fbb60f56e984ab1205c193ce7909df7` against current `dev` `c839aa9db5125dd3127b73581fc0dc8e614b5321`.

**Conformance — implementation PASS at the reviewed head; delivery BLOCKED by project state.** The prior cleared-context R1 traced the trigger/exclusion/no-op rules, per-descriptor one-second limiter, 30–60 s controller monitor, single 250 ms retry, any-status rearm, and targeted deregistration against #69 and reproduced the parent/donor checks. No contrary exact-head defect was subsequently reported. However, #69 is explicitly On Hold and must remain open; this PR body still says `Closes #69`, which is incompatible with the current preservation rule and the public hold.

**RTL / architecture — PASS for the exact-head mechanism.** The reviewed parent integration uses a pending-bit, round-robin descriptor-event arbiter, keeps simultaneous events until selected, drives the processor counter-change tuple without a new CDC, and routes unsolicited transmit through the existing control lane. The pinned donor merge `787404009068bb47f9686170c752c9f08b2baf12` is durable: donor PR #11 is merged and the pin is four commits ahead of the parent-compatible `a25b5cc9` base. The earlier R1 mutation checks caught both dropped dirty accumulation and a lossy clear-all picker.

**Robustness / integration — BLOCKED.** This branch is one PR commit ahead of its 2026-08-19 base but current `dev` is now 60 commits ahead. A deterministic merge-tree attempt conflicts in six files: `docs/reference/MILAN_FEATURE_STATUS.md`, `hdl/ieee1722/avtp/doc/TEST_RESULTS.md`, `syn/ooc/pp_shadow_ooc.tcl`, `syn/yosys/ooc.sh`, `syn/yosys/run.sh`, and `tb/verilator/milan_dp/Makefile`. Several are exactly the source-inventory/build-contract surfaces redesigned by #170, so resolving them in this stale branch risks undoing landed derivation work. GitHub correctly reports `CONFLICTING / DIRTY`.

**Tests / evidence — historical PASS, current candidate absent.** The exact head has recorded local evidence for 51/51 Verilator suites (2,117,848 checks), 44/44 Yosys tops, 14,645 processor checks, 334 BDD scenarios/1,571 steps, lint, documentation, and focused integration. Remote documentation/BDD checks are green, but the remote Verilator and Yosys jobs are `CANCELLED`. More importantly, no conflict-resolved current-`dev` candidate exists, so none of those results validates what would be merged now.

**Documentation / traceability — BLOCKED for re-land currency.** Exact-head documentation gates were green on the old base and the mechanism was described consistently then. Sixty subsequent `dev` commits changed the current ledgers, review contract, source derivation, and related #169/#170/#178 dependency record. A new PR must update the current VERSION/feature/audit surfaces and say `Relates to #69`, not `Closes #69`.

**Disposition:** close PR #121 without merge and preserve its branch/evidence. Keep original #69 open and On Hold. Re-land the reviewed mechanism selectively from current `dev`, preserve #170's derived source consumers, reconcile #169/#178, rerun the complete candidate gates, and obtain fresh exact-head reviews before any merge.

Mister-M-alt:
[R0] Closed without merge after the final exact-head negative disposition: https://github.com/kebag-logic/milan-fpga/pull/121#pullrequestreview-4991496355

Replacement implementation ticket: #183. Original compliance task #69 remains open and On Hold. The PR branch and its validated implementation evidence are preserved; no branch deletion was requested.

Mister-M-alt:
[R1] **NEGATIVE** — round 1, at head `06df037c`, base `dev` `c839aa9d`.

Reviewed from #69, the linked Milan clauses and the diff. **Independence declaration**: I have not seen this lane's reasoning or its submodule PR discussion, but I am not a cleared session in general — I have been working in this repository today on #154/#185, which touch `sw/builder` and CI and do not overlap this diff. A second reviewer outside that is still owed.

## Lens coverage this round

| lens | result |
|---|---|
| `Conformance` | 2 BLOCKER |
| `RTL` | 1 MAJOR, otherwise clean |
| `Robustness` | PASS |
| `Tests` | 1 MAJOR |
| `Docs` | 1 MAJOR |

---

```text
[R1] BLOCKER Conformance — protocol-processor @ 787404009068bb47f9686170c752c9f08b2baf12 — the pin predates four commits merged to the submodule's own main, two of which change RTL and ucode
Requirement/evidence: `git branch -r --contains` puts the pinned commit on `origin/main`, and `git log 7874040..origin/main` lists four further commits from `121-notification-review-fixes`, all merged: `Fix notification review findings`, `Clarify mapping notification semantics`, `Fix notification review regressions`, `Clarify integrator ownership`. Two of them touch `hdl/aecp/KL_aecp_engine.sv`, `hdl/aecp/ucode/gen_ucode.py` and `hdl/top/protocol_processor_top.sv`, and add 153 lines of `tb/pp_top` coverage.
Impact: merging this superproject PR lands the notification feature WITHOUT the fixes that its own review produced. This is the #87/#88 stranding shape with a submodule boundary in the middle, where nothing in the superproject can see it.
Required change: pin the reviewed head that includes those four commits, and re-run the evidence at that pin.
Verification: `git -C protocol-processor merge-base --is-ancestor <new pin> origin/main` and `git log <new pin>..origin/main` empty for the notification work.
```

```text
[R1] BLOCKER Conformance — protocol-processor@7874040 hdl/top/protocol_processor_top.sv:3243 — GET_AS_PATH notifies only on grandmaster change, not on path-sequence change
Requirement/evidence: at the pinned commit the connection is `.ev_asp_i (gm_change_i)`, and its own comment says "the path follows the grandmaster here". #69 quotes Milan Table 5.22: `GET_AS_PATH` fires when **the path sequence changes**. The fix commit adds `gsi_asp_chg_i` and rewrites this as `gm_change_i || gsi_asp_chg_i`, with the comment "Grandmaster changes are derived internally; this pin covers tail changes that retain the same grandmaster".
Impact: a PathTrace tail change under an unchanged grandmaster produces no notification. A registered controller's cached AS path is silently stale, which is the exact condition Table 5.22 lists this trigger for.
Required change: carry the fix (it exists upstream) rather than re-deriving it.
Verification: the `tb/pp_top` arm added by `Fix notification review findings`, run at the new pin.
```

```text
[R1] BLOCKER Conformance — protocol-processor@7874040 hdl/top/protocol_processor_top.sv:3259 — the availability monitor is refreshed by only three of the six defined AECP command types, so a live controller is deregistered
Requirement/evidence: at the pin, `rx_cmd_valid_i` is `(AEM && type==0) || (MVU && type==6) || (AA && type==2)`. #69 quotes Milan 5.4.5.3: "After each valid AECP command received from a registered controller ... the PAAD-AE shall (re)set the monitor timer associated with this controller." The fix enumerates the six defined command message types `{0,2,4,6,8,14}` and states in its own comment that only the reserved 10/12 must not refresh.
Impact: a registered controller that keeps talking using an AVC, HDCP_APM or EXTENDED command lets its monitor timer run out, gets probed with CONTROLLER_AVAILABLE and, on the 27-66 s window #69 cites, is removed from the registry while demonstrably alive. The failure is worse than no monitor, because it drops working controllers.
Required change: as above, carry the fix.
Verification: a `tb/pp_top` arm sending each of the six types from a registered controller and requiring the monitor to rearm; the fix commit adds coverage of this shape.
```

```text
[R1] MAJOR RTL — hdl/milan/KL_pp_shadow.sv:291 and hdl/milan/milan_datapath.sv:6488 — the integration is written against the pre-fix port list, so bumping the pin is not a one-line change
Requirement/evidence: this PR adds `ctr_change_i` / `ctr_change_desc_type_i` / `ctr_change_desc_index_i` through the shadow and drives them from a new round-robin arbiter. The fix commits add ANOTHER top-level input, `gsi_asp_chg_i`, which nothing in this diff drives because it does not exist at this pin.
Impact: taking the required pin bump above leaves a new input undriven. In this repository an undriven processor input is the tie-off class `scripts/check_tied_inputs.sh` exists for, and the symptom is the silent one: the GET_AS_PATH trigger stays dark and every gate stays green.
Required change: wire `gsi_asp_chg_i` from whatever publishes a PathTrace tail change in the gPTP plane, or tie it off with a stated reason recorded where the tie-off gate reads it.
Verification: `scripts/check_tied_inputs.sh` and the wire-accountability gate at the new pin.
```

```text
[R1] MAJOR Tests — the PR body — the recorded evidence was not measured on the merge candidate
Requirement/evidence: the body reports 51 suites / 2,117,848 checks / 44 Yosys tops. `gh pr view` reports `CONFLICTING`, and `git log --oneline origin/dev ^06df037c` counts **50** commits on `dev` that this head does not contain. CONTRIBUTING section 2.1 step 7 requires the candidate merge result to be gated, not the branch: "A merge is a change nobody wrote and nobody reviewed."
Impact: the numbers are true of a tree that cannot be merged. The 2026-08-16 duplicate-pin incident that step 7 exists for was invisible on both branches and only appeared in the merge.
Required change: rebase or merge `dev`, then re-run the full local bar and record both object IDs with the results.
Verification: the check counts from the candidate tree, compared against the current figures rather than only pass/fail.
```

```text
[R1] MAJOR Docs — docs/reference/milan_feature_status.json — the machine-checked ledger is flipped to `implemented` for behaviour the pinned RTL does not have
Requirement/evidence: `notifications.change-events` goes `partial` -> `implemented` with "observed asynchronous sources notify registered controllers", and `state.controller-monitor` goes `missing` -> `implemented` with "Each registered controller has a randomized availability monitor". Both statements are true at submodule `main` and false at the commit this PR pins, per the two Conformance findings above.
Impact: this ledger is consumed by `scripts/check_feature_status.py` and rendered into CONTRIBUTING and TESTING. A status row asserting a completed Milan duty is exactly the artifact a later compliance round trusts instead of re-measuring.
Required change: the rows become true when the pin moves. No separate edit needed - but they must not land ahead of the pin.
Verification: re-read both rows against the new pin.
```

---

## Clean lenses, with the artifact each was applied to

```text
[R1] PASS Robustness — hdl/milan/milan_datapath.sv `pp_ctr_event_pick` / `pp_ctr_event_next` / `pp_ctr_event_queue` — event loss, reset and index bounds on the new descriptor arbiter
```
Three things were checked and hold. **No event is lost without an acknowledgement**, which the arbiter has none of: each pending bit is cleared on the cycle it is presented, and that is safe only because the consumer latches it, which it does at `KL_aecp_notify.sv:788` (`if (ev_ctr_i && ctr_ev_ok_w) ctr_dirty_r[ctr_ev_ix_w] <= 1'b1`) into per-descriptor sticky state. **Indices stay in range**: the parent emits stream-output indices up to `ACMP_SRC_C-1` and `milan_datapath.sv:6432` passes `N_STREAM_OUT_P (ACMP_SRC_C)`, so the CRF media-clock output at index `N_STREAMS` is inside the notifier's `< N_STREAM_OUT_P` guard rather than silently dropped; AVB_INTERFACE and CLOCK_DOMAIN are emitted at index 0, which is what the notifier requires. **Reset** clears all four pending banks and the round-robin pointer.

```text
[R1] PASS Tests — tb/verilator/milan_dp/sim_nxn.cpp:472-489, 1846, 1972-1985, 2157-2179, 2190-2210 — whether the new arbiter arms can fail
```
All four sources are graded, not merely collected: stream inputs at 1846, and stream outputs, AVB_INTERFACE and CLOCK_DOMAIN each **reset to zero and then required to be seen** at 1972/2157/2190. That is the shape that can fail for the defect it claims to detect, and it is the shape I went looking for a hole in: three watchers accumulated and never asserted would have been invisible. They are asserted.

This lens is NOT covered clean, because of the MAJOR above. The arms are sound; the run they were measured in is not the merge candidate.

---

## Not blocking

The submodule pin is reachable and on the submodule's `main`, so the dangling-pin trap CONTRIBUTING names is avoided. The current `dev` pin `a25b5cc` is an ancestor of `7874040`, so nothing on the submodule is regressed by the bump itself.

## What round 2 needs

The three blockers collapse into one action: move the pin to the reviewed submodule head, wire `gsi_asp_chg_i`, rebase onto `dev`, and re-run the bar on the candidate. Re-review is owed on the corrected commit; this round is not approval of those fixes.

PR #123, opened 2026-08-19
fuzz_ptp: field-fuzz the gPTP fabric plane

Mister-M-alt:
[A1] Grows the missing `fuzz_ptp` tier of `tsn_fuzz` against the gPTP fabric plane, closing the 802.1AS field-fuzzing gap that issue #117 lists as work and that the 2026-08-13 campaign deletion opened. Subticket of #110.

The new campaign co-simulates the real fabric slice (`KL_gptp_shadow` + `timestamp_counter` + `KL_gptp_txstamp`, driving the `gptp-processor` engine) and grades it both directions: every spec-constrained field of the plane's own Pdelay_Req/Announce/Sync/Follow_Up against the tsn-gen `8021as_*` models, per-field illegal probes at the parser (drop vs 11.4.1-ignore), BTCA rejection under fuzz, servo pairing, the Milan 4.2.6.2.5 cease rule, and a two-sided asCapable canary. The 802.1AS models carry the full header, so they double as an independent decoder (cross-checked every run).

It surfaced nine real defects in the donor engine, all filed and carried as tracked gaps (a gap fires only on the mismatch, so each turns green when the donor closes it):
- FPGA-gPTP #6 — no domainNumber check (foreign-domain Announce adopted, foreign-domain Sync consumed)
- FPGA-gPTP #7 — qualifyAnnounce (10.3.10.2.1) missing: own-source / stepsRemoved>=255 / path-trace-loop vectors adopted
- FPGA-gPTP #8 — a stale Pdelay_Resp_Follow_Up poisons peer delay and drops asCapable
- FPGA-gPTP #9 — TX control field 0x5 on Sync/Follow_Up (Table 11-7 wants 0x0/0x2)
- FPGA-gPTP #10 — TX Sync writes the live PHC into the reserved body (Table 11-8 reserves it zero)

No RTL changed; new/updated files are under `tb/verilator/tsn_fuzz/` plus regenerated traceability.

Closes part of #117 (the `fuzz_ptp` work item). The two-board silicon rounds remain the rest of #117.

PR #124, opened 2026-08-19
fuzz_ptp: land the stranded review-answer commit (vacuous check + tap-drop control)

Mister-M-alt:
[A1] Follow-up to PR #123 (issue #117): lands the one commit the post-merge containment check (CONTRIBUTING 2.1 step 7) found stranded on the merged branch.

The commit answers the lane's own test-adequacy review (R1), which flagged a check that could never fail: the drought-recovery assertion masked `FL_PRESENT | FL_ASCAP`, truthy on `FL_PRESENT` alone. This replaces it with a specific `asCapable` equality and adds a positive control for the tap-drop counter (an oversize gPTP frame must be counted), which every other probe only ever asserted UNMOVED.

Test-only; the `ptp` campaign is 355 pass / 0 fail / 9 gaps on this tree. It was validated as part of #123's review answers but never pushed before #123 merged — hence this follow-up rather than an amend.


Mister-M-alt:
[A1] Validated. Head `56ab9dfc` == origin branch tip (verified before merge, per the #123 stranding lesson). Cleared-context review POSITIVE: both new checks proven falsifiable by mutation (asCapable forced to 0 -> FAIL; oversize frame shrunk -> tap-drop FAIL). Campaign 355 pass / 0 fail / 9 gaps; delta is 2 files, no RTL. Base origin/dev unchanged, so head tree is the candidate merge tree.

PR #125, opened 2026-08-19
Ship the AX7101 bare-metal gPTP profile

Mister-M-alt:
[A0]

## Status

Ready for cleared-context review. Tracks #120.

## Description

- Replace the shipping AX7101 Linux-capable CPU profile with a one-hart cacheless RV32I bare-metal VexiiRiscv profile: no MMU, supervisor mode, FPU, L1 or L2.
- Boot a minimal ROM firmware that initializes the fabric control plane, verifies the raw AEM QSPI image, exposes explicit TAI/UTC and diagnostic UART commands, and enables ADP only after identity validation.
- Make the host Linux sound-card surface optional and absent from the shipping profile while retaining the AAF, TDM, crossbar, loopback and physical-audio fabric.
- Enable the fabric gPTP option in the shipping configuration with configuration-derived microcode, and run the cacheless CPU and 64-bit Milan plane at 50 MHz across Vexii's supported CDC into the unchanged 100 MHz system/audio recipe.
- Record the three-directive xc7a100t placement sweep, the timing-clean shipping directive and the measured area buy-back.

## how-to-reproduce

```sh
sw/litex/build.sh ax7101
sw/litex/sweep.sh ax7101 i120gptp50
```

Use Vivado 2026.1 for `xc7a100t-fgg484-2`; the full cell recipe and report fields are in `docs/integration/BAREMETAL_FIRMWARE.md`.

## how-to-validate

```sh
suite_logs=$(mktemp -d)
scripts/run_all_suites.sh "$suite_logs"
protocol-processor/scripts/run_suites.sh
make -C gptp-processor
(cd tests && behave)
python3 scripts/lint_rtl.py --check
syn/yosys/run.sh
python3 sw/builder/test_builder.py
python3 scripts/docs_check.py
python3 scripts/check_feature_status.py --self-test
python3 docs/traceability/gen_module_matrix.py --check
python3 scripts/check_doc_paths.py
```

## DoD

- [x] Shipping AX7101 profile is bare-metal RV32I with no MMU, Linux or caches.
- [x] Fabric gPTP and the retained audio datapath are present; the Linux-only sound-card surface is absent.
- [x] Option-on `xc7a100t` placement and routing meet setup and hold timing, with measured resources documented.
- [x] The branch contains current `dev` and the complete local verification bar is green.
- [ ] Cleared-context review bar is met.
- [ ] Maintainer merge, post-merge containment, manual issue close and project Done transition are complete.

Issue #120, opened 2026-08-19
The baremetal downgrade: smallest RISC-V, no MMU, no Linux


Mister-M-alt:
Subticket of #110 (USER directive, 2026-08-19): when the fabric gPTP plane replaces ptp4l, the RISC-V processor is replaced with the smallest version possible -- no MMU, not Linux, just baremetal.

## Why now
The #114 area verdict measured the current dev tip at 59,497 LUT post-synth (93.84 percent) with default-directive placement FAILING outright, and the gPTP plane adds a measured 3,046 LUT: option-ON does not fit beside a Linux-capable SoC. With every control plane in fabric (protocol-processor) and time sync in fabric (gptp-processor), Linux has no media-critical duty left on the board -- the MMU, supervisor mode, the 32 KB L2, OpenSBI and the kernel boot chain are the buy-back that funds the plane.

## Work
- [ ] Define the minimal VexiiRiscv configuration (RV32, no MMU, no supervisor, minimal/no caches) and the baremetal firmware scope: boot/flash, bring-up pokes (CSR init, settime/UTC), diagnostics/UART; everything else is fabric.
- [ ] Replace the buildroot/Linux rootfs flow for the shipping shape with the baremetal image (the Linux flow stays buildable for the bring-up boards until #116 completes).
- [ ] The sound card goes optional, DEFAULT REMOVED (USER directive, 2026-08-19): the host-facing audio surface that exists only to serve the board's Linux ALSA driver (the PCM playback/capture rings, the AVTP-RX-to-ring monitor path, the host ring CSR faces) becomes an elaboration option like the tier-1 prune blocks, and the default build excludes it -- with no Linux there is no ALSA consumer, and the freed fabric joins the buy-back. The audio DATAPATH itself (AAF talkers/listeners, TDM/I2S wire front-ends, the crossbar) is NOT the sound card and stays.
- [ ] Measure the freed LUT/FF/BRAM on the real instrument and re-run the #114 option-ON verdict against the downgraded SoC.
- [ ] All Verilator/yosys/behave/builder gates green in the new shape; hostplane smoke adapted to the baremetal console.

## Acceptance
- [ ] A placed-and-routed option-ON build (fabric gPTP + baremetal core) meets timing on xc7a100t, with the measured record in docs.

## Dependencies
#114 lands the splice (option OFF) first; this ticket funds its option-ON verdict; #117 validates on silicon; #116 flips the defaults. Parent epic #110.

[...]

Issue #116, opened 2026-08-19
gPTP plane default ON: retire ptp4l, phc2sys and the statd mirror


Mister-M-alt:
Subticket of #110. The software-retirement round, after the splice proves itself.

## Work
- [ ] Flip the elaboration option to DEFAULT ON; the software path stays buildable behind the option for bring-up comparison until removal.
- [ ] Rootfs default drops ptp4l, phc2sys and the milan-statd GM mirror; the gptp2csr.sh CLKV lease and the publication chain (UDS/pmc -> statd -> devmem) retire -- the CSR words are fabric-driven.
- [ ] Docs follow: ARCHITECTURE_HW_SW_SPLIT.md, GM_LOSS_RECOVERY.md, the boot diagram and the feature ledger describe the fabric plane as the default.

## Acceptance
- [ ] The shipped rootfs starts no ptp4l/phc2sys; tu reads 0 from the fabric bank on a booted board; all gates green in both option states.

## Dependencies
#114 (the splice) lands first; #117 (silicon validation) proves the flip.

Mister-M-alt:
[A0] TAKEN
Branch: `116-gptp-default-on-retire-software` (PR #135, draft)
Authoritative references: `REQUIREMENTS.md` REQ-PTP-01..06 and acceptance §8; `docs/design/GPTP_PLANE.md`; `docs/design/TIME_SYNC.md`; `docs/integration/BAREMETAL_FIRMWARE.md`; IEEE 802.1AS-2020 time uncertainty/asCapable and Milan v1.2 continuous-output rules.
Interpreted scope: make the fabric gPTP engine the product RTL/builder default; keep one explicit option-off Linux comparison owner; route the selected publication/PHC state through CSR, protocol answers and AVTP validity; retire the software services in the sibling rootfs; update current authority docs without folding unrelated base regressions into this lane.
Validation plan: both ownership profiles through real builder/build/sweep argv, focused PHC/publication/CSR/wire mutations, full Verilator aggregate, Behave, lint ratchet, generated module matrix, documentation/ToC gates, and Yosys.
Blockers: sibling-rootfs retirement remains #116 acceptance work outside this repository; #117 owns two-board latency/sync/GM-loss evidence; donor README dependency is Mister-M-alt/FPGA-gPTP#13 / PR #14; base documentation regression is isolated as #159; two positive reviews including one external and explicit maintainer merge authorization are still required. No agent merge is authorized.

[...]

PR #126, opened 2026-08-19
Make the host PCM capture-ring an elaboration option (SNDCARD_P), #120 datapath lever

Mister-M-alt:
[A1] First increment of issue #120 (the baremetal downgrade, subticket of #110): makes the host-facing PCM capture-ring surface — the "sound card" that exists only to feed the board's Linux ALSA driver — an **elaboration option**, following the tier-1 prune pattern.

`SNDCARD_P` (milan_datapath, DEFAULT 1 = PRESENT, the no-regression axiom) gates the datapath's `m_axis_pcm` output surface (KL_pcm_route's DMA leg + the depacketizer's tkeep). When pruned, the port is tied to the exact value a build with no ring writer bound presents, and the DMA-leg drivers feed dead nets so synthesis removes them. Deliberately **out of scope of the gate**: `m_axis_pcm_tready`, the RENDER leg, the LPF/I2S DAC, and the crossbar — the AAF/TDM/I2S audio datapath is separate and bit-identical either way. No config flips the default in this PR; the SoC-side ring removal (`KL_pcm_ring_bram` + the DMA writer + the host CSR bank — the bulk of the buy-back) and the default-removed flip land with the baremetal shipping shape, exactly the sequencing the issue states (the Linux flow stays buildable until #116).

The lever is wired end to end per the tier-1 pattern: builder `OPTIONAL_BLOCKS` + `resource_instances` + a banked `RESOURCE_COSTS` row (honestly the datapath surface only — KL_pcm_route OOC ~4 LUT/2 FF, not overstated), the `milan_soc.py` `MILAN_OPTIONAL_BLOCKS` map + `--no-sound-card` argparse, and the AREA_BUDGET tier-1 table row — so gate 23d's builder↔SoC↔datapath↔docs consistency check sees 8 blocks.

Local validation:
- `milan_dp` all shapes PASS including the pruned shape (`-GSNDCARD_P=0`): a real AAF PDU reaches the datapath and the gated `m_axis_pcm` stays inert; the present shapes are byte-identical (5570/1622/1630/1622, 0 fail).
- `test_builder` gates 23a–23e PASS (prune posture, refusals, all-pruned estimate, the 8-block decorative-ABI consistency).
- RTL lint 99≤99 (unchanged), docs-check clean, traceability no-drift, Yosys portability.

Not this PR: the SoC ring removal (LiteX) and the config flip. No RTL datapath behaviour changes for any shipping build.

Refs #120.

---
**Overlap with #125:** PR #125 ("Ship the AX7101 bare-metal gPTP profile") implements the whole of #120 in one PR, including a full sound-card removal (the same `sound_card` config key, param `SOUND_CARD`, the shipping config flipped to `false`, plus the SoC ring removal). This PR is the narrower datapath-lever subset with a different param name (`SNDCARD_P`) and no config flip. They touch the same files, so only one lands cleanly. Flagging so the maintainer can decide — if #125 is the chosen path, close this in its favour.

### DIFF
--- a/docs/design/AREA_BUDGET.md
+++ b/docs/design/AREA_BUDGET.md
@@ -199,6 +199,7 @@ enforces that on all three shipped configs.
 | RX address filter | `RXFILT_P` | `--no-rx-mac-filter` | `rx_mac_filter` | the port is promiscuous, or filtering is done in software |
 | PCM low-pass | `LPF_P` | `--no-render-lpf` | `render_lpf` | render path only, and every digital acceptance measurement is taken upstream of it |
 | datapath probe groups | `DPROBES_P` | `--no-datapath-probes` | `datapath_probes` | the APRB (0x8B4–0x8C4) and PBK (0x8C8–0x8D0) groups are closed-finding diagnostics (fabric-listener blocker + item-7 chain, both TB-pinned since); the range reads 0 on a pruned build — the LTAP precedent (added 2026-07-29 to unlock AreaOptimized_medium on the AX, whose AreaOptimized_high timing was congestion-systemic) |
+| sound card | `SNDCARD_P` | `--no-sound-card` | `sound_card` | there is no Linux ALSA consumer for the host PCM capture ring (issue #120: the baremetal shape). This lever gates only the datapath `m_axis_pcm` OUTPUT surface (KL_pcm_route's DMA leg); the AAF/TDM/I2S datapath and the DAC render leg are separate and stay. The BULK of the sound card is the SoC-layer ring (`KL_pcm_ring_bram` + the DMA writer + the host CSR bank), removed and re-measured in the SoC-side follow-up — the figure below is the datapath surface only (added 2026-08-19) |
 
 #### What each one is worth — MEASURED, and every figure a yosys ESTIMATE
 
--- a/hdl/milan/milan_datapath.sv
+++ b/hdl/milan/milan_datapath.sv
@@ -268,6 +268,20 @@ parameter int PB_PREFILL_C = 0,    //! playback prefill release (0 = midpoint;
   //! diagnostics whose counters/latches die with the parameter; the CSR
   //! range reads 0 on a pruned build (the LTAP precedent).
   parameter int DPROBES_P = 1,
+  //! Host-facing PCM capture-ring surface (the "sound card": issue #120).
+  //! The m_axis_pcm output port is the datapath's contribution to the ring
+  //! the SoC layer binds to feed the board's Linux ALSA driver. With every
+  //! control plane and time sync in fabric, a baremetal (no-Linux) shape has
+  //! no ALSA consumer, so this becomes an elaboration option. 0 prunes the
+  //! output surface: the port is tied inert (its documented ring-disabled
+  //! value -- exactly what a build with no ring writer bound produces), the
+  //! KL_pcm_route DMA leg feeding it goes dead and is optimised away, and the
+  //! audio DATAPATH (the render leg, the LPF/I2S DAC, the crossbar) is
+  //! untouched -- render backpressure is left on m_axis_pcm_tready, which the
+  //! SoC still drives. Defaults 1 = PRESENT (the no-regression axiom); the
+  //! SoC-side ring removal and the default-removed flip follow with the
+  //! baremetal shipping shape.
+  parameter int SNDCARD_P = 1,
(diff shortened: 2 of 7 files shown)


Mister-M-alt:
[A1] Validated on branch head 3ffed6ae (== origin). Full local bar: Verilator sweep 2,118,216 checks / 0 failures (milan_dp all shapes incl. -GSNDCARD_P=0; present shapes byte-identical); Yosys 46 tops PASS (tied-input + tap-purity clean); test_builder gates 23a-e PASS (8-block consistency); lint 99<=99; docs-check + traceability clean. No RTL datapath behaviour changes for any shipping build (SNDCARD_P defaults PRESENT). See the PR body re the #125 overlap.

Mister-M-alt:
[R1] Cleared-context review at head 3ffed6ae. **POSITIVE, no blockers**, with honest informational findings.

- **Audio-datapath safety — PASS.** The generate block assigns ONLY the five m_axis_pcm output ports; the render leg, LPF, I2S feed-mux/playback, crossbar, depacketizer, and m_axis_pcm_tready are untouched. KL_pcm_route gets identical inputs in both shapes, so its render outputs are bit-identical.
- **No-regression — PASS (empirical).** The SNDCARD_P=1 arm drives the ports from the same nets as before; the full milan_dp build exits 0 with all present shapes PASS and the pruned shape PASS. SoC emits p_SNDCARD_P=0 only when pruned, so a default top stays byte-identical.
- **Builder consistency — PASS.** Gates 23a-e green; 23d pins sound_card across builder/soc/datapath/AREA_BUDGET (8 blocks). The only test_builder failure is the pre-existing environmental milan-tests-avb entity-conf staleness (fails on clean dev too), untouched by this PR.

Findings (informational, non-blocking):
1. **The pruned sound-card check is partially vacuous.** A reviewer mutation confirmed: leaving tvalid wired through (a realistic forgotten-gate regression) STAYS GREEN; only a self-asserting tie (1'b1) turns it red. The bindless minimal harness can't accept an AAF PDU into the ring, so nothing proves the ring WOULD have carried payload but for the gate. Same structural-inert limitation as the LTAP/MCSRV/DAC prune checks; disclosed in the code, but materially weaker than a bound test.
2. `prune_sound_card` -4 LUT/-2 FF charges the whole N=1 KL_pcm_route while the module stays instantiated; the true datapath-only saving is ~1-2 LUT / ~0 FF. Correctly labeled an upper-bound estimate and feeds no budget verdict.

**Disposition:** this PR is sound but is the narrow datapath-only lever, SUPERSEDED by #125 — which gates inside KL_pcm_route, does the full removal + config flip + SoC ring, and carries a non-vacuous mutation-confirmed sim_soundcard_off.cpp (the exact weakness in finding #1). With the 50 MHz baremetal path (#125) approved, closing this in favour of #125.

Mister-M-alt:
[A1] Superseded by #125 (MERGED), which does the full #120 including a cleaner sound-card removal (gate inside KL_pcm_route, config flip, SoC ring, and a non-vacuous mutation-confirmed sim_soundcard_off.cpp — closing the partially-vacuous-check finding on this PR). Closing in its favour per the 50 MHz baremetal decision. R1 review of this lever remains POSITIVE for the record.

Mister-M-alt:
[R1] Correction to the review above — the reviewer revised to **NEGATIVE** on a re-check, and it was right: this PR's `--no-sound-card` was a DECORATIVE flag. `args.no_sound_card` was never threaded into the `optional_blocks` dict in `milan_soc.py` `main()`, and `dict.fromkeys(MILAN_OPTIONAL_BLOCKS, True)` treats a missing key as PRESENT — so `p_SNDCARD_P=0` would never have been emitted while the build plan and the resource estimate both claimed the prune. That is the `p_AAF_PLAYBACK` silent-no-op class, and the exact false-success this repo has been bitten by before.

Blast radius was nil (no config declared `sound_card`, and the TB drove `-GSNDCARD_P=0` directly, bypassing milan_soc), and the PR is closed as superseded — but the finding stands against this diff and is recorded here rather than lost.

Checked and CLEAR: the merged #125 does NOT have this defect. It threads `sound_card` as a first-class constructor parameter and inverts the default (`parameter int SOUND_CARD_P = 0` in milan_datapath, `p_SOUND_CARD_P = 1` set only when present), so `sound_card: false` genuinely elaborates the sound card away. The shipping bitstream is what the plan says it is.

PR #128, opened 2026-08-19
gPTP shadow: shed guard so an ingress burst cannot lap the ts ring (#122)

### LINKED ISSUE #122: gPTP shadow: unguarded ingress-ts ring overwrites live stamps under burst
Part of epic #110. Found by [R0]'s review of PR #119 (finding 3) at bf299d9a; the defect is in the plane wrapper, not the donor engine.

## The defect

`KL_gptp_shadow`'s ingress-timestamp side FIFO is a raw ring: a push on `ff_good_w` (the frame FIFO's per-delivered-frame commit strobe) increments the write pointer with NO full protection (`hdl/ieee8021as/gptp_plane/KL_gptp_shadow.sv`, the tsf_wp_r block). Depth is `1 << TS_FIFO_LOG2_P` = 32 entries, but the 2048-byte frame FIFO can hold ~128 two-beat frames, and the serializer drains at 1 B/clk. Under a back-to-back small-frame burst, committed-but-unserialized frames exceed 32 and the ring laps: LIVE entries are overwritten and later frames are served STALE arrival timestamps.

Measured by [R0]: a focused 40-frame, two-beat 0x88F7 burst produced 40 pushes and 40 pops with four timestamp mismatches. Push/pop PAIRING held (the commit-pulse transport is sound); the VALUES were lapped.

## The fix (frame-granularity shed guard, not a bigger ring)

Depth-matching the ring to the frame FIFO (128 x 64 b) buys LUTRAM the area budget does not have. The cheap exact guard: at a frame's first tap beat (FW_HEAD0), if the ring is full, shed the WHOLE frame before it enters the frame FIFO. The check is exact at sof, not conservative: tap beats are serial, so every previously accepted frame has already committed (pushed) by the next frame's sof; occupancy at sof is the true pending count and the only future push is this frame's own.

- Widen `tsf_wp_r`/`tsf_rp_r` by one bit (wrap-safe occupancy); full = `(wp - rp) == TSD_C`.
- FW_HEAD0 + full: suppress `fw_valid_w` for the whole frame (atomic: nothing enters the frame FIFO), transit to FW_SKIP.
- Count the shed in `dbg_tap_drop_o` ONLY when the shed frame turns out to be 0x88F7 (the EtherType verdict lands at beat 1 even in the shed path; counting non-gPTP sheds would poison the diagnostic).
- One-beat runts at full stay uncounted, matching today's bad-frame reclaim.

The single `ts_arr_r` stage register is NOT part of the defect: commit-vs-next-sof ordering plus NBA pre-edge reads keep it paired (the same analysis as the donor's runt commit).

## Verification

- New gptp_shadow phase replaying [R0]'s burst: 40 back-to-back two-beat 0x88F7 frames; the popped-stamp sequence must be an order-preserving subsequence of the harness-recorded per-frame arrival stamps (shed frames removed), and pops + gPTP sheds must total 40 after drain. Red on today's RTL (the lap breaks the subsequence law).
- Mutation: guard removed = red; drop counter blind to sheds = red.
- `docs/design/GPTP_PLANE.md` timestamps section gains the shed rule sentence.

Acceptance: the burst phase green at the fixed RTL and red at the unguarded revert, full sweep green, no new lint findings.

Mister-M-alt:
[A1] Fixes issue #122: `KL_gptp_shadow`'s ingress-timestamp side FIFO was an unprotected 32-entry ring pushed per committed frame. Under a back-to-back small-frame burst, committed-but-unserialized frames exceed 32 and the ring laps — later frames are served STALE arrival stamps ([R0] measured four mismatches on a 40-frame 0x88F7 burst).

## The fix — a frame-granularity shed guard
At a frame's first tap beat, if the ring is already spoken for, the WHOLE frame is shed before it enters the frame FIFO (atomic: no beat enters, so no commit can lap the ring). Reuses the existing FW_HEAD1 EtherType verdict via a `shed_r` flag — no new FSM state. The shed is counted in `dbg_tap_drop_o` only once that verdict confirms 0x88F7; one-beat runts at full stay uncounted.

**"Spoken for" is two terms, and getting there took two corrections the ticket's recipe did not anticipate — both measured, not reasoned:**

1. The ticket specifies `full = (wp - rp) == TSD_C` on the ring pointers, asserting "every previously accepted frame has already committed by the next frame's sof". That is **not true for back-to-back 2-beat frames**: the push rides the frame FIFO's commit, which lags the tap, so frames already taken are invisible and the guard sheds too late. Measured: **38 frames into a 32-entry ring — still lapped.**
2. Counting at the tap and releasing on pop closes that gap but **leaks**: a frame the FIFO itself discards (oversize, or arriving full) never commits, so it never pushes and never pops. One slot leaks per discarded frame, and after 32 the guard wedges shut and the plane goes **permanently deaf** — worse than the lap it was fixing. Measured: **40 oversize frames, then nothing was ever accepted again.** The full 2.1M-check sweep was green against that version; only a purpose-built phase caught it.

So the guard uses the ring's exact occupancy **plus** frames that have entered the frame FIFO and are not yet **resolved** by it. A frame resolves exactly once — good (a future push), bad, or overflow (no push) — so the count can neither leak nor underflow whatever the FIFO does with it. It counts non-gPTP frames too, which only makes the guard slightly conservative: they resolve in a couple of cycles and never push. `status_bad_frame`, previously unconnected, is now wired for that.

## Verification
- New `gptp_shadow` burst phase: 40 back-to-back two-beat 0x88F7 frames; popped stamps equal pushed stamps in FIFO order (no lap), pushed stamps strictly increasing, pops + gPTP sheds == 40, and the guard proven to actually shed.
- New no-leak phase: 40 oversize frames the FIFO discards, then a normal frame must still reach the engine.
- **24/24 checks PASS, and all three mutations bite:** guard disabled → lap + no-shed RED; drop counter blind to sheds → accounting RED; leak reintroduced (only `good_frame` releases) → no-leak RED.
- `docs/design/GPTP_PLANE.md` carries the shed rule and both failure modes, so the next reader does not re-derive them.
- lint 99≤99 (no new violations), docs-check clean, traceability no-drift. Full sweep + Yosys on the corrected RTL in progress; evidence to follow.

Closes #122.


Mister-M-alt (commented):
[R0] REQUEST CHANGES / DO NOT MERGE at head `b1e0b37165cdf9476bfddceda9d1f786a9d74c00`.

1. [P1] FIFO-dropped frames permanently leak the new pending reservation and eventually disable gPTP RX. `tap_accept_w` increments `pend_r` at `KL_gptp_shadow.sv:289-290,306-308`, before the frame FIFO commits the frame, but the only decrement is `tsf_pop_w`. The configured `axis_fifo` may later discard that accepted 0x88F7 frame through `DROP_OVERSIZE_FRAME` or `DROP_WHEN_FULL` (`KL_gptp_shadow.sv:228-265`). Such a frame gets no `ff_good_w`, no timestamp-ring entry and no pop, so one pending slot is leaked forever.

This is reproduced on the exact head, not hypothetical: the existing phase-6 oversize leaves `pend_r=1`; phase 7 leaves it at 8; after phase 8 drains for 40,000 cycles it is still 8. Repeating the existing oversize stimulus 32 times drives `pend_r` from 1 through 32; at 32, the following valid Sync/Follow_Up are shed and the published offset stays stale. `full_w` remains asserted until reset. Retire the pending reservation on the correctly aligned FIFO-drop verdict (including simultaneous accept/pop/drop cases), then prove at least 32 accepted-but-dropped frames followed by a normal frame that must commit and pop. Also assert pending returns to zero after a complete drain.

2. [P2] The new test and documentation hide that path. `sim_main.cpp:383-419` uses only small frames that commit, so its push/pop law can pass while dropped-frame reservations accumulate. `GPTP_PLANE.md:69-76` calls pending “frames accepted at the tap but not yet popped”, but those dropped frames are never poppable. Update both the recovery test and the stated accounting rule with the fix.

3. [P2/process] The contribution evidence/template is incomplete. The body does not use the required Status / Description / how-to-reproduce / how-to-validate / DoD structure and embeds `23/23 checks PASS`; GitHub has no separate author self-test comment. Move exact-head evidence to a `[A1]` comment and populate the PR template before re-review.

Positive controls: exact head is one direct commit above current `dev`; focused `gptp_shadow` is green at 23/23; commit style, reset coverage, naming, documentation scope and lint-waiver posture are otherwise sound. The green focused test is precisely why the repeated-drop negative control is required.

GitHub will not allow this authenticated account to submit REQUEST_CHANGES on its own PR, so this COMMENTED review is the negative verdict and must not be counted as approval.

Mister-M-alt (commented):
[R0] ROUND 2 at head `61f80592a07f4d37321c2d977efeb82ce4719ae8`: the P1 pending-counter leak is RESOLVED, but this is not yet a positive review.

The two-term accounting is coherent: `fifo_enter_w` reserves a frame at its first FIFO beat, `ff_good_w | ff_bad_w | ff_ovf_w` resolves every FIFO disposition, the widened ring pointers provide committed occupancy, and simultaneous enter/resolve nets to zero. The exact-head focused suite passes 24/24, including the 40-drop recovery mutation that failed the prior head.

Remaining changes:

1. [P2/test claim] The new phase says the post-drop frame must “reach the engine” (`sim_main.cpp:427-429`), but the only assertion clears and checks `g_pushed` (`:439-446`), which proves FIFO commit, not timestamp pop, serializer delivery, or engine state. Clear/check `g_popped` as well and grade a protocol-visible state change, or narrow the documented claim to what is actually tested. The original R0 request explicitly required commit and pop after the drop storm.

2. [P2/test documentation] `tb/verilator/gptp_shadow/README.md` still documents only phases 1-7 and five old mutations. It omits the new ring-lap phase, drop-leak recovery phase, and all three issue-#122 mutations. Update the runnable suite index so the committed test contract matches the 24-check harness.

3. [P2/process] The PR still lacks the required Status / Description / how-to-reproduce / how-to-validate / DoD template and has no separate `[A1]` exact-head self-test comment. The narrative body still embeds pass results. Complete the contribution evidence before another review round.

The former P1 must not be counted as outstanding after this head; the items above are the remaining round-2 findings. Any changed head needs a fresh review, and this comment is not approval.

Mister-M-alt:
[A1] Review answered at `cf208088`. Both findings were real; the blocking one was masking the suite.

**B1 — the ring difference is now taken at pointer width.** The review is exactly right about the semantics: `N'(a - b)` sets the CONTEXT size of the whole expression, so casting the difference straight to the sum's width evaluated `wp - rp` at 9 bits and threw away the wrap-safe extra pointer bit the line above it had just added. On every wrap the term read ~449-511 instead of the true occupancy, `full_w` latched, and the plane shed frames it had room for. Fixed by taking the difference in its own `TS_FIFO_LOG2_P+1`-bit net first, then widening.

The important part is the second-order effect the review found: **B1 was masking the suite.** Both mutations it named were 24/24 GREEN on the shipped RTL and are RED now:
- `>= TSD_C` -> `> TSD_C` (admits a 33rd entry): **5 FAIL**
- `occ_w` = ring pointers only, i.e. the exact lap bug this PR exists to fix: **5 FAIL**

**B2 — declarations hoisted above first use.** Confirmed with the real tool, both directions: `xvlog -sv` on the superseded head gives `ERROR: [VRFC 10-3380] identifier 'full_w' is used before its declaration` and `module 'KL_gptp_shadow' is ignored`; on `cf208088` it analyses clean. `dev` parsed clean, so this was a genuine regression that Verilator accepts silently — my local bar could not see it, and it would have made this file ineligible for the xsim benches.

**I2 — the shed path is now leak-locked.** New phase 10 repeats the burst four times and requires the per-burst shed count not to grow, plus the no-lap and pops+sheds==40 laws on every repetition. The mutation the review named (counting shed frames as FIFO entries, one leaked slot per shed) went from surviving to **1 FAIL**.

**I3 — `unres_r` narrowed** to `TS_FIFO_LOG2_P+1` bits with the real bound stated (an enter is blocked at `occ == TSD_C` and `unres <= occ`, so it never exceeds TSD_C); the saturations stay as belt-and-braces on a wedge that would cost the plane every frame, now noted as unreachable by construction rather than implied to be load-bearing.

**I8 — the pop is now structurally unable to pass the write pointer** (`tsf_pop_w & (tsf_wp_r != tsf_rp_r)`). It cannot fire today, but with a pointer-difference compare a stray pop is the difference between a stale stamp and a permanently shedding guard, and that is one AND gate.

**I7 + the B1 mechanism are documented** in `GPTP_PLANE.md`, so the next reader does not rediscover either the cast trap or the runt-at-full carve-out.

Suite is **33/33**, and all six mutations bite: guard disabled (6 FAIL), drop counter blind (6 FAIL), resolve=good-only (1 FAIL), `>=`->`>` (5 FAIL), ring-only occupancy (5 FAIL), shed-counted-as-entry (1 FAIL). lint 99<=99, docs clean. Full sweep and Yosys re-running on this head; evidence to follow.

On I4 (area): noted, and the narrowing above should recover part of the +38 LUT. The shape is at ~99.99% slices after the bare-metal round, so the Vivado number is the one that counts and this PR does not produce one.

Mister-M-alt:
[A1] Validated at head `642950a7`. Base `origin/dev` (`213210c3`, unchanged) is an ancestor, so the head tree IS the candidate merge tree. Head == origin tip, checked before merge.

- **Full Verilator sweep: 2,118,237 checks, 0 in-suite failures**, 52/52 suites (`gptp_shadow`, `gptp_plane`, `milan_dp` all PASS). The count moves +11 against `dev`, exactly the phases added here.
- **Yosys portability: 46/46 tops PASS**, tied-input and observer-purity gates PASS (run on `cf208088`; `hdl/` is byte-identical at `642950a7`, the last commit being test-only).
- **Vivado front-end: `xvlog -sv` analyses the module clean.** Verified in both directions -- the superseded revision is rejected with `VRFC 10-3380` and the module ignored.
- RTL lint 99<=99 (no new violations), docs-check clean, traceability no-drift.
- **Suite: 35/35, and eight mutations bite** -- guard disabled (6 fails), drop counter blind to sheds (6), `fifo_resolve_w` = good-only (1), `>=`->`>` (5), occupancy = ring pointers only (5), shed frames counted as FIFO entries (1), a shed frame's beat 1 admitted (10), and the pop gate weakened to the low pointer bits (2). The two in bold above were GREEN before the arithmetic fix -- that bug was masking the suite, which is the main reason this PR took three rounds.

Remote: `bdd-conformance`, `docs-check`, `docs-check-no-git` and `wire-accountability` pass; `verilator-suites` passed on the previous push of this branch, which is also the first confirmation that #127's CI fix works in the real environment (the same job was red on #125).

Mister-M-alt:
[A1] Final validation at head `c18f0572`, base `origin/dev` `213210c3` (unchanged, an ancestor -- the head tree IS the candidate merge tree). Head == origin tip, verified immediately before merge.

- **Full Verilator sweep: 52/52 suites, 0 failed, 0 timed out, 2,118,242 checks, 0 in-suite failures.**
- **Yosys portability: 46/46 tops PASS**, tied-input and observer-purity gates PASS. Run on `cf208088`; the delta since then is comment-only in `hdl/` (verified by filtering the diff), so the result stands for this head.
- **Vivado `xvlog -sv`: 0 errors** -- checked both ways, the superseded revision is rejected with `VRFC 10-3380`.
- lint 99<=99, docs-check clean, traceability no-drift.
- **Suite 40/40 with ten mutations biting**, including every one that survived an earlier revision: guard disabled (6), `>=`->`>` (5), occupancy = ring pointers only (5), resolve = good-only (1), shed counted as an entry (1), drop counter blind to sheds (6), shed frame's beat 1 admitted (10), pop gate weakened to index bits (2), shed frame routed to FW_BODY (5), shed count blind to EtherType (1).

Two cleared-context reviewers, plus a re-review round after each answer. The final independent pass proved the RTL correct against a shadow-FIFO golden model over eight stress phases (584 pushes / 552 pops, including a mid-burst reset), with the ring reaching exactly its 32-slot capacity and never a 33rd.

Worth recording for the next reader: this change took four rounds, and **three of the four defects were in the TESTS, not the RTL** -- a guard that leaked, an arithmetic cast that silently masked two mutations, a vacuous law, and a shed path never exercised past two beats. The RTL that ships here is the smallest part of the work.

Issue #10, opened 2026-05-09
AVTP Packet Switch implementation


cemaldogann:
- KL_avtp_common_parser is parsing the incoming avtp packets and placing TDEST and TUSER.
- The AVTP Packet Switch will convey incoming packets from the common parser to the downstream by differentiating them via TDEST.

<img width="685" height="477" alt="Image" src="https://github.com/user-attachments/assets/45dcc5aa-e573-46e4-b2ab-7493cbb33f21" />

PR #133, opened 2026-08-20
Bump the tsn-gen pin so fuzz_ptp runs in CI

Mister-M-alt:
[A1] Makes the merged `fuzz_ptp` campaign REPRODUCIBLE in CI, answering [R0]'s finding on #117.

## The defect

CI pins tsn-gen at `07534e64`, and the campaign's eight 802.1AS model YAMLs are not in that revision, so `fuzz_ptp` cannot run there. Its 355 checks have never executed anywhere except one developer machine. PR #127 stopped the crash (the campaign now skips cleanly instead of raising `KeyError: pdelay_req`), but skipping is not running -- and as [R0] put it, CONTRIBUTING makes the remote job optional for merge, it does not make the committed oracle allowed to be unusable.

The root cause was NOT a stale pin. `07534e64` was the tip of `kebag-logic/tsn-gen` main; the models existed only as unpushed local commits. They are now upstream in kebag-logic/tsn-gen#11, reviewed there against IEEE 802.1AS-2011 clause by clause (that review found and fixed three real defects in them, including a 96-bit field that silently truncated to 64 bits on both encode and decode).

## The change

One line: `TSN_GEN_REV` in `.github/workflows/rtl.yml` moves to the reviewed tsn-gen head. No RTL, no test logic.

## Validation -- the CI recipe, replicated verbatim

Rather than assume the pin resolves, I ran the workflow's own steps against a clean clone:

```
git clone --filter=blob:none https://github.com/kebag-logic/tsn-gen.git <tmp>
git -C <tmp> checkout affe6a3549a765cd98877decdb80db42a7ea432f     # OK
git -C <tmp> submodule update --init --depth 1 --recursive external/rapidyaml
cmake -S <tmp> -B <tmp>/build -DCMAKE_BUILD_TYPE=Release -DENABLE_PARSER_TESTS=OFF
cmake --build <tmp>/build --target packet_gen --parallel           # OK
```

Checkout by SHA resolves, the build succeeds, and the 7 PTP model files are present at that revision. Then the campaign itself, pointed at that clone:

**`TSN_GEN_ROOT=<the CI-cloned tree> make ptp` -> 352 pass, 0 fail, 9 known gaps.**

So the campaign runs at the exact committed pin, which is what the finding asked to see recorded.

## Note before merge

tsn-gen#11 is not merged yet, so this pin currently names a commit on that PR's branch. It resolves today (a full clone fetches it), but the pin must be re-pointed at the MERGE commit before this lands, or it risks dangling if the branch is deleted. I will re-pin and re-run this evidence at that point rather than merge as-is.

The 9 tracked gaps are the FPGA-gPTP engine defects the campaign surfaced (#6-#10); they print, they do not fail.

Refs #117.

### DIFF
--- a/.github/workflows/rtl.yml
+++ b/.github/workflows/rtl.yml
@@ -17,7 +17,7 @@ env:
   VERILATOR_VERSION: v5.050
   # The tsn_fuzz suite is a required field campaign. Pin its packet generator
   # so suite accounting never degrades to an uncounted local-only skip.
-  TSN_GEN_REV: 07534e64218ebbf486cc651d75727369f0ed3c83
+  TSN_GEN_REV: affe6a3549a765cd98877decdb80db42a7ea432f
 
 jobs:
   # The full Verilator suite sweep needs verilog-axis and protocol-processor.
--- a/.github/workflows/rtl.yml
+++ b/.github/workflows/rtl.yml
@@ -17,7 +17,7 @@ env:
   VERILATOR_VERSION: v5.050
   # The tsn_fuzz suite is a required field campaign. Pin its packet generator
   # so suite accounting never degrades to an uncounted local-only skip.
-  TSN_GEN_REV: affe6a3549a765cd98877decdb80db42a7ea432f
+  TSN_GEN_REV: fde652067d3618f47fba2c2e8eb05f6c9f6fa97f
 
 jobs:
   # The full Verilator suite sweep needs verilog-axis and protocol-processor.


Mister-M-alt:
[A1] Re-pointed and re-validated at head `6e3aef15`. tsn-gen#11 is MERGED, so the pin is now `fde652067d3618f47fba2c2e8eb05f6c9f6fa97f` -- the merge commit on tsn-gen `main`, not a PR branch head, so it cannot dangle if a branch is deleted.

Evidence re-run at that exact pin, not carried over from the previous head:
- CI recipe verbatim (clone --filter=blob:none, checkout by SHA, rapidyaml submodule, cmake, build packet_gen): all steps OK, 7 PTP model files present, and the SHA is reachable from `origin/main`.
- `TSN_GEN_ROOT=<that clone> make` in tb/verilator/tsn_fuzz: **164 AAF + 353 gPTP + 2 traceability contracts, 0 failures**, no SUITE-SKIP.

That is the campaign running at the committed oracle pin, which is what [R0]'s finding on #117 asked to see recorded. Head == origin tip; base `origin/dev` unchanged and an ancestor, so this head tree is the candidate merge tree.

Issue #130, opened 2026-08-19
Make gate 23d prove every --no-* prune flag reaches its RTL parameter


Mister-M-alt:
[A0]

## Summary

A builder configuration can request that an optional hardware block be removed, while the generated bitstream still contains it.

The existing gate 23d checks that all names exist and agree, but it does not check that `main()` actually passes each parsed `--no-<block>` flag into `MilanSoC`. If that handoff is missing, the build command, resource estimate, and build plan all claim that the block was removed, while the SoC silently keeps it.

This gap was exposed by superseded PR #126. It is present on `dev` even though all seven current optional blocks happen to be wired correctly.

## The path that must be proved

For every row in `sw/builder/endstation_builder.py::OPTIONAL_BLOCKS`, gate 23d must prove this complete chain:

```text
board.features.<name> = false
  -> endstation_builder emits --no-<block>
  -> argparse sets args.no_<block>
  -> main() passes not args.no_<block> to MilanSoC
  -> add_milan_datapath emits p_<PARAM>=0
  -> the matching milan_datapath generate arm is pruned
```

Gate 23d already proves the builder key, CLI declaration, SoC-to-RTL parameter map, RTL parameter/generate arm, and AREA_BUDGET documentation. This ticket adds the missing `main()` consumption check.

Most rows are passed through the `optional_blocks={...}` dictionary. `render_lpf` is an intentional compatibility exception and is passed through the dedicated `render_lpf=not args.no_render_lpf` constructor argument. Both forms must be covered.

## Failure mode

`add_milan_datapath` starts with every optional block present:

```python
blocks = dict.fromkeys(MILAN_OPTIONAL_BLOCKS, True)
```

A key omitted by `main()` therefore defaults to present. The CLI accepts the prune flag, but no `p_<PARAM>=0` reaches the generated LiteX Instance. This is a false-success build, not a visible error.

## Scope

- Extend `sw/builder/test_builder.py` gate 23d to verify that every current and future `OPTIONAL_BLOCKS` row consumes its matching `args.no_<block>` value in the `MilanSoC` call.
- Cover both the shared `optional_blocks` dictionary and explicitly threaded compatibility arguments such as `render_lpf`.
- Add negative mutation coverage that removes or neutralizes each handoff and proves the gate turns red.
- Keep the check source-only and fast; no LiteX/Vivado installation should be required.
- Do not change RTL or the shipping block posture.

[...]

PR #134, opened 2026-08-20
Prove every optional-block prune flag is consumed

Mister-M-alt:
[A0]

## Status

🟢 Green — `130-prove-prune-flags-consumed` → `dev`; ready for cleared-context review.

## Description

Gate 23d now proves the missing middle link in every optional-block prune:

- Parses the real `main()` function and its single `MilanSoC` call with Python AST.
- Requires every `OPTIONAL_BLOCKS` flag to reach that call as `not args.no_<block>`.
- Accepts both the shared `optional_blocks={...}` route and the dedicated `render_lpf=` compatibility route.
- Adds one negative mutation per current block; neutralizing any handoff must fail by block name.

This is a source-only test change. RTL and the shipping block posture are unchanged.

Refs #130.

```bash
git clone git@github.com:kebag-logic/milan-fpga.git
cd milan-fpga
git switch 130-prove-prune-flags-consumed
git submodule update --init gptp-processor protocol-processor third_party/verilog-axis
```

## How to validate

```bash
python3 -c 'import sys; sys.path.insert(0, "sw/builder"); import test_builder; test_builder.test_optional_block_names_reach_the_rtl(); test_builder.test_optional_block_consumption_gate_bites()'
python3 -m pytest -q sw/builder/test_builder.py
python3 scripts/docs_check.py
git diff --check origin/dev...
```

Pass criteria: every current handoff is traced, every disconnected-handoff mutation is rejected by name, all builder tests pass, lint remains within its ratchet, documentation checks are clean, and the diff has no whitespace errors.

## Definition of Done

- [x] Every current and future `OPTIONAL_BLOCKS` row must consume its matching argparse destination in the `MilanSoC` call.
- [x] Shared-dictionary and dedicated compatibility routes are both covered.
- [x] All seven current handoffs have negative mutation coverage.
- [x] The complete local verification bar for this source-only test change is green.
- [ ] Two positive cleared-context reviews, including one external, are complete.
- [ ] Candidate merge result is revalidated against the then-current `dev`.
- [ ] Maintainer merge, post-merge containment, manual issue close, and project Done transition are complete.

### DIFF
--- a/sw/builder/test_builder.py
+++ b/sw/builder/test_builder.py
@@ -3686,8 +3686,13 @@ def test_optional_block_names_reach_the_rtl():
           "`p_<PARAM> = 0` under `if not blocks[k]` == a milan_datapath "
           "`parameter int X = 1` guarding a generate WITH an else arm == the "
           "AREA_BUDGET tier-1 table row (the p_AAF_PLAYBACK silent-no-op "
-          "class, closed on the source chain; what elaboration hands "
-          "Instance() is gate 23f, task #144)")
+          "class, closed for the SPELLING of these statements. A statement "
+          "that rebinds or overrides a carrier without touching them - "
+          "reassigning optional_blocks, overriding a fold, deleting a "
+          "dp_params key - is invisible here by construction, because each "
+          "hop is checked in isolation and _reads() is a name-appearance "
+          "test. That family and what elaboration hands Instance() are both "
+          "gate 23f, task #144)")
 
     # gate 23e (task #65): the loopback lane rides the SAME decorative-ABI
     # risk but the OPPOSITE polarity - it is an ADD, default OFF, so it is
--- a/sw/builder/test_builder.py
+++ b/sw/builder/test_builder.py
@@ -3286,6 +3286,79 @@ def _is_not_arg(node, dest):
         node.operand.value.id == "args"
 
 
+def _soc_tree(soc):
+    """The milan_soc.py module AST, from source text or an AST already had."""
+    return soc if isinstance(soc, ast.Module) else \
+        ast.parse(soc, filename="sw/litex/milan_soc.py")
+
+
+def _declared_params(func):
+    """The parameter names a `def` declares EXPLICITLY, by name.
+
+    `*args` and `**kwargs` are excluded ON PURPOSE, and that exclusion is the
+    whole point. `MilanSoC.__init__` ends in `**kwargs` and forwards it
+    verbatim to LiteX `SoCCore.__init__`, which does NOT raise on a keyword it
+    does not know: it logs one warning and carries on. So a prune handed over
+    as a top-level `maap=not args.no_maap` would be accepted by argparse,
+    accepted by the constructor, mentioned once in a wall of build log, and
+    shipped as a bitstream with KL_maap still in it. That is p_AAF_PLAYBACK
+    again, so a keyword `**kwargs` would swallow MUST fail closed here.
+    """
+    a = func.args
+    return {p.arg for p in (*a.posonlyargs, *a.args, *a.kwonlyargs)}
+
+
+def _reads(node, name):
+    """True when the expression still READS the local `name` somewhere inside.
+
+    Deliberately a dependency question, not a spelling one: `render_lpf`,
+    `bool(render_lpf)` and `bool(render_lpf) and x` all pass, `True` does not.
+    """
+    return node is not None and any(
+        isinstance(n, ast.Name) and n.id == name for n in ast.walk(node))
+
+
+def _sole(nodes, what):
+    """The one node there must be, or a failure that says what was missing."""
+    assert len(nodes) == 1, \
+        f"milan_soc.py: expected exactly one {what}, found {len(nodes)}"
+    return nodes[0]
+
+
+def _blocks_key(node, key):
+    """`blocks[<key>]`, where key is a string literal or a variable name."""
+    if not (isinstance(node, ast.Subscript)
+            and isinstance(node.value, ast.Name)
+            and node.value.id == "blocks"):
+        return False
+    idx = node.slice
+    return (isinstance(idx, ast.Constant) and idx.value == key) or \
+           (isinstance(idx, ast.Name) and idx.id == key)
+
+
+def _class_of(tree, name):
+    return _sole([n for n in tree.body
+                  if isinstance(n, ast.ClassDef) and n.name == name],
+                 f"class {name}")
+
+
+def _def_of(scope, name):
+    return _sole([n for n in scope.body
+                  if isinstance(n, (ast.FunctionDef, ast.AsyncFunctionDef))
+                  and n.name == name], f"def {name}()")
+
+
+def _sole_call(scope, callee, where):
+    """The keywords of the SINGLE `callee(...)` call inside `scope`."""
+    call = _sole([n for n in ast.walk(scope)
+                  if isinstance(n, ast.Call) and isinstance(n.func, ast.Name)
+                  and n.func.id == callee], f"{callee}(...) call in {where}")
+    assert not any(kw.arg is None for kw in call.keywords), (
+        f"milan_soc.py {where}: {callee}(...) splats `**` - the prune chain "
+        "has to be readable in the source, not assembled at run time")
+    return {kw.arg: kw.value for kw in call.keywords}
+
+
 def _optional_block_cli_consumption(soc):
     """Map every OPTIONAL_BLOCKS row to its consumption route in main().
 
@@ -3293,8 +3366,13 @@ def _optional_block_cli_consumption(soc):
     reference in a comment or helper must not prove that argparse's value
     reaches the MilanSoC instance. Most blocks ride `optional_blocks`; the
     older render_lpf ABI is intentionally a dedicated constructor keyword.
+
+    A dedicated keyword only counts when `MilanSoC.__init__` DECLARES it (see
+    `_declared_params`). Without that test the escape hatch proves nothing:
+    move any row out of the dict onto its own keyword and the gate would print
+    the block as consumed while the build warns once and ships it unpruned.
     """
-    tree = ast.parse(soc, filename="sw/litex/milan_soc.py")
+    tree = _soc_tree(soc)
     mains = [n for n in tree.body
              if isinstance(n, (ast.FunctionDef, ast.AsyncFunctionDef))
              and n.name == "main"]
@@ -3305,8 +3383,10 @@ def _optional_block_cli_consumption(soc):
              and n.func.id == "MilanSoC"]
     assert len(calls) == 1, \
         f"milan_soc.py main(): expected one MilanSoC call, found {len(calls)}"
-    kwargs = {kw.arg: kw.value for kw in calls[0].keywords
-              if kw.arg is not None}
+    assert not any(kw.arg is None for kw in calls[0].keywords), \
+        "milan_soc.py main(): the MilanSoC call splats `**` - every prune " \
+        "handoff has to be readable in the source"
+    kwargs = {kw.arg: kw.value for kw in calls[0].keywords}
     block_dict = kwargs.get("optional_blocks")
     assert isinstance(block_dict, ast.Dict), \
         "milan_soc.py main(): MilanSoC has no literal optional_blocks dict"
@@ -3315,6 +3395,14 @@ def _optional_block_cli_consumption(soc):
         assert isinstance(key, ast.Constant) and isinstance(key.value, str), \
             "milan_soc.py main(): optional_blocks keys must be string literals"
         entries[key.value] = value
+    # the receiving end of every handoff below. `optional_blocks` itself is a
+    # keyword like any other: renamed away, `**kwargs` eats the whole dict.
+    declared = _declared_params(_def_of(_class_of(tree, "MilanSoC"),
+                                        "__init__"))
+    assert "optional_blocks" in declared, (
+        "MilanSoC.__init__ declares no `optional_blocks` parameter, so the "
+        "whole tier-1 dict is swallowed by its `**kwargs` and forwarded to "
+        "LiteX SoCCore, which warns once and ignores it")
 
     routes = {}
     for name, (flag, _param, _why) in eb.OPTIONAL_BLOCKS.items():
@@ -3322,6 +3410,14 @@ def _optional_block_cli_consumption(soc):
         if _is_not_arg(entries.get(name), dest):
             routes[name] = "optional_blocks"
         elif _is_not_arg(kwargs.get(name), dest):
+            assert name in declared, (
+                f"{name}: main() hands `{flag}` to MilanSoC as a dedicated "
+                f"`{name}=` keyword, but MilanSoC.__init__ declares no "
+                f"`{name}` parameter - its trailing `**kwargs` swallows the "
+                "value and forwards it to LiteX SoCCore, which logs ONE "
+                "warning, ignores it, and builds a bitstream with the block "
+                f"still in it. Put `{name}` in the optional_blocks dict, or "
+                "declare the parameter")
             routes[name] = f"{name} keyword"
         else:
             raise AssertionError(
@@ -3329,6 +3425,182 @@ def _optional_block_cli_consumption(soc):
                 f"`not args.{dest}` in the MilanSoC optional_blocks dict or "
                 f"a dedicated `{name}=` keyword; the missing key defaults "
                 "to PRESENT and the prune flag becomes a silent no-op")
+
+    # every --no-* flag main() DECLARES must be one main() also READS, and it
+    # must stay an off-by-default store_true. `default=True` on a --no-* flag
+    # inverts the lever in silence (an UNFLAGGED build takes the pruned path),
+    # and a flag nothing reads is decorative ABI wearing a CLI hat.
+    read = {n.attr for n in ast.walk(mains[0]) if isinstance(n, ast.Attribute)
+            and isinstance(n.value, ast.Name) and n.value.id == "args"}
+    for call in ast.walk(mains[0]):
+        if not (isinstance(call, ast.Call)
+                and isinstance(call.func, ast.Attribute)
+                and call.func.attr == "add_argument"
+                and call.args and isinstance(call.args[0], ast.Constant)
+                and str(call.args[0].value).startswith("--no-")):
+            continue
+        flag = call.args[0].value
+        opts = {kw.arg: kw.value for kw in call.keywords if kw.arg}
+        named = opts.get("dest")
+        dest = named.value if isinstance(named, ast.Constant) else \
+            flag.removeprefix("--").replace("-", "_")
+        act = opts.get("action")
+        assert isinstance(act, ast.Constant) and act.value == "store_true", (
+            f"{flag}: a --no-* flag must be a plain store_true, not "
+            f"{ast.unparse(act) if act is not None else 'a positional store'}")
+        dflt = opts.get("default")
+        assert dflt is None or (isinstance(dflt, ast.Constant)
+                                and (dflt.value is False
+                                     or dflt.value is None)), (
+            f"{flag}: `default={ast.unparse(dflt)}` inverts the flag - an "
+            "UNFLAGGED build would take the pruned path and ship the block "
+            "missing, which is the silent no-op with its sign flipped")
+        assert dest in read, (
+            f"{flag}: main() declares it and never reads `args.{dest}`, so it "
+            "is decorative - the argv records the intent and the gateware "
+            "never hears about it (the #130 shape, one hop earlier)")
+    return routes
+
+
+def _optional_block_prune_chain(soc):
+    """Trace every tier-1 prune from argparse to `p_<PARAM> = 0`, statically.
+
+    Gate 23d used to stop at the FIRST hop, `main()` -> `MilanSoC(...)`. Three
+    more hops sit between that call and the parameter the RTL reads, and all
+    of them live in this one file:
+
+        main()             -> MilanSoC(optional_blocks=..., render_lpf=...)
+        MilanSoC.__init__  -> MilanNIC(optional_blocks=..., render_lpf=...)
+        MilanNIC.__init__  -> add_milan_datapath(optional_blocks=...)
+        add_milan_datapath -> dp_params[f"p_{param}"] = 0
+
+    A value dropped, tied or inverted on any of them is invisible to a
+    spelling check, and its symptom is the #130 symptom exactly: the config
+    declares the prune, the plan tabulates it, the estimate books the saving,
+    and the bitstream ships the block. The emit guard is worse than a no-op if
+    inverted, because then an UNFLAGGED build prunes all seven and the
+    shipping bitstream quietly loses MAAP, the media-clock servo and the RX
+    MAC filter.
+
+    Task #144 (gate 23f) proves the same chain BEHAVIOURALLY, by elaborating
+    main() and spying on `Instance("milan_datapath", ...)`; that gate needs
+    migen + LiteX, so it SKIPS in CI. This one is text and AST only, which is
+    why it is the half that runs on the pyyaml-only docs job. The last link,
+    `dp_params` -> the parameters `Instance` is actually handed, is #144's:
+    no amount of reading can prove what elaboration does with the dict.
+    """
+    tree = _soc_tree(soc)
+    routes = _optional_block_cli_consumption(tree)
+    dedicated = sorted(k for k, r in routes.items() if r != "optional_blocks")
+    dict_routed = sorted(k for k, r in routes.items()
+                         if r == "optional_blocks")
+    soc_init = _def_of(_class_of(tree, "MilanSoC"), "__init__")
+    nic_init = _def_of(_class_of(tree, "MilanNIC"), "__init__")
+    dp_def = _sole([n for n in tree.body
+                    if isinstance(n, (ast.FunctionDef, ast.AsyncFunctionDef))
+                    and n.name == "add_milan_datapath"],
+                   "def add_milan_datapath()")
+
+    # hops 2 and 3: the carriers are FORWARDED, and each still READS the
+    # parameter it arrived in. Deleting `optional_blocks=optional_blocks`
+    # here defaults it to None one call further down, which un-prunes every
+    # dict-routed block; tying `render_lpf=True` does the same to the banked
+    # 428 LUT lever. Both are one deleted line and neither raises.
+    carriers = {"optional_blocks": dict_routed}
+    carriers.update({name: [name] for name in dedicated})
+    for src, callee, dst, where in (
+            (soc_init, "MilanNIC", nic_init, "MilanSoC.__init__"),
+            (nic_init, "add_milan_datapath", dp_def, "MilanNIC.__init__")):
+        passed = _sole_call(src, callee, where)
+        for carrier, carried in carriers.items():
+            assert carrier in _declared_params(src), (
+                f"{where} declares no `{carrier}` parameter, so the prune "
+                f"never arrives at all for: {', '.join(carried)}")
+            assert carrier in _declared_params(dst), (
+                f"{callee} declares no `{carrier}` parameter, so {where} "
+                f"cannot hand the prune on for: {', '.join(carried)}")
+            got = passed.get(carrier)
+            saw = f"`{carrier}={ast.unparse(got)}`" if got is not None else \
+                f"no `{carrier}=` keyword at all"
+            assert _reads(got, carrier), (
+                f"{where}: the {callee}(...) call passes {saw}, which no "
+                f"longer reads its own `{carrier}` parameter - the prune is "
+                f"dropped one hop short of the RTL for: {', '.join(carried)}")
+
+    # hop 4a: the map starts ALL PRESENT (AREA_BUDGET rule 1, at the source)
+    seed = _sole([n for n in ast.walk(dp_def) if isinstance(n, ast.Assign)
+                  and isinstance(n.targets[0], ast.Name)
+                  and n.targets[0].id == "blocks"], "`blocks = ...` seed")
+    assert (isinstance(seed.value, ast.Call)
+            and _reads(seed.value, "MILAN_OPTIONAL_BLOCKS")
+            and any(isinstance(a, ast.Constant) and a.value is True
+                    for a in seed.value.args)), (
+        f"add_milan_datapath: `{ast.unparse(seed)}` does not seed every block "
+        "PRESENT from MILAN_OPTIONAL_BLOCKS - anything else prunes on an "
+        f"UNFLAGGED build: {', '.join(sorted(routes))}")
+
+    # hop 4b: each dedicated keyword folds its OWN parameter into the map
+    for name in dedicated:
+        fold = _sole([n for n in ast.walk(dp_def) if isinstance(n, ast.Assign)
+                      and _blocks_key(n.targets[0], name)],
+                     f'`blocks["{name}"] = ...`')
+        assert _reads(fold.value, name), (
+            f"{name}: add_milan_datapath sets `{ast.unparse(fold)}`, which "
+            f"never reads its `{name}` parameter - the dedicated keyword is "
+            "inert and the block ships PRESENT whatever the CLI asked for")
+
+    # hop 4c: the dict fold still reads the caller's value
+    fold = _sole([n for n in ast.walk(dp_def) if isinstance(n, ast.For)
+                  and _reads(n.iter, "optional_blocks")],
+                 "`for ... in optional_blocks` fold")
+    assert (isinstance(fold.target, ast.Tuple) and len(fold.target.elts) == 2
+            and all(isinstance(e, ast.Name) for e in fold.target.elts)), \
+        "add_milan_datapath: the optional_blocks fold is not `for k, v in ...`"
+    kvar, vvar = (e.id for e in fold.target.elts)
+    row = _sole([n for n in ast.walk(fold) if isinstance(n, ast.Assign)
+                 and _blocks_key(n.targets[0], kvar)],
+                f"`blocks[{kvar}] = ...` inside the optional_blocks fold")
+    assert _reads(row.value, vvar), (
+        f"add_milan_datapath: `{ast.unparse(row)}` no longer reads the "
+        "caller's value, so EVERY dict-routed prune is a no-op: "
+        f"{', '.join(dict_routed)}")
+
+    # hop 4d: the emit guard, and above all its POLARITY. One `not` decides
+    # the sign of the whole tier-1 table for every block at once.
+    emit = _sole([n for n in ast.walk(dp_def) if isinstance(n, ast.For)
+                  and _reads(n.iter, "MILAN_OPTIONAL_BLOCKS")],
+                 "`for k, param in MILAN_OPTIONAL_BLOCKS.items()` emit loop")
+    assert (isinstance(emit.target, ast.Tuple) and len(emit.target.elts) == 2
+            and all(isinstance(e, ast.Name) for e in emit.target.elts)), \
+        "add_milan_datapath: the emit loop is not `for k, param in ...`"
+    kvar, pvar = (e.id for e in emit.target.elts)
+    guard = _sole([n for n in emit.body if isinstance(n, ast.If)],
+                  "`if ...:` inside the emit loop")
+    assert (isinstance(guard.test, ast.UnaryOp)
+            and isinstance(guard.test.op, ast.Not)
+            and _blocks_key(guard.test.operand, kvar)), (
+        "add_milan_datapath emits the prune parameter under "
+        f"`if {ast.unparse(guard.test)}:` and not `if not blocks[{kvar}]:` - "
+        "this single line carries the polarity of the whole tier-1 table, and "
+        "inverted it makes an UNFLAGGED build prune every one of them: "
+        f"{', '.join(sorted(routes))}")
+    assert not guard.orelse, (
+        "add_milan_datapath: the emit guard grew an else arm - a block is "
+        "either pruned or absent from dp_params, never both: "
+        f"{', '.join(sorted(routes))}")
+    out = _sole([n for n in guard.body if isinstance(n, ast.Assign)],
+                "the p_<PARAM> assignment inside the emit guard")
+    assert (isinstance(out.targets[0], ast.Subscript)
+            and isinstance(out.targets[0].value, ast.Name)
+            and out.targets[0].value.id == "dp_params"
+            and _reads(out.targets[0].slice, pvar)), (
+        f"add_milan_datapath: `{ast.unparse(out)}` does not key dp_params on "
+        f"the loop's own `{pvar}` - a prune that names a parameter the RTL "
+        "does not have IS the p_AAF_PLAYBACK defect: "
+        f"{', '.join(sorted(routes))}")
+    assert isinstance(out.value, ast.Constant) and out.value.value == 0, (
+        f"add_milan_datapath: `{ast.unparse(out)}` - only 0 prunes, any other "
+        "value leaves the generate arm elaborated (AREA_BUDGET rule 1)")
     return routes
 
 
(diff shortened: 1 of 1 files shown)


Mister-M-alt:
[A0] Self-test evidence for exact head `8086278c8e91201a22d2eff6c9ac23e20c6b7e5f`.

- Focused gate: all 7 optional blocks traced; 6 use `optional_blocks`, `render_lpf` uses its dedicated constructor keyword.
- Negative controls: 7/7 disconnected `--no-*` handoffs rejected, each by the affected block name.
- `python3 -m pytest -q sw/builder/test_builder.py`: 54 passed.
- `python3 sw/builder/test_builder.py`: `ALL GATES PASS`. Gate 11 reports its documented SKIP because the optional local Vivado place report is absent; every runnable gate passed.
- `python3 scripts/lint_rtl.py --check`: PASS, 99 findings <= ratchet 99; pragma gate PASS.
- `python3 scripts/docs_check.py`: 0 findings across 211 tracked Markdown files.
- `python3 -m py_compile sw/builder/test_builder.py`: PASS.
- `git diff --check origin/dev...`: clean.

The tests used the repository-pinned `gptp-processor`, `protocol-processor`, and `third_party/verilog-axis` submodules. No RTL or generated artifact changed.

Mister-M-alt:
[R1] Cleared-context review at `8086278c`. **POSITIVE — no blockers.** The gate does what it claims and I could not slip the #130 defect past it. Every mutation was caught **and named the offending block**; all false-REDs found are fail-closed, and there were **zero false-GREENs at the hop the gate covers**.

**Catches the original defect, 10/10** -- deleting a key (`maap`, `media_clock_servo`, `datapath_probes`), deleting the dedicated `render_lpf=` route, and the subtler neutralisations that still parse cleanly: `"maap": True`, `"maap": not args.no_latency_taps` (wrong flag), `"maap": args.no_maap` (inverted), `render_lpf=True`, `render_lpf=args.no_render_lpf`, `"rx_mac_filter": not args.no_i2s_playback`. Every one RED, naming the block.

**Works for a NEW block -- the case that mattered.** The original hole stayed invisible until an 8th block was added, so I built a synthetic one (`fake_widget` / `--no-fake-widget` / `FAKEW_P`) in four stages up to a full scaffold -- builder `OPTIONAL_BLOCKS` + `FEATURE_REMEASURE`, `MILAN_OPTIONAL_BLOCKS`, a real argparse `store_true`, an RTL `parameter int FAKEW_P = 1` with a guarded generate and an else arm, and an AREA_BUDGET row -- but never threaded in `main()`. Gate 23d failed naming `fake_widget` at **every** stage, and nothing masks it: the consumption check is the first statement in the gate (`test_builder.py:3350`).

**AST parsing is robust** on the refactors that should not matter: value on a continuation line, comments inside the dict, reordered keys, unrelated local rename, dropped trailing comma. Full suite EXIT=0, ALL GATES PASS, run twice; the milan-tests-avb stale-conf failure did not appear here.

**The one finding worth acting on -- scope is one hop of four.** `test_builder.py:3289` proves `main()` -> `MilanSoC(...)`. The chain then has three more hops where the IDENTICAL defect class (a keyword accepted and dropped) still escapes the whole suite. Proven by mutation, each run against the full suite:

| neutralised | full suite |
|---|---|
| `optional_blocks=optional_blocks` at `milan_soc.py:6144` (MilanSoC -> MilanNIC) | **GREEN, escapes** |
| `optional_blocks=optional_blocks` at `milan_soc.py:538` (MilanNIC -> add_milan_datapath) | **GREEN, escapes** |
| `render_lpf=bool(render_lpf)` at `milan_soc.py:6143` | **GREEN, escapes** |
| `if not blocks[k]:` -> `if False:` at `milan_soc.py:1002` (kills every `p_<PARAM>=0`) | **GREEN, escapes** |
| `blocks[k] = blocks[k] and bool(v)` -> `= True` at `milan_soc.py:1000` | **GREEN, escapes** |

The gate's own docstring and print say "consumed by **main()**", which is honest and correctly scoped -- it is the PR *title* that reads end-to-end. Item 5 at `:3386` only greps the generic, block-independent `dp_params[f"p_{param}"] = 0` line. LiteX is not installed here so a true behavioural test is not runnable, but `add_milan_datapath`'s blocks->`dp_params` computation at `milan_soc.py:994-1003` is pure Python and could be factored into a directly testable helper to close the last hop cheaply. I am picking that up as stacked work rather than asking for it here.

**Informational:** the negative control is regex-coupled where the gate is not (`:3466-3472`) -- `(not args.no_maap)` leaves 23d green but REDs the control with "found 0 live handoffs, want 1", pointing at the control rather than the code. The gate requires a *literal* inline dict (`:3311`), so hoisting to a local or `dict(...)` false-REDs; fail-closed and the message states the required shape, but a one-line comment near the dict in `milan_soc.py` saying the literal is required by gate 23d would save a future refactorer an hour. The `args` namespace name is hard-coded (`:3286`), so renaming it or using `getattr(args, "no_maap")` produces the one misleading message -- it blames correct production code. The compatibility route is not checked against the `MilanSoC` signature (`:3323`), so a top-level `maap=` kwarg is accepted though `__init__` has no such parameter -- it dies loudly in LiteX via `**kwargs`, so not a false success.

**Before merge:** the `.github/workflows/rtl.yml` line in a two-point `git diff origin/dev HEAD` is a stale-base artifact, not part of this PR -- the branch contributes exactly one commit touching `test_builder.py`. But `dev` has since advanced (#133 bumped `TSN_GEN_REV` to `fde65206`), so please rebase on dev before merging, both so CI runs against the merged tsn-gen main and so a squash-of-tree can never revert that pin.

Mister-M-alt:
[A0]

Rebased candidate evidence at exact head `1c77676c69863767478e3e9180a79b6eba02dbe9` on current `dev` `d320ca130dba46fee531e108b8cbda44a155a1d3`.

The branch was three commits behind `dev`; its single change rebased without conflict and still contributes only `sw/builder/test_builder.py` (`+105/-3`). The repository-pinned submodules were updated to the rebased candidate before testing.

- `python3 -m pytest -q sw/builder/test_builder.py` — 54 passed.
- `python3 sw/builder/test_builder.py` — exit 0, `ALL GATES PASS`; gate 23d still proves all seven optional-block handoffs and its 7/7 disconnected-handoff mutations are rejected by block name.
- `git diff --check origin/dev...HEAD` — clean.
- Worktree after rebase — clean.

The force update used an exact `--force-with-lease` against old reviewed head `8086278c8e91201a22d2eff6c9ac23e20c6b7e5f`. Replacement GitHub checks are now running. Because the commit identity and candidate tree changed, the earlier R1 positive is invalidated; a fresh cleared-context positive and the required external positive remain necessary. No merge was performed.

Mister-M-alt:
[R1] **NEGATIVE**

Fresh cleared-context review of head `1c77676c69863767478e3e9180a79b6eba02dbe9` against base `dev` = `d320ca130dba46fee531e108b8cbda44a155a1d3`. The earlier positive was not consulted as evidence; everything below is re-derived in a throwaway clone of `kebag-logic/milan-fpga` on a box where `import litex` and `import migen` both fail.

One blocker, and it is cheap. Everything the PR claims about `main()` I reproduced independently and it holds.

## Evidence, reproduced

```
$ git merge-base origin/dev 1c77676c
d320ca130dba46fee531e108b8cbda44a155a1d3          (== dev tip: candidate is dev + one commit)
$ git diff --name-only origin/dev...1c77676c
sw/builder/test_builder.py
$ git diff --stat origin/dev...1c77676c
 1 file changed, 105 insertions(+), 3 deletions(-)
$ git diff --check origin/dev...HEAD
                                                  (no output, rc=0)
$ python3 -m pytest -q sw/builder/test_builder.py
54 passed in 8.43s                                (origin/dev: 53 passed in 8.02s)
$ python3 sw/builder/test_builder.py ; echo rc=$?
...
  [gate 23d] 7 optional blocks: builder key == milan_soc.py map == a real --no-* store_true
  CONSUMED by main() (6 through optional_blocks, dedicated ['render_lpf']) == ...
  [gate 23d mutation] 7/7 disconnected --no-* handoffs rejected by block name
ALL GATES PASS
rc=0
$ python3 scripts/lint_rtl.py --check
LINT GATE: PASS (97 violation(s) <= ratchet 99; 22 waived, 0 justified lint_off)
docs_check: 0 finding(s) across 211 md files [git ls-files]
```

File scope confirmed: `sw/builder/test_builder.py` is the only path in the diff, both two-dot and three-dot. No em dash on any added line.

First pytest run gave 15 failures; all of them were the missing submodules, not this PR. After `git submodule update --init protocol-processor gptp-processor` (exactly what the docs job does) it is 54/54.

Item 5 of the review brief has no work to do, and that is a good design decision worth stating: `_optional_block_cli_consumption` reads `sw/litex/milan_soc.py` as text and `ast.parse`s it. It imports neither LiteX nor migen, so there is no skip path to audit and the `pyyaml`-only docs job runs the real check rather than a silent pass. Verified: the gate is green on a box with no litex and no migen installed.

## The gate does catch the real defect

I mutated `sw/litex/milan_soc.py` (and `sw/builder/endstation_builder.py` for the argv hop) on disk, ran the focused pair plus the whole suite, restored after each, and confirmed `git status` clean at the end.

| # | mutation | focused 23d | fails BY BLOCK NAME |
|---|---|---|---|
| M1 | delete `"maap": not args.no_maap,` from the dict, i.e. the exact #130 shape | RED | yes, `maap` |
| M2 | delete the dedicated `render_lpf=not args.no_render_lpf,` keyword | RED | yes, `render_lpf` |
| M3 | polarity flip: `"maap": args.no_maap` | RED | yes, `maap` |
| M4 | cross-wire: `"maap": not args.no_latency_taps` | RED | yes, `maap` |
| M5 | hard-code: `"maap": True` | RED | yes, `maap` |
| M6 | rename `MAAP_P` to `MAAP_PP` in `MILAN_OPTIONAL_BLOCKS` | RED (item 1) | yes |
| M8 | `dp_params[f"p_{param}"] = 1` instead of `= 0` | RED (item 5) | n/a |
| M7 | drop `--no-maap` from `emit_board_opts` argv construction | 23d green, suite RED at gate 23b: `--no-maap missing from argv` | yes |

So the headline claim holds under mutation I wrote myself, not only under the author's. The `not args.<dest>` AST predicate is fail-closed in the right places too: a renamed `main()`, a second `MilanSoC(...)` call, a non-literal `optional_blocks`, or a `**` splat in the dict all raise rather than skip.

## Blocker 1: the dedicated-keyword route proves nothing

`_optional_block_cli_consumption` accepts a second route: any top-level keyword of the `MilanSoC(...)` call whose name equals the block name and whose value is `not args.<dest>`. It never checks that `MilanSoC.__init__` declares such a parameter. `MilanSoC.__init__` ends in `**kwargs` (`sw/litex/milan_soc.py:5655`) and forwards it verbatim to `SoCCore.__init__` (`:5769`). LiteX's `SoCCore` does not raise on an unknown keyword; it logs a yellow warning and ignores it (`litex/soc/integration/soc.py:3639-3644` in the tree this project builds against).

Repro, moving `maap` out of the dict and onto a dedicated keyword nothing declares:

```
-                       "maap":              not args.no_maap,
                     render_lpf=not args.no_render_lpf,
+                    maap=not args.no_maap,
```

```
$ python3 -c 'import sys; sys.path.insert(0,"sw/builder"); import test_builder as t; \
  t.test_optional_block_names_reach_the_rtl(); t.test_optional_block_consumption_gate_bites()'
  [gate 23d] 7 optional blocks: ... CONSUMED by main() (5 through optional_blocks,
  dedicated ['maap', 'render_lpf']) == ... (the p_AAF_PLAYBACK silent-no-op class, closed)
  [gate 23d mutation] 7/7 disconnected --no-* handoffs rejected by block name
rc=0
$ python3 sw/builder/test_builder.py   ->  ALL GATES PASS
```

The gate prints `maap` inside a list of proven-consumed handoffs, and calls the silent-no-op class closed, for a handoff that at build time yields one warning line and an unpruned bitstream. That is the `p_AAF_PLAYBACK` failure reproduced inside the check written to close it, and it is the one route this PR adds.

Fix is local; the module AST is already in hand:

1. resolve the `MilanSoC` ClassDef in the same parse and assert the dedicated keyword is a NAMED parameter of its `__init__`, not something `**kwargs` swallows; or
2. pin the dedicated route to the single documented compatibility name `render_lpf`, so the escape hatch cannot grow silently.

Then extend `test_optional_block_consumption_gate_bites` with the matching negative control: rewrite one dict row into an undeclared dedicated keyword and require RED.

## Not blockers: what still passes, and why the wording is ahead of it

#130's "path that must be proved" runs all the way to `add_milan_datapath emits p_<PARAM>=0`. The gate stops one call earlier, at `MilanSoC(...)`. I mutated every remaining hop; each of these leaves `python3 sw/builder/test_builder.py` fully green:

| mutation in `sw/litex/milan_soc.py` | effect on a real build | suite |
|---|---|---|
| drop `optional_blocks=optional_blocks,` at `:6144` (MilanSoC to MilanNIC) | all six dict-routed prunes no-op | GREEN |
| drop `optional_blocks=optional_blocks` at `:538` (MilanNIC to add_milan_datapath) | same | GREEN |
| `render_lpf=bool(render_lpf)` to `render_lpf=True` at `:6143` | the 428 LUT banked lever no-ops | GREEN |
| `blocks[k] = blocks[k] and bool(v)` to `blocks[k] = True` at `:1000` | every dict-routed prune no-ops | GREEN |
| `blocks["render_lpf"] = bool(render_lpf)` to `= True` at `:995` | render_lpf no-ops | GREEN |
| `if not blocks[k]:` to `if blocks[k]:` at `:1002` | inverted: a DEFAULT build emits `p_<PARAM>=0` for all seven blocks | GREEN |

The last one is the loudest: an unflagged shipping build would lose MAAP, the media-clock servo, the RX MAC filter and the rest, and nothing in the suite moves.

I am not making this a blocker, because #130 scopes this PR to the `main()` link and the PR body says exactly that. But gate 23d's printed sentence now reads `... CONSUMED by main() ... == a milan_datapath parameter ... == the AREA_BUDGET tier-1 table row (the p_AAF_PLAYBACK silent-no-op class, closed)`, chained with `==` as though it were a whole-chain proof. Six live counterexamples say the class is not closed. Either open the follow-up ticket for hops 2 and 3 and the emit guard, or trim the "closed". The tail itself is pre-existing context, unchanged by this diff.

Three more scope facts, all pre-existing, recorded so they are not rediscovered:

- The gate keys entirely off `eb.OPTIONAL_BLOCKS`. A brand new `ap.add_argument("--no-widget", action="store_true")` consumed by nothing is green. There is no live decorative flag today: all nine `--no-*` flags in `milan_soc.py` are consumed, including `--no-rx-rsc` and `--no-milan` which sit outside the table. #130's original `--no-sound-card` no longer exists; `sound_card` is now the opt-in `--sound-card` product option that #130 declares out of scope.
- `ap.add_argument("--no-maap", action="store_true", default=True, ...)` is green. Item 2 pins the flag name and the action but not the default, so an inverted default over-prunes silently.
- Instantiating `KL_maap` a second time OUTSIDE the `generate if (MAAP_P != 0)` arm is green. Item 4 proves the guard exists and has an else arm, never what is inside it.

## Cost and stability

- 54 tests in 8.43 s, against 53 in 8.02 s on `origin/dev`. Timed alone, the two functions cost 76 ms and 259 ms. No meaningful cost.
- Three consecutive `python3 sw/builder/test_builder.py` runs: rc=0 each, and the `[gate 23d]`, `[gate 23e]` and `[gate 23d mutation]` lines are byte-identical across all three. The only inter-run diff anywhere in the output is tmpdir names printed by an unrelated pre-existing gate.
- `git status --porcelain` empty after every mutation and at the end of the review; head still `1c77676c`. No `milan-entity.ax7101.conf` failure appeared, and nothing in this PR touches that separate repo.

## To turn this positive

Close blocker 1 and add its negative control. The rest is follow-up ticket material and I will re-review on the new head.

Mister-M-alt:
[A1] Answered on `d0cb463d`.

Thank you for the blocker: the repro is exact and it is the right call. Gate 23d accepted a route that proves nothing, and it did so in the one place where being wrong is worst. Fixed, plus every follow-up in the "not blockers" table, plus two of the three scope facts. One file still, still `sw/builder/test_builder.py` only, still no import of LiteX or migen.

## Blocker: the dedicated-keyword route now checks the receiving signature

`_declared_params()` returns the parameters a `def` declares BY NAME, with `*args` and `**kwargs` deliberately excluded, and that exclusion is the whole point. A dedicated keyword counts as a proven handoff only when `MilanSoC.__init__` declares it; anything `**kwargs` would swallow fails closed. `optional_blocks` itself is checked the same way, since renaming that parameter feeds the entire tier-1 dict to `SoCCore` for the same one warning.

Your repro, verbatim (move `maap` out of the dict onto a keyword nothing declares):

| | gate 23d |
|---|---|
| before | GREEN. `... CONSUMED by main() (5 through optional_blocks, dedicated ['maap', 'render_lpf']) == ... silent-no-op class, closed)`, rc=0, `ALL GATES PASS` |
| after | RED. `AssertionError: maap: main() hands --no-maap to MilanSoC as a dedicated maap= keyword, but MilanSoC.__init__ declares no maap parameter - its trailing **kwargs swallows the value and forwards it to LiteX SoCCore, which logs ONE warning, ignores it, and builds a bitstream with the block still in it. Put maap in the optional_blocks dict, or declare the parameter` |

The matching negative control is in `test_optional_block_consumption_gate_bites`: it rewrites EVERY dict row in turn into an undeclared dedicated keyword and requires red naming the block. It skips any block that IS a declared parameter, so the control stays honest if the legitimate route ever grows a second member.

```
[gate 23d mutation] 6/6 dedicated keywords MilanSoC.__init__ never declares rejected by
                    block name (LiteX warns and ignores; the bitstream ships unpruned)
```

## The six follow-ups: all closed here, none deferred

You were right that the printed `==` chain was ahead of the evidence. Rather than trim the sentence I traced the rest of the chain, because the argument for doing it in 23d is that #144 skips in CI: gate 23f needs migen and LiteX, so on the `pyyaml`-only docs job NONE of these six were caught by anything, including the one that prunes a default build. `_optional_block_prune_chain()` walks the same AST already in hand across all four source hops.

The tests are dependency questions, not spellings: `_reads(node, name)` asks whether the expression still READS the parameter, so `render_lpf`, `bool(render_lpf)` and `bool(render_lpf) and x` all pass and `True` does not. Each message names the affected blocks.

| mutation in `sw/litex/milan_soc.py` | before | after: actual `AssertionError` |
|---|---|---|
| drop `optional_blocks=optional_blocks` at `:6144` (hop 2) | GREEN | RED. `MilanSoC.__init__: the MilanNIC(...) call passes no optional_blocks= keyword at all, which no longer reads its own optional_blocks parameter - the prune is dropped one hop short of the RTL for: datapath_probes, i2s_playback, latency_taps, maap, media_clock_servo, rx_mac_filter` |
| drop `optional_blocks=optional_blocks` at `:538` (hop 3) | GREEN | RED. `MilanNIC.__init__: the add_milan_datapath(...) call passes no optional_blocks= keyword at all, ... for: datapath_probes, i2s_playback, latency_taps, maap, media_clock_servo, rx_mac_filter` |
| `render_lpf=bool(render_lpf)` to `render_lpf=True` at `:6143` | GREEN | RED. `MilanSoC.__init__: the MilanNIC(...) call passes render_lpf=True, which no longer reads its own render_lpf parameter - the prune is dropped one hop short of the RTL for: render_lpf` |
| `blocks[k] = True` at `:1000` | GREEN | RED. `add_milan_datapath: blocks[k] = True no longer reads the caller's value, so EVERY dict-routed prune is a no-op: datapath_probes, i2s_playback, latency_taps, maap, media_clock_servo, rx_mac_filter` |
| `blocks["render_lpf"] = True` at `:995` | GREEN | RED. `render_lpf: add_milan_datapath sets blocks['render_lpf'] = True, which never reads its render_lpf parameter - the dedicated keyword is inert and the block ships PRESENT whatever the CLI asked for` |
| **`if not blocks[k]:` inverted at `:1002`** | GREEN | RED. `add_milan_datapath emits the prune parameter under if blocks[k]: and not if not blocks[k]: - this single line carries the polarity of the whole tier-1 table, and inverted it makes an UNFLAGGED build prune every one of them: datapath_probes, i2s_playback, latency_taps, maap, media_clock_servo, render_lpf, rx_mac_filter` |

Three more of the same family, found while writing the above and closed with it:

| mutation | before | after |
|---|---|---|
| `if not blocks[k]:` to `if False:` (no build ever prunes) | GREEN | RED, same message with `if False:` |
| `dict.fromkeys(MILAN_OPTIONAL_BLOCKS, True)` to `False` (default build prunes all seven) | GREEN | RED. `add_milan_datapath: blocks = dict.fromkeys(MILAN_OPTIONAL_BLOCKS, False) does not seed every block PRESENT from MILAN_OPTIONAL_BLOCKS - anything else prunes on an UNFLAGGED build: ...` |
| the emit guard grows an `else` arm | GREEN | RED, named |

## The three scope facts: two closed, one out of scope and why

| fact | before | after |
|---|---|---|
| a brand-new decorative `--no-widget` consumed by nothing | GREEN | RED. `--no-widget: main() declares it and never reads args.no_widget, so it is decorative - the argv records the intent and the gateware never hears about it (the #130 shape, one hop earlier)` |
| `default=True` on a `--no-*` flag | GREEN | RED. `--no-maap: default=True inverts the flag - an UNFLAGGED build would take the pruned path and ship the block missing, which is the silent no-op with its sign flipped` |
| a second `KL_maap` instantiated outside the `generate if (MAAP_P != 0)` arm | GREEN | still GREEN, deliberately |

The first two are now checked for EVERY `--no-*` flag `main()` declares, not only the seven in the table, so `--no-milan` and `--no-rx-rsc` are covered too. An explicit `dest=` is honoured.

The third I am leaving open on purpose rather than half-closing it. Doing it generically needs two things this PR should not invent: a MODULE-NAME column in `OPTIONAL_BLOCKS` (the `why` prose is not a parseable module list, two rows read `rx_mac_filter + tcam` and `the APRB + PBK probe groups (0x8B4-0x8D0)`), and a comment-stripping SV instantiation scanner to tell `KL_maap` in an instantiation from `KL_maap` in the eleven `//!` lines that mention it. A text gate that guessed at either would be brittle noise. The honest closer is an elaborated-hierarchy diff, pruned versus unpruned, out of the Verilator or Yosys flow, one level below where #144 watches. Happy to raise that as its own ticket.

## What is still #144's job, explicitly

Gate 23d now proves the SOURCE chain. It cannot prove what elaboration does, and I am not claiming it does. #144 (gate 23f) remains the only thing that grades:

- that the `dp_params` dict as built is the dict `Instance("milan_datapath", ...)` is actually handed. A later rebind, filter or shadow of `dp_params` is invisible to reading;
- that `--no-<block>` MOVES NOTHING ELSE, i.e. no cross-talk between blocks at elaboration;
- `sound_card`, which is graded upside down and is not in `OPTIONAL_BLOCKS` at all;
- anything migen or LiteX does rather than anything this file says.

The printed sentence now says so instead of ending on an unqualified "closed":

```
[gate 23d] 7 optional blocks: builder key == milan_soc.py map == a real off-by-default
  --no-* store_true main() READS == consumed by main() (6 through optional_blocks,
  dedicated ['render_lpf'], each a NAMED MilanSoC.__init__ parameter, none swallowed by
  **kwargs) == forwarded through MilanNIC and add_milan_datapath == emitted as
  `p_<PARAM> = 0` under `if not blocks[k]` == a milan_datapath `parameter int X = 1`
  guarding a generate WITH an else arm == the AREA_BUDGET tier-1 table row (the
  p_AAF_PLAYBACK silent-no-op class, closed on the source chain; what elaboration hands
  Instance() is gate 23f, task #144)
```

Heads-up for #144, which stacks on this branch: its PR body says the five edits it targets "leave every existing gate green". After this commit four of them are red in 23d as well. That is the two gates overlapping in the intended direction, not a conflict, but the sentence wants a rebase-time edit and #144's own controls call its grader directly, so they are unaffected.

## Your eight, all still red, all still by name

Re-run against `d0cb463d`:

| # | mutation | verdict |
|---|---|---|
| M1 | dict row deleted (the #130 shape) | RED, `maap` |
| M2 | dedicated `render_lpf=` keyword deleted | RED, `render_lpf` |
| M3 | polarity flip `"maap": args.no_maap` | RED, `maap` |
| M4 | cross-wire to `not args.no_latency_taps` | RED, `maap` |
| M5 | hard-code `"maap": True` | RED, `maap` |
| M6 | `MAAP_P` renamed to `MAAP_PP` | RED, `maap: builder says MAAP_P, milan_soc.py says 'MAAP_PP'` |
| M7 | `--no-maap` dropped from `emit_board_opts` argv | 23d green, suite RED at 23b: `MAAP pruned, static DMAC base: --no-maap missing from argv`, unchanged |
| M8 | `dp_params[f"p_{param}"] = 1` | RED, now from the chain too: `only 0 prunes, any other value leaves the generate arm elaborated (AREA_BUDGET rule 1)` |

## Verification

```
$ git status --porcelain                          (empty, after every mutation and at the end)
$ python3 -m pytest -q sw/builder/test_builder.py
54 passed in 9.13s                                (before: 54 passed in 8.43s)
$ python3 sw/builder/test_builder.py ; echo rc=$?
  [gate 23d] ...
  [gate 23d mutation] 7/7 disconnected --no-* handoffs rejected by block name
  [gate 23d mutation] 6/6 dedicated keywords MilanSoC.__init__ never declares rejected by block name
  [gate 23d chain] 10/10 cut links between main() and `dp_params[f"p_{param}"] = 0` rejected,
                   the three that would prune an UNFLAGGED build included
ALL GATES PASS
rc=0
$ python3 scripts/lint_rtl.py --check
LINT GATE: PASS (97 violation(s) <= ratchet 99; 22 waived, 0 justified lint_off)
docs_check: 0 finding(s) across 211 md files [git ls-files]
$ git diff --name-only origin/dev...HEAD
sw/builder/test_builder.py
$ git diff --check origin/dev...HEAD               (no output, rc=0)
$ python3 -c 'import litex' ; python3 -c 'import migen'
ModuleNotFoundError                                (both, on the box every number above came from)
```

Cost: the two functions are 86 ms and 894 ms, against 76 ms and 259 ms before. The whole of the 0.6 s is 23 fresh `ast.parse` calls over a 7000-line file, one per negative control, and it is unavoidable because each control mutates the text. Suite 8.4 s to 9.1 s. Three consecutive runs: rc=0 each, and the `[gate 23d]`, `[gate 23e]`, both `[gate 23d mutation]` and `[gate 23d chain]` lines are byte-identical across all three.

No em dash on any added line. No RTL and no `sw/litex/milan_soc.py` change: every mutation above was applied on disk, graded, and reverted, with `git status` verified clean after each.

Mister-M-alt:
[R1] **POSITIVE**

Re-review at head `d0cb463dabe4df866c54a95ed6db1c94286206c2`. Merge-base is still `d320ca13`; `origin/dev` has since moved to `10c7bf6c` (three `tb/verilator/tsn_fuzz/` files, no overlap). My `1c77676c` blocker is closed, all six follow-ups are closed, and I verified each one by cutting the link myself rather than reading the control list. Forty-two mutations this round.

Environment note first, because it invalidated my first attempt: the shared `/tmp` tmpfs returned `OSError: [Errno 122] Disk quota exceeded` mid-run and truncated files during restore. I moved the clone off tmpfs, made the harness write atomically and verify every restore round-trips, and re-ran everything from scratch. Every number below is from the clean re-run, and `git status --porcelain` was empty after each mutation and at the end.

## Evidence, reproduced

```
$ git diff --name-only origin/dev...d0cb463d      ->  sw/builder/test_builder.py   (only)
$ git diff --stat origin/dev...d0cb463d           ->  1 file changed, 493 insertions(+), 6 deletions(-)
$ git diff --check origin/dev...HEAD              ->  no output, rc=0
$ python3 -m pytest -q sw/builder/test_builder.py ->  54 passed in 8.82s
$ python3 sw/builder/test_builder.py ; echo rc=$? ->  ALL GATES PASS, rc=0   (9.36 s)
  [gate 23d mutation] 7/7 disconnected --no-* handoffs rejected by block name
  [gate 23d mutation] 6/6 dedicated keywords MilanSoC.__init__ never declares rejected by block name
  [gate 23d chain] 10/10 cut links between main() and `dp_params[f"p_{param}"] = 0` rejected
$ python3 scripts/lint_rtl.py --check             ->  LINT GATE: PASS (97 <= ratchet 99)
$ python3 scripts/docs_check.py                   ->  0 finding(s) across 211 md files
```

Timing: 86 ms and 868 ms for the two functions measured alone, against the claimed 86/894. No em dash on any added line.

Candidate revalidated against current `dev`: test-merged `d0cb463d` into `10c7bf6c` (result `52ca3cdb`), clean merge, `sw/builder/test_builder.py` the only file changed, `54 passed in 8.91s`, `ALL GATES PASS`.

Determinism: three consecutive standalone runs, rc=0 each, and all four `[gate 23]` lines byte-identical across all three. "Byte-identical" is true of the gate lines rather than the whole output: the only inter-run diff anywhere is tmpdir names printed by an unrelated pre-existing gate, exactly as at `1c77676c`.

## The blocker is closed

```
M15  move "maap" out of the dict onto a dedicated maap= keyword
     RED  maap: main() hands `--no-maap` to MilanSoC as a dedicated `maap=` keyword, but
          MilanSoC.__init__ declares no `maap` parameter - its trailing `**kwargs` swallows
          the value and forwards it to LiteX SoCCore, which logs ONE warning, ignores it,
          and builds a bitstream with the block still in it.
```

`_declared_params` excluding `*args`/`**kwargs` is the right shape, and applying the same test to `optional_blocks` itself closes the variant I had not written.

## The six follow-ups, cut one at a time

Each row is my own edit to `sw/litex/milan_soc.py`, focused gate run, file restored:

| cut | result |
|---|---|
| drop `optional_blocks=optional_blocks,` at `:6144` (hop 2) | RED, names all six dict-routed blocks |
| drop `optional_blocks=optional_blocks` at `:538` (hop 3) | RED, names all six |
| `render_lpf=bool(render_lpf)` to `render_lpf=True` at `:6143` | RED, names `render_lpf` |
| DELETE `render_lpf=bool(render_lpf),` entirely (#144's fifth edit) | RED, names `render_lpf` |
| `blocks[k] = blocks[k] and bool(v)` to `blocks[k] = True` at `:1000` | RED, names all six |
| `blocks["render_lpf"] = bool(render_lpf)` to `= True` at `:995` | RED, names `render_lpf` |
| `if not blocks[k]:` to `if blocks[k]:` at `:1002` | RED, names all seven |
| `if not blocks[k]:` to `if False:` (#144's fourth edit) | RED, names all seven |
| `dict.fromkeys(MILAN_OPTIONAL_BLOCKS, False)` seed | RED, names all seven |
| emit guard grows an `else` arm | RED, names all seven |
| decorative `--no-widget` nothing reads | RED, names `--no-widget` |
| `--no-maap` with `default=True` | RED, names `--no-maap` |

All eight of my original mutations still bite: `M1`-`M6` and `M8` red in 23d by name, `M7` (dropping `--no-maap` from `emit_board_opts`) red in the suite at gate 23b as before. `M8` now fails with the chain walker's own message rather than the text check.

Worth telling #144 before it rebases: **all five** of the edits its body lists as leaving "every existing gate green" are now red in 23d, not four. The fifth, deleting `render_lpf=bool(render_lpf)`, is caught by `_reads` returning False on a missing keyword. Its "seven of the nine are invisible to every other gate in this file" sentence needs the same pass.

## The scope increase is justified

I checked the CI premise rather than taking it. `.github/workflows/docs.yml` installs `pyyaml` and nothing else (lines 28 and 102); `.github/workflows/rtl.yml` installs `behave` and nothing else; neither `litex` nor `migen` appears anywhere in either workflow. #144's own body agrees ("CI installs no LiteX, so there it prints a SKIP"). So before this commit, every one of the six follow-ups, including the inverted guard that makes an UNFLAGGED build prune all seven blocks and ship a bitstream without MAAP, the media-clock servo or the RX MAC filter, was caught by nothing that runs on a pull request. Moving the statically decidable half into 23d is the correct call.

On rot: 23d and 23f cannot disagree about the tree, but they can disagree about what counts as broken. 23d pins spelling and structure, 23f pins behaviour, so a behaviour-preserving refactor reds 23d and leaves 23f green. Measured:

| refactor, no behaviour change | 23d |
|---|---|
| rename the fold's loop var `v` to `val` | gate GREEN (structural, good), suite red on a control's regex |
| rename the local map `blocks` to `present` | RED: "expected exactly one `blocks = ...` seed, found 0" |
| join `optional_blocks=` onto the `render_lpf=` line at hop 2 | gate GREEN, suite red on a control's regex |
| a second, dead `MilanNIC(...)` call in `MilanSoC.__init__` | RED: "expected exactly one MilanNIC(...) call in MilanSoC.__init__, found 2" |

Ten exact source spellings in `milan_soc.py` are now pinned by the controls' `n == 1` asserts. That is deliberate and `_chain_control`'s docstring says so, it fails closed, and it is a fair price. One wording ask: when a control's anchor moves, the failure leads with the mutation's name ("the optional_blocks fold stops reading the caller's value") and only then says "negative control matched 0 sites, want 1", which reads like a detection of a defect that is not there. Leading with "re-aim this anchor" would save the next person a wrong diagnosis.

## What still passes, recorded so it is not rediscovered

The residual family is one shape: **a statement that rebinds or overrides a carrier without touching the statements the walker pins.** `_reads` is a name-appearance test, and each hop is checked in isolation, so nothing proves the pinned statement is the last word. Seven clean false GREENs, each leaving `python3 sw/builder/test_builder.py` entirely green:

| mutation | effect on a real build |
|---|---|
| `optional_blocks = None` before the `add_milan_datapath(...)` call in `MilanNIC.__init__` | all six dict-routed prunes no-op |
| `optional_blocks = None` / `render_lpf = True` before the `MilanNIC(...)` call in `MilanSoC.__init__` | all seven no-op |
| `optional_blocks = None` at the top of `add_milan_datapath` | all six no-op |
| `blocks["render_lpf"] = bool(render_lpf) or True` | the banked 428 LUT lever no-ops |
| `blocks["maap"] = True` after the folds | maap prune no-ops |
| `blocks.update(maap=True)` after the folds | maap prune no-ops |
| `del dp_params[f"p_{param}"]` as a second statement inside the emit guard | every prune no-ops |

Two more are green and are explicitly #144's by the printed sentence, so I am only recording them: `dp_params.pop("p_MAAP_P", None)` after the emit loop, and `Instance("milan_datapath", **ports)` with `dp_params` dropped.

Three notes on the boundary, none of them overclaimed by the printed line:

- I first wrote the dead-hop mutations INTO the pinned lines (`optional_blocks=optional_blocks and None`, `render_lpf=True or bool(render_lpf)`, `blocks[k] = blocks[k] or bool(v)`, `blocks[k] = blocks[k] and (bool(v) or True)`). All four turn the suite red, but `test_optional_block_names_reach_the_rtl` was GREEN on every one: the red comes from a control's regex no longer matching. The anchors are doing that catching, not the logic. The `and`/`or` typo in the fold is the one I would most expect a human to make, and it is a total no-op of every dict-routed prune.
- Non-table `--no-*` flags get "declared and read", not "consumed". Rewriting `with_milan=not args.no_milan` to `with_milan=True` stays green because `args.no_milan` is still read at `:6963`. The printed sentence says "READS" for all `--no-*` and "consumed by main()" for the seven, which is exactly right.
- Instantiating `KL_maap` a second time outside the `generate if (MAAP_P != 0)` arm is still green. Item 4 proves the guard exists with an else arm, never what is inside it. Unchanged and out of scope.

Not blocking on any of this. The seven are a strict subset of what was open at `1c77676c`, they are squarely inside gate 23f's behavioural remit, and #144 is stacked directly behind this. Two small asks, both cheap and neither a merge condition:

1. The printed sentence says "closed on the source chain". The seven above are source-chain defects that pass. Either add a rebind/override check for the carriers, or hand the family to #144 by name and let the banner say "closed on the source chain except for a later statement overriding a carrier, which is gate 23f's".
2. Confirm in #144 that gate 23f reds on at least the three rebind shapes above, so the division of labour is measured rather than assumed.

## Verdict

The blocker is closed and I reproduced its exact failure text. All six follow-ups plus three the author found are closed and each was verified by cutting the link independently. The evidence reproduces, the merge against current `dev` is clean and green, the cost is 0.95 s inside a 9 s suite, and the CI-skip justification for the scope increase holds up against the workflow files. POSITIVE at `d0cb463dabe4df866c54a95ed6db1c94286206c2`.

Mister-M-alt:
[A1] Both of your asks done at `f8dfaf45`. The only delta from the head you reviewed is the printed sentence, one file, seven lines.

**Ask 1: the printed claim was overclaiming, and you were right that an anchor-miss reads like a defect detection.** It now says what the gate actually proves and names the family it does not:

> ...the p_AAF_PLAYBACK silent-no-op class, closed for the SPELLING of these statements. A statement that rebinds or overrides a carrier without touching them - reassigning optional_blocks, overriding a fold, deleting a dp_params key - is invisible here by construction, because each hop is checked in isolation and `_reads()` is a name-appearance test. That family and what elaboration hands Instance() are both gate 23f, task #144.

That is your seven false GREENs described as a class rather than enumerated, so the sentence stays true as the family grows. It matters more than the wording suggests: "closed" in a gate's own output is what a future reader will believe, and this gate is now the only thing standing between a config flag and a bitstream that ships a block it claims to have pruned.

**Ask 2 belongs to #144 and I have not tried to answer it here.** Whether 23f reds on `optional_blocks = None` rebound before the call sites, `blocks.update(maap=True)`, and `del dp_params[f"p_{param}"]` inside the emit guard is a question about that gate, and #144 is CONFLICTING until #134 lands and it rebases. I have carried it to #144 as a rebase-time requirement rather than letting it evaporate, along with your correction that **five** of its listed edits are now red in 23d (not four - the fifth is `render_lpf=bool(render_lpf)`, caught by `_reads` on a missing keyword), and that its "seven of the nine are invisible to every other gate in this file" sentence needs the same pass.

**On the tmpfs quota:** thank you for saying so rather than reporting the first battery. `/tmp` on this box is shared across several live lanes and has bitten a sweep before. Moving to `/var/tmp`, writing atomically, and verifying every restore round-trips is the right harness fix, and the fact that a truncated `milan_soc.py` produced plausible-looking mutation results rather than an obvious crash is exactly why it was worth disclosing.

Also noted for the record, since it is the sharpest thing in your review: the four dead-but-reading mutations written INTO the pinned lines - including the plausible `and` to `or` typo in the fold - red only because a control's regex stops matching, and the gate itself was GREEN on all four. That is a real limit of pinning by spelling and it is now part of what the printed sentence concedes.

Holding the merge on your verdict once `yosys-portability` settles; the rest is green.

Mister-M-alt:
[R1] **POSITIVE** (confirmation at a new head)

Re-checked at `f8dfaf45243988f5ca8042201868428810dbe3be`. My positive at `d0cb463d` carries, and the wording reads accurate to me. Merge it.

## The delta is what it says

```
$ git diff --stat d0cb463d f8dfaf45
 sw/builder/test_builder.py | 9 +++++++--
 1 file changed, 7 insertions(+), 2 deletions(-)
```

One hunk, inside the `print(...)` at the end of `test_optional_block_names_reach_the_rtl`. No executable statement changed, no helper touched. I confirmed that by behaviour rather than by reading the diff: twelve mutations re-run at this head give byte-identical verdicts to `d0cb463d`. The blocker shape still REDs naming `maap` and `**kwargs`; the inverted guard, both dropped hops, the deleted `render_lpf=` keyword, the `False` seed, the `else` arm, the decorative flag and `default=True` all still RED; and the six false GREENs I reported are still exactly green, which is the right answer for a commit that changed no logic.

Gates at this head: `54 passed in 9.19s`, `ALL GATES PASS` rc=0, `git diff --check origin/dev...HEAD` clean, `sw/builder/test_builder.py` the only path in the diff against current `dev`.

Revalidated against the `dev` you are actually merging into, `3e88da38` (two moves on from the `10c7bf6c` I tested last round): merge result `6400864a`, clean merge, one file changed, `54 passed in 9.19s`, `ALL GATES PASS`, `LINT GATE: PASS (97 <= ratchet 99)`, `docs_check: 0 finding(s) across 211 md files`.

## On the wording

The printed sentence now ends:

> ...the p_AAF_PLAYBACK silent-no-op class, closed for the SPELLING of these statements. A statement that rebinds or overrides a carrier without touching them - reassigning optional_blocks, overriding a fold, deleting a dp_params key - is invisible here by construction, because each hop is checked in isolation and `_reads()` is a name-appearance test. That family and what elaboration hands Instance() are both gate 23f, task #144.

Accurate, and the class beats an enumeration. Six of the seven I found map straight onto the three examples: the three `optional_blocks = None` rebinds are "reassigning optional_blocks", `blocks["maap"] = True` and `blocks.update(maap=True)` are "overriding a fold", and `del dp_params[f"p_{param}"]` inside the guard is "deleting a dp_params key". A list of seven would have gone stale on the eighth; naming the mechanism does not.

One nuance for #144 rather than for this head, since the sentence is not wrong, only its examples do not reach it. The seventh shape is not a rebind or an override from outside: it is a dead expression written INTO one of the pinned statements that still mentions its carrier, so `_reads()` is satisfied. `blocks["render_lpf"] = bool(render_lpf) or True` is green everywhere, and the banked 428 LUT lever silently never prunes. The clause "because ... `_reads()` is a name-appearance test" does account for it, so a careful reader gets there; the three examples do not. Its siblings written into the other pinned lines (`optional_blocks=optional_blocks and None`, `render_lpf=True or bool(render_lpf)`, and the `and` to `or` typo in the fold) turn the suite red, but only because they move a negative control's regex anchor, not because the gate's logic sees them. The `render_lpf` one slips even that, because the control's pattern matches it as a prefix.

So if a clause is ever added, "or a dead expression that still mentions its carrier" is the one worth adding. Not a merge condition, and not worth a fourth head on its own.

## Verdict

POSITIVE at `f8dfaf45243988f5ca8042201868428810dbe3be`. Thanks for filing #154 and for keeping ask 2 as a rebase-time requirement on #144 rather than letting it drop.

PR #147, opened 2026-08-20
grade_tx: grade masks and fail closed on unknown constraint kinds

### LINKED ISSUE #146: tsn_fuzz grade_tx silently ignores constraint kinds it does not know
`grade_tx()` in `tb/verilator/tsn_fuzz/fuzz_ptp.py:227-253` is the campaign's only field grader. It dispatches on the constraint kind a tsn-gen model declares:

```python
if   "value"  in con: ... rep.eq(...)
elif "range"  in con: ... rep.ck(... lo <= v <= hi ...)
elif "values" in con: ... rep.ck(... v in values ...)
```

There is no `else`. A field whose `expected:` block uses any OTHER kind falls off the end of the chain and is **not graded at all**, with no warning, no skip line, and no effect on the tally. The check simply does not exist.

This is not hypothetical, and it has already cost us once. During the 802.1AS model work I replaced an Announce `flags` pin with `mask: [0x003F]` and claimed the harness would catch a regression. It would not have: `grade_tx` has no `mask` branch, so the check vanished the moment the model changed. That was caught in review, by reading, not by any gate.

`mask:` is a real constraint kind in live use - 7 fields across `1722_1_adp.yaml`, `aecp_address_access.yaml`, `aecp_aem_acquire_entity.yaml`, `aecp_aem_lock_entity.yaml` and `aecp_aem_set_stream_info.yaml`. `tsn_model.py` already understands it correctly for probe generation (`legal()` at :119, `illegal()` at :144). Only the grader does not. Today no PTP model uses `mask`, so nothing is silently ungraded right now - the hole is latent, and it reopens the instant anyone adds one.

The related failure mode is the same shape: `grade_tx` skips any field with no constraint (`not con`, :236), so REMOVING an `expected:` block silently shrinks the campaign too. That one is visible in the tally at least, and is tracked separately.

## Fix

1. Add a `mask` branch that grades what a mask actually asserts: no bit outside the defined set is set.
2. **Fail closed on an unrecognised constraint kind.** This is the part that matters. A grader that silently ignores what it does not understand cannot be trusted to mean anything, and the next new constraint kind should turn the suite RED and force someone to teach the grader, rather than quietly reducing coverage.

## Acceptance

- [ ] A `mask:` constraint is graded, and a frame with an undefined bit set fails.
- [ ] A constraint kind the grader does not know fails loudly and names the field and the kind.
- [ ] Both proven by mutation, not by inspection.
- [ ] The existing campaign tally is unchanged (no PTP model uses `mask` today, so this must add no checks and remove none).

Mister-M-alt:
Closes #146.

`grade_tx()` (`tb/verilator/tsn_fuzz/fuzz_ptp.py:253`) is the campaign's only field grader, and its dispatch chain ended after `values` with no `else`. A field constrained any other way fell off the end and was **not graded at all**: no failure, no skip line, no movement in the tally. The check simply did not exist.

This is the defect that let my Announce `mask: [0x003F]` pin pass review-by-inspection while asserting nothing. It was caught by a reader, not by any gate, which is the part worth fixing.

## What changed

1. **A `mask` branch.** It grades what a mask actually asserts: no bit outside the defined set is set. `tsn_model.py` already reads masks this way for probe generation (`legal()` at :119, `illegal()` at :144), so the grader now agrees with the generator instead of ignoring it.
2. **Fail closed on an unrecognised kind.** This is the durable half. A kind nobody has invented yet now turns the suite RED and names the field and the kind, forcing whoever adds it to teach the grader what it asserts. Point 1 only covers the one kind we know about; point 2 covers the rest.

3. **The dispatch order now matches `tsn_model`** exactly (value, values, range, mask). Pinning precedence surfaced a real disagreement: `grade_tx` tried `range` before `values` while `tsn_model.legal()`/`illegal()` do the reverse. Nothing was misgraded, because no field at the pinned rev declares two kinds, but a grader that resolved a combination differently from the generator that built the stimulus would be judging by one rule what was made by another.

An absent `expected:` block still SKIPS rather than failing, deliberately: failing it would redden every legitimately unconstrained field (`sequence_id`, the clock identities). The related problem, that deleting a pin silently shrinks the campaign, is a different shape and is now tracked as **#150** with its measured evidence.

## The gate

New `tb/verilator/tsn_fuzz/test_grade_tx.py`, wired into the Makefile as `graderself` and run FIRST in `run`. It drives `grade_tx` with synthetic one-field models and hand-built frames, so it needs neither Verilator nor tsn-gen. That is the point: it guards a defect the campaigns cannot see, so it has to run where the campaigns skip.

**22 checks, 22 PASS.** Each negative case also asserts the failure MESSAGE contains the right needle, so a check that reddens for an unrelated reason does not read as a catch. The mask fixtures run at the widths that actually occur - 16, 32 and 64 bits - with the undefined bit placed above any narrower arithmetic, because an all-8-bit fixture set cannot see a grader that truncates. `MIN_CASES` refuses to report success on an emptied case list.

## Mutation proof

Reverting `grade_tx` to its pre-#146 shape (drop the `mask` branch and the `else`) takes the self-test red. Three further mutants, all found by review rather than by me, are also caught: truncating the mask arithmetic to 8 bits fails 3 of 22 (one per width); dropping the constraint kind from the failure message fails 1 of 22; emptying `CASES` exits 1, and `make graderself` exits 2.

The original revert produces:

```
[FAIL] mask: no undefined bit set                 pass=0/1 fail=0/0
[FAIL] mask: an undefined bit IS set              pass=0/0 fail=0/1  needle MISSING
[FAIL] mask: every defined bit set is still legal pass=0/1 fail=0/0
[FAIL] unknown kind fails closed                  pass=0/0 fail=0/1  needle MISSING
[FAIL] unknown kind names the field               pass=0/0 fail=0/1  needle MISSING
```

Read `pass=0/1 fail=0/0` carefully: the old grader produced **no check at all** for those fields, neither passing nor failing. That is the silent skip, visible.

`make graderself` exits **2** with the reverted grader and **0** with it restored, so the failure genuinely propagates rather than being printed and swallowed.

## Blast radius

None to the existing tally. No 802.1AS model uses `mask` or any kind outside `value` / `values` / `range`, verified across all seven `protocols/data_link/ptp/*.yaml` at the CI-pinned tsn-gen rev, so the campaign adds no checks and removes none. `mask:` is live in 7 fields of the 1722.1 models (`1722_1_adp.yaml`, `aecp_address_access.yaml`, `aecp_aem_acquire_entity.yaml`, `aecp_aem_lock_entity.yaml`, `aecp_aem_set_stream_info.yaml`), which `fuzz_ptp` does not grade today; the hole was latent for the PTP tier and reopened the moment anyone added one.

`python3 scripts/docs_check.py` clean, 0 findings across 211 md files. No RTL touched.

### DIFF
--- a/tb/verilator/tsn_fuzz/Makefile
+++ b/tb/verilator/tsn_fuzz/Makefile
@@ -70,7 +70,7 @@ AAF_SRCS = $(RTL)/common/ethernet_packet_pkg.sv \
 AAF_FLAGS = --top-module avtp_rxmon_wrap -GCLK_FREQ_HZ_P=10000 \
             +incdir+$(RTL)/common +incdir+$(RTL)/ieee1722/avtp --Mdir obj_aaf
 
-.PHONY: all build run aaf ptp matrix matrix-check clean
+.PHONY: all build run graderself aaf ptp matrix matrix-check clean
 
 all: run
 
@@ -85,8 +85,15 @@ gptp_ucode.hex: $(GPTP_DIR)/hdl/ucode/gen_gptp_ucode.py
 obj_ptp/Vptp_cosim: cosim_ptp.cpp cosim_axis.h $(PTP_SRCS) gptp_ucode.hex
 	$(VERILATOR) $(VCOMMON) $(PTP_FLAGS) $(PTP_SRCS) cosim_ptp.cpp -o Vptp_cosim
 
-run: build aaf ptp matrix-check
+run: graderself build aaf ptp matrix-check
 	@echo "== tsn_fuzz: all campaigns passed =="
+
+# The field grader's own self-test. Deliberately FIRST and deliberately free of
+# Verilator and tsn-gen: it guards a defect the campaigns cannot see, because a
+# constraint kind grade_tx does not know produces no check to fail. It therefore
+# has to run even where the campaigns skip.
+graderself:
+	@$(PY) test_grade_tx.py
 
 # regenerate the module<->spec<->test traceability matrix from the tree
 matrix:
--- a/tb/verilator/tsn_fuzz/fuzz_ptp.py
+++ b/tb/verilator/tsn_fuzz/fuzz_ptp.py
@@ -261,6 +261,11 @@ class Campaign:
         for name, _bits, con in model.fields:
             if name in skip or name not in got or not con:
                 continue
+            # Dispatch order matches tsn_model.legal()/illegal() exactly:
+            # value, values, range, mask. No field declares two kinds today,
+            # but a grader that resolved a combination differently from the
+            # generator that produced the stimulus would be judging by one
+            # rule what was built by another.
             if "value" in con:
                 if name in gaps:
                     self.eq_or_gap("%s.%s" % (label, name), got[name],
@@ -268,14 +273,14 @@ class Campaign:
                 else:
                     self.rep.eq("%s.%s" % (label, name), got[name],
                                 int(con["value"]))
-            elif "range" in con:
-                lo, hi = int(con["range"][0]), int(con["range"][1])
-                self.rep.ck("%s.%s in [%d,%d]" % (label, name, lo, hi),
-                            lo <= got[name] <= hi, "got=%d" % got[name])
             elif "values" in con:
                 self.rep.ck("%s.%s legal" % (label, name),
                             got[name] in [int(v) for v in con["values"]],
                             "got=%d" % got[name])
+            elif "range" in con:
+                lo, hi = int(con["range"][0]), int(con["range"][1])
+                self.rep.ck("%s.%s in [%d,%d]" % (label, name, lo, hi),
+                            lo <= got[name] <= hi, "got=%d" % got[name])
             elif "mask" in con:
                 # A mask names the DEFINED bits; the spec claim is that no bit
                 # outside that set is set. tsn_model already reads masks this
--- a/tb/verilator/tsn_fuzz/fuzz_ptp.py
+++ b/tb/verilator/tsn_fuzz/fuzz_ptp.py
@@ -276,6 +276,26 @@ class Campaign:
                 self.rep.ck("%s.%s legal" % (label, name),
                             got[name] in [int(v) for v in con["values"]],
                             "got=%d" % got[name])
+            elif "mask" in con:
+                # A mask names the DEFINED bits; the spec claim is that no bit
+                # outside that set is set. tsn_model already reads masks this
+                # way for probe generation (legal/illegal), so grade to match.
+                m = int(con["mask"][0])
+                undef = got[name] & ~m
+                self.rep.ck("%s.%s within defined bits" % (label, name),
+                            undef == 0,
+                            "got=0x%X undefined=0x%X mask=0x%X"
+                            % (got[name], undef, m))
+            else:
+                # FAIL CLOSED. A constraint kind this grader does not know must
+                # never pass in silence: that is exactly how an Announce `mask:`
+                # pin left the campaign once already, with the tally unmoved and
+                # nothing red to notice. Anyone adding a kind has to teach the
+                # grader what it asserts before the suite will go green again.
+                self.rep.ck("%s.%s constraint is gradeable" % (label, name),
+                            False,
+                            "ungradeable constraint kind %s -- teach grade_tx"
+                            % sorted(con))
         return got
 
     def become_gm(self, budget=10 * SECOND):
--- a/tb/verilator/tsn_fuzz/test_grade_tx.py
+++ b/tb/verilator/tsn_fuzz/test_grade_tx.py
@@ -78,10 +78,50 @@ CASES = [
      8, {"mask": [0x0F]}, [0x85], 0, 1, "undefined=0x80"),
     ("mask: every defined bit set is still legal",
      8, {"mask": [0x0F]}, [0x0F], 1, 0, ""),
+
+    # REAL WIDTHS. Every mask in the live models is 16, 32 or 64 bits wide
+    # (talker/listener_capabilities 16, controller_capabilities 32,
+    # tlv_length_and_mode 16, acquire/lock_entity_flags 32, msrp_flags 64) and
+    # every undefined bit that matters sits ABOVE bit 7. An 8-bit fixture set
+    # cannot see a grader that truncates -- `got & ~m & 0xFF` passes all three
+    # cases above while waving an undefined bit through five of the six real
+    # fields. Each case below puts the offending bit high enough that any
+    # narrower mask arithmetic misses it.
+    ("mask 16b talker_capabilities: legal",
+     16, {"mask": [0x803F]}, [0x80, 0x3F], 1, 0, ""),
+    ("mask 16b: undefined bit 14, above the low byte",
+     16, {"mask": [0x803F]}, [0x40, 0x00], 0, 1, "undefined=0x4000"),
+    ("mask 32b acquire_entity_flags: legal",
+     32, {"mask": [0xC0000000]}, [0xC0, 0, 0, 0], 1, 0, ""),
+    ("mask 32b: undefined bit 29, above 16 bits",
+     32, {"mask": [0xC0000000]}, [0x20, 0, 0, 0], 0, 1, "undefined=0x20000000"),
+    ("mask 64b msrp_flags: legal",
+     64, {"mask": [0xF800000000000000]},
+     [0xF8, 0, 0, 0, 0, 0, 0, 0], 1, 0, ""),
+    ("mask 64b: undefined bit 58, above 32 bits",
+     64, {"mask": [0xF800000000000000]},
+     [0x04, 0, 0, 0, 0, 0, 0, 0], 0, 1, "undefined=0x400000000000000"),
+
     ("unknown kind fails closed",
      8, {"regex": ["x"]}, [0x00], 0, 1, "ungradeable constraint kind"),
     ("unknown kind names the field",
      8, {"regex": ["x"]}, [0x00], 0, 1, "t.f"),
+    # #146 asks the failure to name the field AND THE KIND. Without this the
+    # message could drop the kind and both needles above would still be met.
+    ("unknown kind names the KIND itself",
+     8, {"regex": ["x"]}, [0x00], 0, 1, "regex"),
+
+    # Dispatch precedence. No field at the pinned rev declares two kinds, so
+    # a reordering of the if/elif chain is otherwise invisible. Pin the order
+    # tsn_model.legal() uses (value, values, range, mask) so the grader and the
+    # generator cannot silently disagree about which constraint governs.
+    ("precedence: value wins over mask",
+     8, {"value": 0x03, "mask": [0x0F]}, [0x01], 0, 1, "exp=3"),
+    ("precedence: values wins over range",
+     8, {"values": [1, 2], "range": [0, 255]}, [9], 0, 1, "legal"),
+    ("precedence: range wins over mask",
+     8, {"range": [1, 9], "mask": [0xFF]}, [200], 0, 1, "in [1,9]"),
+
     ("value: matching", 8, {"value": 0x42}, [0x42], 1, 0, ""),
     ("value: mismatching", 8, {"value": 0x42}, [0x43], 0, 1, "got=67"),
     ("range: inside", 8, {"range": [1, 9]}, [5], 1, 0, ""),
@@ -91,13 +131,22 @@ CASES = [
     # An absent `expected:` block must keep SKIPPING. Failing it here would
     # redden every legitimately unconstrained field (sequence_id, clock
     # identities), so the fail-closed rule deliberately stops at non-empty
-    # constraints. The shrinking-tally problem that causes is tracked apart.
+    # constraints. The shrinking-tally problem that causes is issue #150.
     ("no constraint at all still skips",
      8, {}, [0xFF], 0, 0, ""),
 ]
 
+# A floor, for the same reason the campaigns carry one: an empty CASES list
+# would print "0 checks: 0 PASS, 0 FAIL" and exit 0, which reads as success.
+MIN_CASES = 22
+
+
 
 def main():
+    if len(CASES) < MIN_CASES:
+        print("grade_tx self-test: FAIL, only %d cases, expected at least %d"
+              % (len(CASES), MIN_CASES))
+        return 1
     bad = 0
     for name, bits, con, pdu, want_p, want_f, needle in CASES:
         npass, nfail, msg = grade(bits, con, pdu)
--- a/tb/verilator/tsn_fuzz/test_grade_tx.py
+++ b/tb/verilator/tsn_fuzz/test_grade_tx.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: 2026 Kebag Logic
+# SPDX-License-Identifier: CERN-OHL-W-2.0
+"""
+Self-test for the campaign's field grader (`Campaign.grade_tx`).
+
+This does NOT touch the DUT, Verilator or tsn-gen: it drives grade_tx with
+synthetic one-field models and hand-built frames, so it runs everywhere the
+campaigns themselves skip. That matters, because the defect it guards is
+invisible to the campaigns by construction.
+
+WHAT IT GUARDS. grade_tx dispatches on the constraint kind a tsn-gen model
+declares. Before issue #146 the chain ended after `values`, so a field
+constrained any other way was simply not graded -- no failure, no skip line,
+no movement in the tally. The check did not exist. That is not hypothetical:
+an Announce `flags` pin was once replaced with `mask: [0x003F]` on the claim
+that the harness would catch a regression, and it would not have.
+
+So the two properties below are the point of this file:
+  1. `mask:` is graded, and it grades what a mask actually asserts.
+  2. Any kind grade_tx does NOT know fails CLOSED, naming the field and kind.
+
+A grader that silently ignores what it does not understand cannot be trusted
+to mean anything, so (2) is the durable half: it holds for constraint kinds
+nobody has invented yet, while (1) only covers the one we know about.
+"""
+
+import contextlib
+import io
+import os
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+
+from cosim import Report                                    # noqa: E402
+import fuzz_ptp                                             # noqa: E402
+
+
+class OneFieldModel:
+    """The whole surface grade_tx and extract_fields touch: `.fields`."""
+
+    def __init__(self, fields):
+        self._fields = fields
+
+    @property
+    def fields(self):
+        return self._fields
+
+
+def grade(field_bits, constraint, pdu):
+    """Grade one synthetic field and return (npass, nfail, first message).
+
+    The inner Report prints a banner and a `[FAIL]` line for every case that
+    is SUPPOSED to fail. Those are this file's expected results, not its
+    verdict, so stdout is swallowed while the grader runs -- otherwise the log
+    reads as a wall of failures and `suite_tally.py` would be scanning our
+    fixtures. Only the per-case verdict below is printed.
+    """
+    with contextlib.redirect_stdout(io.StringIO()):
+        rep = Report("grade_tx self-test", verbose=False)
+        camp = fuzz_ptp.Campaign.__new__(fuzz_ptp.Campaign)  # no cosim socket
+        camp.rep = rep
+        frame = bytes(14) + bytes(pdu)      # extract_fields reads from byte 14
+        camp.grade_tx(OneFieldModel([("f", field_bits, constraint)]),
+                      frame, "t")
+    msg = ""
+    if rep.failures:
+        _sec, what, detail = rep.failures[0]
+        msg = "%s | %s" % (what, detail)
+    return rep.npass, rep.nfail, msg
+
+
+CASES = [
+    # (name, bits, constraint, pdu, want_pass, want_fail, msg_must_contain)
+    ("mask: no undefined bit set",
+     8, {"mask": [0x0F]}, [0x05], 1, 0, ""),
+    ("mask: an undefined bit IS set",
+     8, {"mask": [0x0F]}, [0x85], 0, 1, "undefined=0x80"),
+    ("mask: every defined bit set is still legal",
+     8, {"mask": [0x0F]}, [0x0F], 1, 0, ""),
+    ("unknown kind fails closed",
+     8, {"regex": ["x"]}, [0x00], 0, 1, "ungradeable constraint kind"),
+    ("unknown kind names the field",
+     8, {"regex": ["x"]}, [0x00], 0, 1, "t.f"),
+    ("value: matching", 8, {"value": 0x42}, [0x42], 1, 0, ""),
+    ("value: mismatching", 8, {"value": 0x42}, [0x43], 0, 1, "got=67"),
+    ("range: inside", 8, {"range": [1, 9]}, [5], 1, 0, ""),
+    ("range: outside", 8, {"range": [1, 9]}, [200], 0, 1, "got=200"),
+    ("values: legal", 8, {"values": [1, 2]}, [2], 1, 0, ""),
+    ("values: illegal", 8, {"values": [1, 2]}, [3], 0, 1, "got=3"),
+    # An absent `expected:` block must keep SKIPPING. Failing it here would
+    # redden every legitimately unconstrained field (sequence_id, clock
+    # identities), so the fail-closed rule deliberately stops at non-empty
+    # constraints. The shrinking-tally problem that causes is tracked apart.
+    ("no constraint at all still skips",
+     8, {}, [0xFF], 0, 0, ""),
+]
+
+
+def main():
+    bad = 0
+    for name, bits, con, pdu, want_p, want_f, needle in CASES:
+        npass, nfail, msg = grade(bits, con, pdu)
+        ok = (npass == want_p and nfail == want_f
+              and (not needle or needle in msg))
+        if not ok:
+            bad += 1
+        print("  [%s] %-46s pass=%d/%d fail=%d/%d%s"
+              % ("ok  " if ok else "FAIL", name, npass, want_p, nfail, want_f,
+                 "" if not needle else
+                 ("  needle=%r %s" % (needle, "found" if needle in msg
+                                      else "MISSING in %r" % msg))))
+
+    print("\ngrade_tx self-test: %d checks: %d PASS, %d FAIL"
+          % (len(CASES), len(CASES) - bad, bad))
+    return 1 if bad else 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())


Mister-M-alt:
[A1] **Cross-PR interaction with #147**, flagged before either lands.

Both PRs edit `tb/verilator/tsn_fuzz/Makefile`. I test-merged them with `git merge-tree --write-tree`: they merge **cleanly** (exit 0, no conflict markers) and the result is semantically sound - `run: graderself build aaf ptp matrix-check` keeps both, and both campaign targets keep their `suite_tally.py --campaign-guard && check_results_fresh.py` chain intact.

The interaction is the **headline number**, not the merge. `run_all_suites.sh:168` captures the full `make` output into `$OUT/tsn_fuzz.log` and `suite_tally.py` sums every `N checks: P PASS, F FAIL` line in it. #147 adds a `graderself` target that prints `grade_tx self-test: 12 checks: 12 PASS, 0 FAIL`, which matches the labelled tally shape and will therefore be counted. So once both land the suite reports **533**, not 521.

Whichever merges second should update the other's documented figure. Nothing needs re-measuring beyond that arithmetic.

One thing that does NOT interact, checked rather than assumed: the freshness gate judges `obj_ptp/ptp.out`, which is tee'd from `fuzz_ptp.py` alone, so `graderself` output cannot reach the log whose tally must match the artifact headline. The tally-match rule is unaffected.

Mister-M-alt:
[R1] **POSITIVE**

Reviewed at `eb73358` against base `d320ca13`, in a throwaway clone under `/tmp`. Every claim in the PR body checks out, including the mutation numbers, and I verified claim 4 against the CI-pinned tsn-gen rev rather than the local working tree. The code change is correct and its blast radius really is zero.

The self-test is real, not decorative: it catches the reverted grader and three of the four wrong-but-plausible mutations I threw at it. It does have four survivors, one of them sharp, and I report those below. None of them makes the tree worse than it is today, so they are asks and not blockers, and two of them are one line each.

---

## 1. The premise holds, and by construction

`git show d320ca13:tb/verilator/tsn_fuzz/fuzz_ptp.py | sed -n '253,282p'` is a bare `if "value" / elif "range" / elif "values"` chain with no `else`. The only other exit is the `continue` at the top of the loop body, which is gated on `name in skip or name not in got or not con`. So a field with a non-empty `con` that is none of the three kinds executes nothing at all: no `rep.ck`, no `rep.gap`, no `print`, no `continue` line. The check does not exist.

Confirmed empirically, which is the part that makes it undeniable. Reverted grader, self-test output:

```
  [FAIL] mask: no undefined bit set                     pass=0/1 fail=0/0
  [FAIL] mask: every defined bit set is still legal     pass=0/1 fail=0/0
```

`pass=0/1 fail=0/0`. Neither a pass nor a failure. That is the silent skip on the record.

## 2. The mask semantics are right, and the agreement is wider than the PR claims

The PR argues agreement with `tsn_model.py`. That holds:

* `legal()` yields `[0, m] + [1 << b for b in range(bits) if m & (1 << b)]`, every one of which satisfies `v & ~m == 0`;
* `illegal()` yields `[undef] + single undefined bits`, where `undef = (~m) & full`, every one of which fails it.

So each generated legal probe passes the new check and each generated illegal probe fails it. Exact.

But `tsn_model.py` is this repo's re-reading of the model, so I went to the generator itself, at the pinned rev:

```
$ git show fde6520:docs/03_protocol_yaml_reference.md | grep -n mask
67:| `expected.mask` | No | `[M]` | Only bits within `M` may be set in the random value. |
97:      mask: [0xC0000000]       # only the top two bits may be set

$ git show fde6520:traffic-gen/src/packet_builder.cpp
    /* Bitmask: only bits within the mask may be set. */
    if (var.hasMask()) {
        value &= var.getMask();
    }
```

tsn-gen's reference doc, its authoring guide, and its C++ builder all say the same thing the new branch asserts. Grader, generator and builder agree. Not a blocker, the opposite: this is the strongest-grounded part of the change.

Spot-checked two of the seven live masks against the standard, and the model reading is right there too: `acquire_entity_flags` mask `0xC0000000` is exactly PERSISTENT plus RELEASE, and `lock_entity_flags` mask `0x80000000` is exactly UNLOCK.

## 3. The mutation claim reproduces exactly

```
$ python3 test_grade_tx.py            # PR as merged
grade_tx self-test: 12 checks: 12 PASS, 0 FAIL      rc=0

$ # grade_tx reverted to its pre-#146 shape
$ python3 test_grade_tx.py
  [FAIL] mask: no undefined bit set                     pass=0/1 fail=0/0
  [FAIL] mask: an undefined bit IS set                  pass=0/0 fail=0/1  needle='undefined=0x80' MISSING in ''
  [FAIL] mask: every defined bit set is still legal     pass=0/1 fail=0/0
  [FAIL] unknown kind fails closed                      pass=0/0 fail=0/1  needle='ungradeable constraint kind' MISSING in ''
  [FAIL] unknown kind names the field                   pass=0/0 fail=0/1  needle='t.f' MISSING in ''
grade_tx self-test: 12 checks: 7 PASS, 5 FAIL       rc=1

$ make graderself                                    MAKE_RC=2
```

7 PASS, 5 FAIL, exit 1, `make graderself` exit 2, and the five are the five the PR names. Verbatim.

## 4. No tally movement, verified against the PIN

Read straight out of `git show fde652067d3618f47fba2c2e8eb05f6c9f6fa97f:<path>`, never the working tree. Constraint-kind histogram over the whole `protocols/` tree at that rev:

```
  8021as_announce.yaml               {'value': 16, 'values': 1, 'range': 1, '<none>': 10}
  8021as_eth_header.yaml             {'value': 2, '<none>': 1}
  8021as_follow_up.yaml              {'value': 14, 'values': 1, 'range': 2, '<none>': 9}
  8021as_pdelay_req.yaml             {'value': 13, 'range': 1, '<none>': 3}
  8021as_pdelay_resp.yaml            {'value': 11, 'range': 3, '<none>': 5}
  8021as_pdelay_resp_follow_up.yaml  {'value': 11, 'range': 3, '<none>': 5}
  8021as_sync.yaml                   {'value': 12, 'values': 1, 'range': 1, '<none>': 2}

  TOTAL across all protocols/: {'value': 135, 'values': 67, 'range': 29, 'mask': 7}
```

Zero `mask` and zero unrecognised kinds in all seven PTP models, so the new branches are unreachable from `fuzz_ptp`. The seven live masks are exactly the seven the PR names, in exactly the five files it names. No field anywhere in the tree declares two constraint kinds.

Rather than stop at exhaustion over the model set, I drove the real pinned models through both graders:

```
### AFTER (PR #147)                      ### BEFORE (pre-#146)
  8021as_announce.yaml         15          8021as_announce.yaml         15
  8021as_eth_header.yaml        2          8021as_eth_header.yaml        2
  8021as_follow_up.yaml        17          8021as_follow_up.yaml        17
  8021as_pdelay_req.yaml       14          8021as_pdelay_req.yaml       14
  8021as_pdelay_resp.yaml      14          8021as_pdelay_resp.yaml      14
  8021as_pdelay_resp_fu.yaml   14          8021as_pdelay_resp_fu.yaml   14
  8021as_sync.yaml             14          8021as_sync.yaml             14
  TOTAL: 90                                TOTAL: 90
```

Identical. Zero checks added, zero removed. Claim confirmed.

For the record on the trap: local `tsn-gen` HEAD `f52f1f9a` is NOT a descendant of the pin, but `git diff fde6520 HEAD -- protocols/data_link/ptp/` is empty, so the working tree happens to give the same answer here. That was luck, not method.

## 5. The deliberate exclusion is the right call, but the deferral is not backed

Failing on an absent `expected:` would be wrong, and I agree with the PR. The two cases are different claims: an unknown kind means the model asserts something the grader cannot read, while an absent block means the model asserts nothing. Only the first should fail closed. Concretely, at the pin the seven PTP models carry 35 unconstrained fields, so the alternative reddens 35 legitimate fields.

What is not backed is "is tracked separately". `gh issue list --state all --limit 100` plus two `gh api search/issues` passes over titles and bodies return no such issue. The only issues mentioning it are #146, which this PR closes, and #147 itself. #148 and #149 are the stale gPTP TEST_RESULTS work, unrelated.

That matters more than usual here, because deleting a pin is how the campaign already shrank once without anyone noticing, and #146's own text waves it off with "That one is visible in the tally at least". Nothing gates the tally against a previous run, so it is visible only to someone who happens to compare. **Please file the issue before merging this one**, otherwise closing #146 closes the file on both halves of the hole.

## 6. Output hygiene is clean

The `redirect_stdout` cannot hide a real failure, because the verdict never reads stdout: `grade()` returns `rep.npass`, `rep.nfail` and `rep.failures[0]`, which are object state that `Report.ck` updates before it prints. An exception inside the `with` escapes the context manager, prints to stderr, and exits non-zero, so that path is loud too.

`suite_tally.py` is not confused. Scanning the real graderself output:

```
  checks=12 failures=0
  matched  = [('n-checks-pass-fail', 12, 0, 'grade_tx self-test: 12 checks: 12 PASS, 0 FAIL')]
  UNPARSED = []
```

One clean match, no UNPARSED. The twelve per-case lines (`pass=1/1 fail=0/0`) trip neither `SHAPES` nor `TALLY_SHAPED`, since the latter wants `N pass,` with a space and a comma. `--campaign-guard` reads only `obj_aaf/aaf.out` and `obj_ptp/ptp.out`, which `graderself` never writes, so the two campaign guards are untouched.

Worth stating out loud so nobody is surprised: the tsn_fuzz **suite** headline gains +12 checks, because `run_all_suites.sh` redirects all of `make` into the suite log and `suite_tally` reads that line. That is honest new work and not a contradiction of the PR, whose "no tally movement" claim is about the campaign. Just flagging it so the next sweep delta is expected.

## 7. Makefile placement is correct

With an emptied environment, no verilator on PATH and a bogus generator root:

```
$ env -i PATH=/usr/bin:/bin HOME=/nonexistent TSN_GEN_ROOT=/nonexistent/tsn-gen \
        VERILATOR=/nonexistent/verilator make graderself
grade_tx self-test: 12 checks: 12 PASS, 0 FAIL      RC=0
```

Needs neither Verilator nor tsn-gen, as claimed. And a failing self-test does abort the whole run, before `build`:

```
$ env -i ... make run
grade_tx self-test: 12 checks: 7 PASS, 5 FAIL
make: *** [Makefile:96: graderself] Error 1              MAKE_RUN_RC=2
```

`run_all_suites.sh:168` invokes `timeout "$TMO" make -C "$d"` with no `-j`, so prerequisites stay serial and `graderself` really is first.

---

## Findings: what the self-test does NOT catch

The PR proves the forward direction. I ran the converse. Seven mutations, four survive with the self-test at **12 PASS, 0 FAIL, rc=0**.

Caught, for completeness: mask branch inverted (9/3), mask branch always passes (11/1), `else` downgraded from `ck(False)` to `gap()` (10/2). Good.

**F1. A width bug in the mask branch is invisible. This is the sharp one.**

Every fixture in `CASES` uses `8` bits. Every mask field that actually exists is 16, 32 or 64 bits. So clamping the check to the low byte survives:

```python
undef = got[name] & ~m & 0xFF     # the mutation
```

Driven with the six real mask fields from the pinned rev, each given an undefined bit above bit 7:

```
  adp.talker_capabilities     bits=16 mask=0x803F              val=0x100          -> *** ACCEPTED ***
  adp.controller_capabilities bits=32 mask=0x8000              val=0x10000        -> *** ACCEPTED ***
  aecp.acquire_entity_flags   bits=32 mask=0xC0000000          val=0x10000        -> *** ACCEPTED ***
  aecp.lock_entity_flags      bits=32 mask=0x80000000          val=0x10000        -> *** ACCEPTED ***
  aecp.tlv_length_and_mode    bits=16 mask=0xFECF              val=0x10           -> CAUGHT
  aecp.msrp_flags             bits=64 mask=0xF800000000000000  val=0x10000000000  -> *** ACCEPTED ***

  grade_tx self-test: 12 checks: 12 PASS, 0 FAIL       rc=0
```

Five of six real fields wave an undefined bit through, and the gate stays green. That is precisely the failure shape this PR exists to kill: a check that appears to assert something and asserts less. One extra case with a 32-bit field and a high undefined bit closes it. Please add it here.

**F2. The `else` is not proven to name the kind, only the field.**

Deleting `% sorted(con)` from the detail string leaves both needles intact, so it survives:

```
  failure record: (None, 't.f constraint is gradeable', 'ungradeable constraint kind -- teach grade_tx')
  grade_tx self-test: 12 checks: 12 PASS, 0 FAIL       rc=0
```

The kind is gone and the gate does not notice. This is a direct miss against #146's own acceptance list, which says "fails loudly and names the field **and the kind**", and the case is even named `unknown kind names the field`. Adding `"regex"` as a needle on that case is one character of real work.

**F3/F4. Dispatch order is unguarded.** Swapping `range` and `values`, or hoisting `mask` ahead of `value`, both survive at 12/12, because no fixture declares two kinds. Low severity today, and I confirmed no field at the pin declares two either. But tsn-gen documents a precedence (`value`/`values` beats `range` beats `mask`, `docs/mid-level/writing-protocols.md:56`) that nothing in `grade_tx` enforces or tests. Worth a one-line case with `{"value": 5, "mask": [0x0F]}` asserting the value branch wins, if you want the gate to encode the precedence rather than inherit it from source order.

**F5. The self-test has no floor, so it can be emptied silently.** Truncating `CASES` to `[]`:

```
grade_tx self-test: 0 checks: 0 PASS, 0 FAIL         SELFTEST_RC=0
make graderself                                      MAKE_GRADERSELF_RC=0
```

`suite_tally` would call that log NOCOUNT on its own, but the unconditional 2-check `matrix-check` floor rescues the suite, so the sweep goes green:

```
  vacuous alone:  checks=0 failures=0 NOCOUNT?=True
  vacuous+floor:  checks=2 failures=0 NOCOUNT?=False
```

The Makefile's own fifteen-line comment above `aaf` describes this exact hazard and solves it with `suite_tally.py --campaign-guard`. The new target is the one place in this Makefile that does not get that treatment. Either give `graderself` the same tee-and-guard discipline, or assert a minimum case count inside `main()`. Not a blocker, but it is the one thing that would let this gate become a gate that asserts nothing, which is the thing the PR is about.

---

## Verdict

POSITIVE. Merge it. The grader change is correct, the semantics are triple-sourced, the pinned-rev blast radius is provably nil, and the gate does what it says. F1 and F2 are each one line and would close real holes in the new gate; I would like them in this PR. F5 and the missing tracking issue from section 5 are the durability asks, and the tracking issue is the one I would not let slide, since without it #146 closes over a hole nobody owns.

Note for the record: `verilator-suites` and `yosys-portability` were still pending when I reviewed. The other four checks pass, and I reproduced `docs_check.py` locally at `0 finding(s) across 211 md files`. Nothing was committed or pushed; my clone is clean and no existing worktree was touched.

Mister-M-alt:
[A1] All five findings fixed in `ddc599bd`. F1 is the one that mattered and it lands squarely on this PR's own thesis, so I will take it first.

**F1: the self-test asserted less than it appeared to.** Every fixture was 8 bits wide; every mask in the live models is 16, 32 or 64. `undef = got & ~m & 0xFF` therefore passed all twelve checks while waving an undefined bit through five of the six real fields. That is exactly the "looks like a check, asserts nothing" shape this PR exists to kill, reproduced inside the file written to kill it. You were right to go looking for it rather than accepting the mutation proof I offered.

Fixed by testing at the widths that actually occur, with the offending bit placed above any narrower arithmetic:

| fixture | mask | undefined bit |
|---|---|---|
| 16b `talker_capabilities` | `0x803F` | `0x4000`, above the low byte |
| 32b `acquire_entity_flags` | `0xC0000000` | `0x20000000`, above 16 bits |
| 64b `msrp_flags` | `0xF800000000000000` | `0x0400000000000000`, above 32 bits |

Your truncation mutant now fails **3 of 22**, one per width, each naming the bit it let through.

**F2: the gate proved the field was named but never the kind**, which misses #146's own acceptance wording. Added a third needle asserting the kind itself appears. Dropping `% sorted(con)` from the message now fails 1 of 22 with `needle='regex' MISSING in 't.f constraint is gradeable | ungradeable constraint kind -- teach grade_tx'`.

**F3/F4: dispatch precedence was not merely unenforced, it was wrong.** Pinning it surfaced a real disagreement: `grade_tx` tried `range` before `values`, while `tsn_model.legal()`/`illegal()` try `values` before `range`. No field at the pinned rev declares two kinds, so nothing was misgraded, but a grader resolving a combination differently from the generator that built the stimulus would be judging by one rule what was made by another. `grade_tx` now follows `tsn_model`'s order exactly - value, values, range, mask - with a comment saying why, and three fixtures pin each adjacent pair.

**F5: the target had no floor.** Emptying `CASES` gave `0 checks: 0 PASS, 0 FAIL` and exit 0. Added `MIN_CASES`, checked before anything runs. Measured: with `CASES = []` the script exits **1** and `make graderself` exits **2**; restored, `make` exits 0. You are right that this target lacks the `--campaign-guard` discipline the Makefile prescribes for its neighbours; the floor is the cheap equivalent, and I would rather say that plainly than claim parity.

**And the one you asked not be let slide: you were right, it was false.** "The related problem is tracked separately" described an issue that did not exist. I have filed it as **#150**, and the code comment now names it rather than gesturing at a tracking that was not happening. #150 carries the measured evidence: dropping the `tx_pdreq.correction_field` and `tx_announce.steps_removed` pins took the campaign from 355 to 353 with zero failures either way, both unpins spec-justified, which is precisely what makes the mechanism hard to police. Merging this no longer closes #146 over a hole nobody owns.

Self-test is **22 checks, 22 PASS**. `docs_check.py` clean, 0 findings across 211 files. Thank you for the corroboration on mask semantics too - `packet_builder.cpp:74` doing `value &= var.getMask()` is stronger evidence than the two Python call sites I had.

Note for sequencing: this PR and #149 both edit the tsn_fuzz Makefile. They merge cleanly and I verified the merged targets are semantically sound, but the suite total shifts, so whichever lands second updates the other's documented figure. Details in the comment above.

Mister-M-alt:
[A1] **Correction to my cross-PR note above.** #147's self-test grew from 12 checks to 22 while answering its review, so the arithmetic moves: once both land the tsn_fuzz suite reports **543**, not the 533 I said, against the 521 documented in #149. The mechanism is unchanged - `run_all_suites.sh:168` captures the full make output and `suite_tally.py` sums every `N checks: P PASS, F FAIL` line in it, and the `graderself` line matches that shape. Whichever PR lands second updates the other's figure.

Mister-M-alt:
[R1] **POSITIVE** (re-review at `ddc599bd18bc11ee0dfd9ef47cfc8fd58874b87e`)

Supersedes my review at `eb73358`. Re-ran every mutation from that round against the new 22-case suite rather than taking the answers on trust, plus six new ones aimed at the new fixtures. All five findings are genuinely answered. The dispatch reorder is not just safe, it is a real bug fix, and I verified that against the generator's own C++ rather than its docs.

**A correction to my own method first, since it cuts against me.** My first pass at the reorder mutants was contaminated by stale `__pycache__/fuzz_ptp.cpython-314.pyc`: three different mutants printed a byte-identical failure line, which is what gave it away. Every number below is from a re-run with `python3 -B` and the cache purged between mutants. One conclusion flipped as a result, and I report it flipped. This is the same trap as my F1: a proof that looked stronger than it was.

---

## Mutation results at the new head, 13 mutants, cache purged

```
MUTATION                 rc   SUMMARY                                         VERDICT
revert                   1    grade_tx self-test: 22 checks: 10 PASS, 12 FAIL caught
mask_inverted            1    grade_tx self-test: 22 checks: 13 PASS,  9 FAIL caught
mask_always_pass         1    grade_tx self-test: 22 checks: 18 PASS,  4 FAIL caught
else_reports_gap         1    grade_tx self-test: 22 checks: 19 PASS,  3 FAIL caught
mask_clamp8              1    grade_tx self-test: 22 checks: 19 PASS,  3 FAIL caught
mask_clamp16             1    grade_tx self-test: 22 checks: 20 PASS,  2 FAIL caught
mask_clamp32             1    grade_tx self-test: 22 checks: 21 PASS,  1 FAIL caught
else_hides_kind          1    grade_tx self-test: 22 checks: 21 PASS,  1 FAIL caught
reorder_values_range     1    grade_tx self-test: 22 checks: 21 PASS,  1 FAIL caught
mask_first               1    grade_tx self-test: 22 checks: 20 PASS,  2 FAIL caught
empty_cases              1    grade_tx self-test: FAIL, only 0 cases, expected at least 22
reorder_value_values     0    grade_tx self-test: 22 checks: 22 PASS,  0 FAIL *** survives ***
vacuous_22               0    grade_tx self-test: 22 checks: 22 PASS,  0 FAIL *** survives ***
```

Eleven of thirteen caught, against four of seven last round. Both survivors are discussed below and neither is a defect.

## F1 answered, and the widths are load-bearing rather than decorative

`got & ~m & 0xFF` now fails 3 of 22, exactly as claimed. I pushed further to check the new cases are not just three restatements of one idea, by clamping at each width:

* `& 0xFF` caught by the 16b, 32b and 64b cases (3 failures);
* `& 0xFFFF` caught by the 32b and 64b cases only (2 failures);
* `& 0xFFFFFFFF` caught by the **64b case alone** (1 failure).

So each of the three widths is individually load-bearing. The 64-bit fixture in particular is the only thing standing between a 32-bit truncation and a green suite, which is the case I would most have expected to be ornamental. It is not. Failing line for `mask_clamp32`:

```
  [FAIL] mask 64b: undefined bit 58, above 32 bits   pass=1/0 fail=0/1  needle='undefined=0x400000000000000' MISSING in ''
```

## F2 answered

Dropping `% sorted(con)` now fails 1 of 22, and the failure is legible about what was lost:

```
  [FAIL] unknown kind names the KIND itself  pass=0/0 fail=1/1
         needle='regex' MISSING in 't.f constraint is gradeable | ungradeable constraint kind -- teach grade_tx'
```

#146's acceptance criterion "names the field **and the kind**" is now proven by mutation rather than by reading.

## F3/F4: the reorder is a real fix, and I checked the direction independently

You asked me to satisfy myself on two things. Both hold.

**Is `tsn_model`'s order the right one to standardise on, or should the fix have gone the other way?** `tsn_model` is this repo's re-reading of the model, so I did not want to settle it from that file. Upstream at the pinned rev, `parser/src/protocol.cpp:146-186` parses an `expected:` block into three separate slots, and `traffic-gen/src/packet_builder.cpp` resolves them in a fixed order:

```
uint64_t PacketBuilder::pickValue(const Var& var)
{
    const auto& expected = var.getExpectedValues();

    /* Exhaustive list takes priority over all other constraints. */
    if (!expected.empty()) { ... }

    /* Inclusive [min, max] range. */
    if (var.hasRange()) { ... }
```

and then the uniform draw is ANDed with the mask. So the generator's own precedence is `value`/`values` beats `range` beats `mask`. Its docs agree at the same rev (`docs/mid-level/writing-protocols.md:56`, `docs/mid-level/using-the-cli.md:93`). The OLD `grade_tx` had `range` ahead of `values`, which contradicts "Exhaustive list takes priority over all other constraints"; the new order matches upstream exactly. The reorder went the right way, and reversing it to match the old grader would have entrenched the disagreement. Good catch by the author, and a genuine defect surfaced by the act of writing the fixture.

One more agreement I noticed while in there: the upstream parser takes only the first mask entry (`break; /* take first entry only */`), which is exactly `con["mask"][0]` in the new branch.

**Can the reorder change an existing verdict?** No, and I did not settle for comparing totals this time. I instrumented `Report.ck` to record every emitted check name and its verdict in order, then ran the pre-#146 grader and the `ddc599bd` grader over all seven pinned PTP models with three different frame fill patterns:

```
new head:  TOTAL EMITTED: 270
pre-#146:  TOTAL EMITTED: 270
$ diff emit_old.txt emit_new.txt
IDENTICAL: the reorder changes no verdict, no name, no ordering
```

270 checks, identical in name, verdict and sequence. Confirmed against the pin that no field in `protocols/` declares two constraint kinds, so the only pair whose relative order moved (`values` before `range`) is unreachable from any live model. Claim 4 still holds at the new head: zero checks added, zero removed.

**Survivor 1, and why I am NOT reporting it as a finding.** Swapping the adjacent top pair, `values` ahead of `value`, survives at 22/22. I flagged that shape as unpinned before running it. It turns out to be harmless for a reason worth recording: upstream does not distinguish them at all. `protocol.cpp` pushes both into the same vector:

```
/* expected.values: [v0, v1, ...] -- exhaustive allowed list */
if (expectedNode.has_child("values")) { ... expectedValues.push_back(...) }
/* expected.value: N -- single fixed value */
if (expectedNode.has_child("value"))  { expectedValues.push_back(...) }
```

There is no `value` vs `values` precedence to pin, because upstream unions them into one kind. Pinning an order there would assert something the generator does not believe. The three fixtures cover every pair that is actually distinguishable. Leave it.

(Latent, pre-existing, not this PR: because upstream unions them, a field declaring both `value: 3` and `values: [1,2]` has legal set `{1,2,3}` upstream, while both `grade_tx` and `tsn_model.legal()` would take only `3` and reject a frame the generator legitimately produced. No model does this, `tsn_model` has had the same reading since before #146, and fixing it is not this PR's job. Noting it only so it is written down somewhere.)

## F5 answered, with one residue

The floor works and it propagates:

```
empty_cases: grade_tx self-test: FAIL, only 0 cases, expected at least 22
             python3 rc=1        make graderself rc=2
```

`MIN_CASES = 22` against `len(CASES) == 22` leaves zero slack, so removing any case trips it and removing one deliberately requires editing the constant. That is the property you wanted.

**Survivor 2.** A count-based floor counts cases, not assertions. Replacing all 22 with `("filler N", 8, {}, [0x00], 0, 0, "")` gives 22 cases that emit no checks at all, and the suite reports `22 checks: 22 PASS, 0 FAIL`, rc=0, `make graderself` rc=0. The realistic version is not sabotage but rebalancing: drop the nine mask cases, pad with nine copies of something cheap, stay at 22, stay green.

I am reporting this as a residue and not a blocker, because it is inherent to every floor of this shape, including the `--campaign-guard` floor it was modelled on, and because the current state is strictly better than the unguarded version I reviewed. If you want it closed cheaply, assert a second floor over the cases that expect a failure, something like `sum(1 for c in CASES if c[5]) >= 13`. A file full of fillers then has to fake failures too, which is no longer something anyone does by accident.

## The tracking issue is real

#150 is not a placeholder. It states the mechanism, and it grounds the shrink with a controlled experiment at the pinned rev rather than an anecdote: 355 to 353, with a per-row table naming `tx_pdreq.correction_field` (`8021as_pdelay_req.yaml` dropped `value: 0`) and `tx_announce.steps_removed` (`8021as_announce.yaml` dropped `range: [0, 254]`), and it argues both unpins were spec-correct, which is what makes the problem hard rather than merely annoying. It carries four acceptance criteria including "proven by mutation", and the code comment at `test_grade_tx.py:134` names it. That is a real handover, and my section 5 objection is fully answered.

## Unchanged from my last review, re-confirmed at this head

* `graderself` still needs neither Verilator nor tsn-gen: `env -i PATH=/usr/bin:/bin TSN_GEN_ROOT=/nonexistent VERILATOR=/nonexistent make graderself` gives `22 checks: 22 PASS, 0 FAIL`, rc=0.
* `suite_tally` reads the new summary as one clean `n-checks-pass-fail` match with zero UNPARSED; the per-case `pass=1/1 fail=0/0` lines trip nothing. Suite headline now gains +22 rather than +12.
* `python3 scripts/docs_check.py` gives `0 finding(s) across 211 md files`.
* No em dash or en dash on any line added since `eb73358`.

---

## Verdict

POSITIVE at `ddc599bd`. Merge it.

All five findings answered, and F3/F4 turned out to be worth more than I asked for: pinning the precedence exposed a real disagreement between `grade_tx` and the generator, the fix went in the direction upstream's own source supports, and I proved by full emitted-check diff that it moves no existing verdict. The two surviving mutants are a no-op against upstream semantics and a generic property of count-based floors respectively. Neither blocks.

The one thing I would still like, at your discretion and not as a condition of merge, is the second floor over failure-expecting cases described under F5.

Checks at the time of writing: `bdd-conformance`, `docs-check`, `docs-check-no-git` and `wire-accountability` pass; `verilator-suites` and `yosys-portability` still pending, and `verilator-suites` is the one that actually exercises `graderself`, so it is worth a glance before merging. Reviewed in a throwaway clone; nothing committed or pushed, no existing worktree touched, every tree left clean.

Mister-M-alt:
[A1] Thank you for the re-review, and particularly for disclosing the `__pycache__` contamination and re-running with `python3 -B` rather than letting the first conclusion stand. That is the same failure mode this PR exists to fix, one layer up: a result that looked like evidence and was not.

Three things from your re-review that I am recording rather than acting on here.

**Each width is load-bearing, which I had not established.** I added the 16/32/64 fixtures because 8 bits was obviously too narrow, but you showed the finer fact: `& 0xFFFF` fails 2 of 22 and `& 0xFFFFFFFF` fails **1**, caught by the 64-bit case alone. So the 64b fixture is not redundant coverage of the same idea, it is the only thing standing between a 32-bit truncation and green. Worth knowing before anyone trims the case list.

**The reorder went the right way for a better reason than I had.** I aligned `grade_tx` to `tsn_model` because two consumers in this repo disagreed. You verified the direction against upstream's own C++ - `packet_builder.cpp::pickValue` resolving `expectedValues` first, then `hasRange()`, then mask - which is the actual authority. And instrumenting `Report.ck` to diff every emitted check name and verdict across 7 pinned models by 3 fill patterns, 270 checks identical, is a stronger no-verdict-change proof than anything I offered.

**Your two survivors, and what I did with them.**

- The `value`/`values` swap surviving is harmless today for exactly the reason you give, and I have taken your advice not to add a fixture asserting a precedence the generator does not believe in. But the underlying disagreement is real and latent: both consumers here implement "first kind wins" while upstream pushes `value` and `values` into the same `expectedValues` vector, so a field declaring both would have legal set `{1,2,3}` upstream and be graded "must equal 3" here. That would fail a conformant device and blame it in the message. Filed as **#151**, with refusing multi-kind fields as the likely close, since there is no useful meaning for one and refusing removes the class rather than picking a winner.
- `MIN_CASES` being satisfiable by 22 assert-nothing fillers is inherent to a count floor, as you say. I am deliberately NOT adding the second floor in this PR: it is discretionary by your own assessment, and pushing another commit would invalidate this positive at a third head for a hole that requires someone to deliberately write 22 empty fixtures. Recorded here so it is a known limit rather than an unnoticed one.

Holding the merge until `verilator-suites` lands, since you are right that it is the job which actually exercises `graderself`; the other four are green. Head stays `ddc599bd`.

PR #149, opened 2026-08-20
Regenerate the stale gPTP campaign results and gate their freshness

### LINKED ISSUE #148: Stale generated gPTP TEST_RESULTS.md, and no gate compares it against a fresh run
## The stale artifact

`hdl/ieee8021as/gptp_plane/doc/TEST_RESULTS.md` is generated by the gPTP field
campaign (`tb/verilator/tsn_fuzz/fuzz_ptp.py`) and says "Do not hand-edit". The
copy committed on `dev` claims:

```
**PASS** - 355 pass, 0 fail, 9 known gaps  .  2026-08-19 21:09
| boot: the first Pdelay_Req, graded field by field | 24 | 0 | 0 |
| as grandmaster: Announce / Sync / Follow_Up conform | 56 | 0 | 3 |
```

A fresh run at the CI-pinned tsn-gen rev (`fde65206`, `.github/workflows/rtl.yml`)
produces **353 pass, 0 fail, 9 known gaps**, with those two rows at **23** and
**55**. Zero failures either way, so this is not an RTL regression: two checks
stopped being graded.

Timeline: `cf208088` (08-19 21:22) regenerated the artifact against the
then-local tsn-gen models. `1185d512` and `6e3aef15` (08-20) then moved
`TSN_GEN_REV` to `affe6a35` and finally `fde65206`. Nobody re-ran the campaign
against the new pin, so the committed counts describe models that no longer
exist.

Both lost checks are spec-justified unpins in the tsn-gen models, not a
coverage regression:

- `8021as_pdelay_req.yaml` `correction_field` dropped `expected: value: 0`,
  because IEEE 802.1AS-2011 Table 11-5 assigns Pdelay_Req a correctionField
  role, so pinning zero would fail a conformant device.
- `8021as_announce.yaml` `steps_removed` dropped `expected: range: [0, 254]`,
  because 10.3.10.2.1(b) is a receiver qualification rule and 10.5.3.2.6 puts
  no bound on the transmitted UInteger16. Bounding it would encode a receive
  rule into a transmit model and make the >= 255 stimulus unreachable.

`grade_tx()` in `fuzz_ptp.py` skips any field with no constraint, so unpinning
a field silently removes its check from the tally.

## The missing gate

Nothing compares the committed `TEST_RESULTS.md` against a fresh run. There is
no such check in `.github/`, in `scripts/`, or in the `tb/verilator/tsn_fuzz/`
Makefile. The repo already has the right pattern for this problem: `make
matrix-check` fails the build when the committed
`docs/traceability/MODULE_MATRIX.md` is stale.

Two constraints the gate has to respect:

1. The campaign needs tsn-gen and SKIPS cleanly when it is absent. The gate
   must skip honestly in that case, never silently pass while claiming
   freshness.
2. The artifact embeds a generation timestamp, so a byte comparison would go
   red on every run for a trivial reason. The sibling artifact
   `hdl/ieee1722/avtp/doc/TEST_RESULTS.md` is substantively fresh today and
   differs from its committed copy in the timestamp alone, so this is live,
   not hypothetical.

CI only provides tsn-gen in the `verilator-suites` job, which runs
`scripts/run_all_suites.sh`, so the gate belongs in the tsn_fuzz suite next to
`matrix-check` rather than in the docs job.

## Definition of done

- The committed artifact matches what the campaign actually produces.
- A freshness gate fails on a stale committed copy, ignores the timestamp,
  skips honestly without tsn-gen, and carries a self-test proving it can fail.

Mister-M-alt:
[A1]

## Status

Validated. Closes #148.

## Description

`hdl/ieee8021as/gptp_plane/doc/TEST_RESULTS.md` is generated by the gPTP field
campaign and says "Do not hand-edit". The copy on `dev` claimed **355 pass, 0
fail, 9 known gaps**. A fresh run at the CI-pinned tsn-gen rev produces **353
pass, 0 fail, 9 known gaps**. Zero failures either way, so nothing regressed in
the RTL: two checks stopped being graded and the file went on asserting the old
number, with every gate in the repository green.

This PR regenerates the artifact and adds the gate that would have caught it.

### The true tally, measured

353 pass, 0 fail, 9 known gaps, deterministic across three runs. The two rows
that moved are `boot` 24 -> 23 and `as grandmaster` 56 -> 55.

The delta was isolated by a controlled experiment rather than inferred: one
`packet_gen` binary at the pinned rev, run twice, with only the seven
`8021as_*.yaml` model files swapped between the state that produced 355 and the
pinned state. The old models reproduce 355 exactly, which pins the models as
the only variable. Diffing the two verbose check-label sets accounts for every
line.

### Per-field account of the delta

Two checks were lost, both because the tsn-gen model dropped an `expected:` pin,
and `grade_tx()` skips a field carrying no constraint:

| check | model change | why the pin went |
|---|---|---|
| `tx_pdreq.correction_field` | `8021as_pdelay_req.yaml` dropped `value: 0` | IEEE 802.1AS-2011 Table 11-5 assigns Pdelay_Req a correctionField role, so a conformant requester may carry a nonzero value and pinning zero would fail it |
| `tx_announce.steps_removed` | `8021as_announce.yaml` dropped `range: [0, 254]` | 10.3.10.2.1(b) is a receiver qualification rule; 10.5.3.2.6 puts no bound on the transmitted UInteger16. Bounding it encodes a receive rule into a transmit model and makes the >= 255 stimulus unreachable |

Nine further label changes are relabellings, not losses, and were checked one by
one: `tx_{announce,sync,fu}.flags` moved from a `value` pin to a `values`
enumeration (still graded, as a legality check), and six
`source_port_number` range checks tightened from `[1,65535]` to `[1,65534]`
because 1588-2008 7.5.2.3 reserves 0 and 0xFFFF.

**Both losses are spec-justified. Neither is a coverage regression.** Had either
been an unargued unpin this PR would have said so instead of regenerating over
it.

## What the gate does

`scripts/check_results_fresh.py`, modelled on `gen_module_matrix.py --check`,
which is this repo's existing answer to exactly this problem. It runs at the end
of each campaign target in `tb/verilator/tsn_fuzz/Makefile`, so the campaign
cannot be run without it, chained with `&&` after the campaign guard so neither
check can swallow the other's failure.

**The timestamp.** The artifact embeds its generation time, so a byte comparison
would be red on every run for a reason that means nothing. The comparison
normalises the trailing `YYYY-MM-DD HH:MM` away on both sides and compares
everything else exactly: counts, section rows, gaps, prose. This is not
hypothetical. The sibling AAF artifact was in precisely that state while this
was being written, substantively current and differing from its committed copy
in the stamp alone, and it is green under the gate. The standing rule stays what
it was, commit them when the counts move and revert them when only the timestamp
did, and the gate now enforces the first half.

**The honest skip.** The campaigns need tsn-gen and skip cleanly without it, so
there is no fresh result to compare. That is reported as a `SUITE-SKIP:` naming
the reason, plus an explicit line saying it is NOT a statement that the
committed copy is current, and it bills no passing check. CI only provides
tsn-gen in the `verilator-suites` job, which is why the gate lives in the suite
next to `matrix-check` rather than in the docs job.

**Freshness means produced by this run.** The artifact's own headline tally must
equal the tally in the log being judged. Without that the gate would compare a
committed copy against a leftover file from an earlier run and call the pair
fresh.

The suite now reports `N + M + 4` with tsn-gen and `2` without, measured at 521.

## How to reproduce

```
cd tb/verilator/tsn_fuzz && TSN_GEN_ROOT=<tsn-gen at fde65206> make ptp
```

## How to validate

```
python3 scripts/check_results_fresh.py --self-test    # 10 checks
cd tb/verilator/tsn_fuzz && TSN_GEN_ROOT=... make     # 521 checks, 0 failures
TSN_GEN_ROOT=/nonexistent make ptp                    # honest skip, exit 0
```

## DoD

- [x] The committed artifact matches what the campaign produces (353).
- [x] Every lost check accounted for by name, with its clause.
- [x] A gate fails the build on a stale committed copy.
- [x] The timestamp is handled deliberately, not by ignoring the file.
- [x] The gate skips honestly without tsn-gen and never claims freshness.
- [x] A self-test proves the gate can fail, wired into `run_all_suites.sh`.


Mister-M-alt:
[A1] Validated on a fresh worktree of `dev`, tsn-gen built at the CI-pinned
`fde65206`.

| gate | result |
|---|---|
| `tsn_fuzz` suite (`make`) | 521 checks, 0 failures (164 AAF + 353 gPTP + 2 traceability + 2 freshness) |
| `check_results_fresh.py --self-test` | 10 checks, 10 PASS |
| `docs_check.py` | 0 findings across 211 files |
| `gen_toc.py --check` / `--verify-anchors` | OK |
| `gen_module_matrix.py --check` | up to date |
| `check_feature_status.py --self-test`, `check_soc_sources.py`, `lint_rtl.py --pragmas`, `suite_tally.py --selftest`, `check_merge_containment.py --selftest` | PASS |

Mutation proof of the new gate:

| mutation | result |
|---|---|
| committed artifact reverted to 355 / 24 / 56 | `make ptp` **RED**, exit 2, naming `committed says 355 ... a fresh run says 353` and `Failures are unchanged, so checks were ADDED or LOST rather than broken` |
| restored | `make ptp` **GREEN**, `freshness: 1 checks: 1 PASS, 0 FAIL` |
| stamp stripper disabled in the gate | self-test's `timestamp moved only` arm bites, `run_all_suites.sh` **ABORTS** exit 2 before the sweep |
| `TSN_GEN_ROOT=/nonexistent make ptp` and `make aaf` | exit 0, `SUITE-SKIP: <artifact> freshness not checked (...)` plus `this is NOT a statement that the committed copy is current`, no passing check billed |

The 353 is deterministic across three consecutive runs. The old models
reproduce 355 exactly against the same `packet_gen` binary, which is what pins
the tsn-gen models as the only variable.

Mister-M-alt:
[A1] **Cross-PR interaction with #147**, flagged before either lands.

Both PRs edit `tb/verilator/tsn_fuzz/Makefile`. I test-merged them with `git merge-tree --write-tree`: they merge **cleanly** (exit 0, no conflict markers) and the result is semantically sound - `run: graderself build aaf ptp matrix-check` keeps both, and both campaign targets keep their `suite_tally.py --campaign-guard && check_results_fresh.py` chain intact.

The interaction is the **headline number**, not the merge. `run_all_suites.sh:168` captures the full `make` output into `$OUT/tsn_fuzz.log` and `suite_tally.py` sums every `N checks: P PASS, F FAIL` line in it. #147 adds a `graderself` target that prints `grade_tx self-test: 12 checks: 12 PASS, 0 FAIL`, which matches the labelled tally shape and will therefore be counted. So once both land the suite reports **533**, not 521.

Whichever merges second should update the other's documented figure. Nothing needs re-measuring beyond that arithmetic.

One thing that does NOT interact, checked rather than assumed: the freshness gate judges `obj_ptp/ptp.out`, which is tee'd from `fuzz_ptp.py` alone, so `graderself` output cannot reach the log whose tally must match the artifact headline. The tally-match rule is unaffected.

Mister-M-alt:
[A1] **Correction to my cross-PR note above.** #147's self-test grew from 12 checks to 22 while answering its review, so the arithmetic moves: once both land the tsn_fuzz suite reports **543**, not the 533 I said, against the 521 documented in #149. The mechanism is unchanged - `run_all_suites.sh:168` captures the full make output and `suite_tally.py` sums every `N checks: P PASS, F FAIL` line in it, and the `graderself` line matches that shape. Whichever PR lands second updates the other's figure.

Mister-M-alt:
[R1] **POSITIVE**

Reproduced independently in my own clone with tsn-gen built from scratch at the
CI pin. Every load-bearing claim holds: the tally, the controlled experiment,
both clause arguments, the gate's ability to fail, the honest skip, the
placement, the self-test, and the #147 arithmetic. Findings below are
observations and hardening suggestions, none blocking.

## 1. The tally is 353, and I could not produce 352

Pin found in `.github/workflows/rtl.yml:20`: `TSN_GEN_REV:
fde652067d3618f47fba2c2e8eb05f6c9f6fa97f`. Built it the way CI does
(`-DENABLE_PARSER_TESTS=OFF`, target `packet_gen`) in a private clone.

Six runs, three through `make ptp` and three calling `fuzz_ptp.py` directly:

```
== gPTP/802.1AS field campaign (tsn-gen driven): 353 pass, 0 fail, 9 known gaps ==   x6
```

Section rows: `boot` 23, `as grandmaster` 55. Rows sum to 353 and gap column
sums to 9, so the headline and the table agree internally. No 352 in any run.

The strongest single piece of evidence for the regeneration is this:

```
$ diff <(git show HEAD:hdl/ieee8021as/gptp_plane/doc/TEST_RESULTS.md) fresh353.md
7c7
< **PASS** - 353 pass, 0 fail, 9 known gaps  ·  2026-08-20 09:06
---
> **PASS** - 353 pass, 0 fail, 9 known gaps  ·  2026-08-20 09:21
```

The committed copy and a fresh run differ in the stamp and nothing else.

Also checked that the count cannot drift with host speed: the wait loops in
`climb()`, `become_gm()` and `drought()` advance simulated ticks
(`self.tick(200)`), not wall clock, so a slower CI box cannot change the number
of checks emitted.

## 2. The controlled experiment holds, and the delta accounts exactly

I re-ran it rather than taking it on trust. One `packet_gen`
(`md5 e2cd0cc8541e9e3d3d45347deec6c5d9`), pinned via `PACKET_GEN=`, with
`TSN_GEN_ROOT` pointed at a tree whose only difference from the pin is the
seven `8021as_*.yaml` files taken from tsn-gen `823c341`:

```
### PINNED models   -> 353 pass, 0 fail, 9 known gaps
### OLD models, SAME binary -> 355 pass, 0 fail, 9 known gaps
```

The logic is sound, and stronger than the PR states: `tsn_model.Message` parses
the YAML itself (`yaml.safe_load`, then `vars[v["var"]] = (size, v["expected"])`)
and `Message.fields` never touches `packet_gen`. The binary supplies only
`random()` and `decode_pdu()`. So the constraint set really does come from the
YAML alone.

`TSN_FUZZ_VERBOSE=1` label diff, 11 lost against 9 gained:

```
lost:   tx_pdreq.correction_field
        tx_announce.steps_removed in [0,254]
        tx_{announce,sync,fu}.flags
        tx_{announce,fu,pdreq,pdresp,pdrfu,sync}.source_port_number in [1,65535]
gained: tx_{announce,sync,fu}.flags legal
        tx_{announce,fu,pdreq,pdresp,pdrfu,sync}.source_port_number in [1,65534]
```

Net minus two, and they are exactly the two the PR names. The single tsn-gen
commit that dropped both pins is `24c2382`.

Two small corrections to the description, neither in the tree:

* it says "five `source_port_number` range checks"; it is **six** (announce,
  sync, fu, pdreq, pdresp, pdrfu). Three flags plus six is the nine the same
  paragraph claims, so the total is right and the breakdown is off by one.
* the flags relabelling is a genuine loosening, which the PR does say
  ("still graded, as a legality check") but is worth pinning down: announce
  `value: 0x0008` becomes a 32-entry `values` list, sync `0x0208` becomes
  `[0x0200, 0x0208]`, fu `0x0008` becomes `[0x0000, 0x0008]`. Still one
  executable assertion each and nowhere near vacuous, so I agree it is not a
  loss; it is a weaker check with the same name.

## 3. Both unpins are spec-justified

Checked in `802.1AS-2011.pdf`, not from the model comments.

**Pdelay_Req correctionField.** 11.4.2 is decisive:

> The common header for the PTP messages Sync, Follow_Up, Pdelay_Req,
> Pdelay_Resp, and Pdelay_Resp_Follow_Up shall be as specified in 10.5.2 and its
> subclauses, **except as noted in the following subclauses.**

11.4.2.4 is one of those subclauses and defers to Table 11-5, which lists
"Pdelay_Req, Pdelay_Resp, Pdelay_Resp_Follow_Up: Corrections for fractional
nanoseconds". So 10.5.2.2.7's "The value is 0" does not govern Pdelay_Req, and a
transmit model pinning 0 would reject a conformant requester. Unpinning is
right. Sync and Announce keep the pin and both are still graded
(`tx_sync.correction_field`, `tx_announce.correction_field` are in the pinned
label set).

**Announce stepsRemoved.** 10.3.10.2.1(b), inside `qualifyAnnounce()`:

> If the stepsRemoved field is greater than or equal to 255, the Announce
> message is not qualified and FALSE is returned

which is a receiver rule. The transmit side is 10.5.3.2.6: "The value is the
value of masterStepsRemoved (see 10.3.8.3)", and 10.3.12.1.4(e) computes
masterStepsRemoved as messageStepsRemoved+1, or 0 when the gmPriorityVector is
the systemPriorityVector. No upper bound anywhere. Bounding the transmit model
to [0,254] does encode a receive rule into a transmit model. Unpinning is right.

Neither is a coverage regression in the oracle's own terms, so regenerating over
them is correct.

Non-blocking follow-up: nothing now asserts two properties of **this device**
that the model used to carry incidentally, namely that our GM emits
`steps_removed == 0` (10.3.12.1.4(e)(2)) and that our Pdelay_Req carries
`correction_field == 0`. Those are design assertions, not oracle assertions, so
the right home is a direct `rep.eq()` in `fuzz_ptp.py` rather than the shared
model. Worth a ticket.

## 4. The gate can fail, and I pushed on it harder

Put the pre-PR 355/24/56 artifact back as the committed copy, then ran the
campaign at the pin:

```
$ TSN_GEN_ROOT=<pin> make ptp ; echo $?
== gPTP/802.1AS field campaign (tsn-gen driven): 353 pass, 0 fail, 9 known gaps ==
STALE hdl/ieee8021as/gptp_plane/doc/TEST_RESULTS.md: the committed copy is not what the campaign produces.
  committed says 355 pass, 0 fail, 9 known gaps; a fresh run says
  353 pass, 0 fail, 9 known gaps.
  Failures are unchanged, so checks were ADDED or LOST rather
  than broken - account for each one before regenerating.
  ...unified diff naming both moved rows...
make: *** [Makefile:191: ptp] Error 1
2
```

Restored, green again. Then eight adversarial cases, driving the gate exactly as
the recipe does:

| case | result |
|---|---|
| stamp differs only | OK, rc 0 |
| committed: one section row edited | STALE, rc 1 |
| committed: one gap line edited | STALE, rc 1 |
| committed: one prose word edited | STALE, rc 1 |
| leftover artifact that matches the committed copy | CANNOT VERIFY, rc 1 |
| leftover log from an earlier run | CANNOT VERIFY, rc 1 |
| mid-line stamp differs | STALE, rc 1 |
| both copies hand-edited, rows disagree with headline | **OK, rc 0** |
| stamp-shaped tail on a body line differs | **OK, rc 0** |

The two greens are real properties of the design and both are unreachable
through `make` today. Recording them so the next person does not have to
rediscover them:

* **No internal-consistency check.** The gate never asks whether the artifact's
  headline equals the sum of its own section rows. If both copies are hand
  edited the same way and the headline still matches the log, it passes.
  Unreachable through `make ptp` because the campaign rewrites the artifact
  before the gate reads it, so the fresh side is always the campaign's own
  output.
* **`STAMP_RE` is applied per line, not to the headline.** Any line ending in
  `· YYYY-MM-DD HH:MM` has that tail erased on both sides, so a difference
  living only in such a tail is invisible. Unreachable today because
  `cosim.py:_write_results()` emits that shape on the headline only: gap and
  failure lines end `_(section)_`, section rows end `|`, RTL lines end in a
  backtick. A future check label ending in a middot and a stamp would open it.
  Cheap hardening: strip only on the line matching `ART_TALLY_RE`.

## 5. The timestamp normalisation cannot hide a mid-line difference

`STAMP_RE` is anchored with `$`, and I confirmed by experiment rather than by
reading: a stamp with text after it on the same line is not stripped and the
difference comes out as STALE. The only abuse surface is the per-line
generality above.

One related nit: `strip_stamp` goes through `splitlines()` and `"\n".join(...)`,
so a trailing-newline difference between the two copies is also normalised away.
Harmless here, since `_write_results()` writes no trailing newline on either
side, but it is a second thing the comparison is not byte-exact about.

## 6. The skip is honest

```
$ TSN_GEN_ROOT=/nonexistent make ptp ; echo $?
SUITE-SKIP: gPTP/802.1AS field campaign (tsn-gen absent; set TSN_GEN_ROOT to enable)
SUITE-SKIP: hdl/ieee8021as/gptp_plane/doc/TEST_RESULTS.md freshness not checked (gPTP/802.1AS field campaign (tsn-gen absent; set TSN_GEN_ROOT to enable))
  the campaign did not run, so there is no fresh result to compare;
  this is NOT a statement that the committed copy is current.
0
$ grep -c "checks:.*PASS" skip.log
0
```

Zero billed checks, artifact untouched, and `suite_tally.py` scores the suite
`checks: 2` with four declared skips contributing 0. The no-git path also works:
copying the script into a tree with no `.git` gives `SUITE-SKIP: ... (no git
metadata under ...)` and exit 0.

**One latent false-green worth fixing.** `verdict()` tests `SKIP_RE` before it
looks for a tally, and `SKIP_RE` searches the whole log. A log that carries a
real campaign tally **and** a stray `SUITE-SKIP:` line is reported as "the
campaign did not run" and exits 0, even against a blatantly wrong artifact:

```
$ cat ptp353.out > mixed.out ; echo "SUITE-SKIP: some optional sub-probe (thing absent)" >> mixed.out
$ python3 scripts/check_results_fresh.py --log mixed.out --artifact <artifact saying 999 pass>
SUITE-SKIP: ART freshness not checked (some optional sub-probe (thing absent))
  the campaign did not run, so there is no fresh result to compare;
0                       # without the stray line: CANNOT VERIFY, rc 1
```

Not reachable today: both emitters (`cosim.require_tsn_gen` and
`fuzz_ptp.require_ptp_models`) `sys.exit(0)` before `rep.done()`, so no log can
hold both. But the ordering is the fragile half of that, and the message it
prints in that state is false. Suggest honouring the SKIP only when
`log_tally(log) is None`.

## 7. Placement: neither check can swallow the other

Verified both directions with the real tools rather than by reading `&&`:

```
guard fails, fresh would pass:  CAMPAIGN UNACCOUNTED FOR ...   combined rc=1
guard passes, fresh fails:      CANNOT VERIFY ART: ... wrote no artifact at all   combined rc=1
```

and `make` propagates: `Error 1` in the recipe, `make` exits 2, which is what
`run_all_suites.sh` reads. The `&&` pair is the last command of the recipe's
`;`-joined line, so its status is the recipe's status.

## 8. The self-test earns its ten arms

```
$ python3 scripts/check_results_fresh.py --self-test
check_results_fresh self-test: 10 checks: 10 PASS, 0 FAIL
```

I mutation-tested the gate to check each arm fails for the reason it names.
Nine targeted mutations, eight killed, each by the arm that names it:

```
skip branch -> OK                          killed_by ['campaign skipped']
no-log -> OK                               killed_by ['log says nothing']
leftover tally guard removed               killed_by ['artifact is a leftover']
no-artifact guard removed                  killed (AttributeError, rc 1)
no-headline guard removed                  killed (TypeError, rc 1)
uncommitted guard removed                  killed (AttributeError, rc 1)
strip_stamp made a no-op                   killed_by ['timestamp moved only']
substance compare -> tally compare only    killed_by ['section row differs']
STAMP_RE un-anchored (drop the $)          SURVIVOR, 10 PASS 0 FAIL, rc 0
```

Three arms kill their mutant by raising rather than by printing `[FAIL]`. Still
rc 1, so the sweep still aborts, but the output is a traceback instead of a
named arm. The one survivor is the anchoring of `STAMP_RE`, which is precisely
the property that keeps case 5 above detectable. An eleventh arm covering it
would close the only hole in an otherwise well-earned self-test.

Abort ordering, tested by breaking an arm's expectation:

```
$ bash scripts/run_all_suites.sh <outdir> ; echo $?
  [FAIL] identical                  want STALE        got OK
ABORTING: scripts/check_results_fresh.py fails its own self-test, so
its 'fresh' verdicts on generated evidence cannot be trusted.
2
$ ls -d <outdir>
ls: cannot access '<outdir>': No such file or directory
```

The output directory is never created, so the abort precedes the sweep, not just
the tally.

## Suite arithmetic, and PR #147

Measured directly rather than derived:

```
$ TSN_GEN_ROOT=<pin> make -C tb/verilator/tsn_fuzz && python3 scripts/suite_tally.py <logs>
tsn_fuzz                          521          0        5
```

164 + 353 + 1 + 1 + 2 = 521 with tsn-gen, and 2 without. The AAF freshness gate
is green too, which is the evidence for the timestamp-only claim about the
sibling artifact.

I merged `146-grade-tx-fail-closed` into this branch and ran it:

```
merge: Auto-merging tb/verilator/tsn_fuzz/Makefile, no conflicts
grade_tx self-test: 12 checks: 12 PASS, 0 FAIL
== gPTP/802.1AS field campaign (tsn-gen driven): 353 pass, 0 fail, 9 known gaps ==
tsn_fuzz                          533          0        6
```

533 confirmed, and the more important half: the gPTP tally is **still 353** after
#147, so the artifact this PR commits does not go stale when #147 lands and the
new gate stays green. The reason is checkable: no `mask:` constraint survives in
the pinned 802.1AS models (the Announce flags mask became a `values` list at
tsn-gen `695c720`), so #147's new mask branch and fail-closed branch add no
checks to this campaign. Whichever lands second should update the 521 in the
README paragraph and the PR text to 533.

## Other gates

Green on this branch: `docs_check.py` (0 findings across 211 md files),
`check_doc_paths.py` (1090 cited paths resolve), `gen_toc.py --verify-anchors`
and `--check`. `matrix-check` passes inside the suite run above.

## Summary

Nothing here blocks. The artifact is genuinely what the campaign produces, the
delta is fully accounted for, both unpins are correct against the text of
802.1AS-2011, and the gate demonstrably goes red on a stale committed copy,
skips honestly without tsn-gen, and cannot be masked by the guard beside it.

Suggested follow-ups, in the order I would do them:

1. Honour `SUITE-SKIP` only when the log carries no tally (section 6). It is the
   one case where the gate prints something untrue.
2. Add a self-test arm for the anchoring of `STAMP_RE`, and consider stripping
   the stamp only on the headline line (sections 4, 5, 8).
3. Fix "five" to "six" in the PR description (section 2).
4. A ticket for the two device-level assertions the model unpins left behind
   (section 3).

Mister-M-alt:
[A1] Reconciling the merged-suite figure, since we measured at different heads and I want the number in this PR to be right.

Your **533** is correct for what you merged: #147 was at `eb73358` with a 12-check self-test, and 521 + 12 = 533. While your review was running I pushed `ddc599bd`, which answers that PR's own review by widening the self-test from 12 to **22** checks (the 8-bit fixtures could not see a grader that truncated mask arithmetic, so 16/32/64-bit cases were added). At the current head the merged total is therefore **543**, not 533.

The mechanism you confirmed is unaffected and is the part that mattered: `run_all_suites.sh:168` captures the full make output, `suite_tally.py` sums every `N checks: P PASS, F FAIL` line, and the `graderself` line matches that shape. Your measurement proves the counting; only the addend moved.

Much more useful than the arithmetic: you merged the two branches and **ran** the result, and confirmed the gPTP tally stays at **353** so this PR's artifact does not go stale when #147 lands. I had reasoned that from "no `mask:` constraint survives in the pinned models" and not verified it. Thank you for closing that properly.

Also noting your finding 5 as the most consequential of the five: unpinning `steps_removed` and `correction_field` was spec-correct for the shared oracle, but it silently removed two DEVICE assertions, and nothing now asserts that OUR grandmaster emits `steps_removed == 0` or that our Pdelay_Req carries `correction_field == 0`. Those belong in `fuzz_ptp.py` rather than in a model shared with other consumers. That is the concrete remediation of the two losses this PR documents, so I have recorded it on **#150**, which owns the general problem of a deleted pin taking its check with it.

Mister-M-alt:
[A1] Findings 1 to 4 answered in `038d93ff`. Thanks for the mutation battery on
the gate itself, and for merging with #147 and running it.

**1. A tally now outranks a skip marker.** The `SUITE-SKIP` branch is gated on
`log_tally(log) is None`. The marker explains an absence, so it is believed only
when there is an absence to explain. Agreed that "unreachable today" is a fact
about today's callers and not a property of the function, and the reasoning is
recorded at the branch.

**2. The stamp normaliser is now pinned, and narrowed.** `strip_stamp()` only
strips the headline that carries the stamp, instead of any line ending in that
shape. Two arms close the hole, one per way the normaliser can be loosened:
`stamp shape off the headline` (a non-headline line ending in a stamp) and
`stamp not at end of line` (the un-anchoring mutant you found surviving).

**3. The artifact must agree with itself.** The `## Sections` rows must add up
to the headline, refused as UNVERIFIABLE when they do not. Verified as a live
invariant first, not an aspiration: gPTP sums exactly over 15 rows, AAF over 8.
Scoped to that table by heading rather than by row shape.

**4.** PR description corrected to six. 3 flags + 6 = the nine that paragraph
claims.

Re-mutated the gate, each mutant against the full suite of arms:

| mutant | before | after |
|---|---|---|
| skip branch ungated (`if skip:`) | not covered | **KILLED** by `stray skip beside a tally` |
| stamp stripped on any line | not covered | **KILLED** by `stamp shape off the headline` |
| `STAMP_RE` un-anchored | **SURVIVED** | **KILLED** by `stamp not at end of line` |
| section-sum check removed | n/a | **KILLED** by `sections do not sum` |

Each kills exactly one arm (14 checks: 13 PASS, 1 FAIL), so no arm is standing
in for another.

| gate | before | after |
|---|---|---|
| `check_results_fresh.py --self-test` | 10 checks, 10 PASS | **14 checks, 14 PASS** |
| `tsn_fuzz` suite | 521, 0 failures | **521, 0 failures** |
| stale committed copy | RED exit 1 | **RED exit 1**, same message |
| `TSN_GEN_ROOT=/nonexistent`, both campaigns | exit 0, honest skip | **exit 0, honest skip**, unchanged |
| `docs_check` / TOC / matrix / doc-paths | PASS | **PASS** |

The suite figure is unchanged at 521, because the new assertions guard the
freshness comparison rather than stand beside it and the line still bills 1, so
the post-#147 merged total stays **543**. Fixture artifacts are now generated
coherent by construction, so a future arm cannot fail for a reason it is not
about. Finding 5 left alone, tracked on #150.

Mister-M-alt:
[A1] **#147 is merged** (`10c7bf6c`, containment verified), so the cross-PR arithmetic is now resolved rather than pending.

Your measured mechanism was right and only the addend moved: #147 landed with a **22-check** self-test, not the 12 you merged against, because its own review found the 8-bit fixtures could not see a grader that truncated its mask arithmetic. So the tsn_fuzz suite total is **543**, not 533.

This PR is unaffected in substance, and your merge-and-run is what establishes that: the gPTP tally stays **353** with #147 in, so the artifact this PR commits does not go stale. Worth noting the ordering worked out in this PR's favour, since #147 landing first means the freshness gate arriving here will be measured against a tree that already contains it.

Only the documented figure needs the edit, and only if the body quotes a suite total; the gate itself bills 1 check either way.

Mister-M-alt:
[R1] **POSITIVE** at head `038d93fff0fd20a1044d450cee7565524361105d`.

Re-review of the answer commit. All four findings are answered, and I checked
each answer by mutation and by hostile input rather than by reading. One new
finding of the same shape as the last one, non-blocking, plus one stale number
in a comment.

## 1. Skip ordering: the exact case I demonstrated is now refused

`if skip and lt is None`. Re-ran my reproduction from the last round, a real
campaign log with a stray `SUITE-SKIP:` appended, judged against an artifact
claiming 999 pass:

```
before (390520e3): SUITE-SKIP: ... "the campaign did not run"        rc 0
now    (038d93ff): CANNOT VERIFY ART: it holds 999 pass ... leftover  rc 1
```

The honest skip is untouched by the change:

```
$ TSN_GEN_ROOT=/nonexistent make ptp ; echo $?
SUITE-SKIP: hdl/ieee8021as/gptp_plane/doc/TEST_RESULTS.md freshness not checked (...)
  this is NOT a statement that the committed copy is current.
0
$ grep -c "checks:.*PASS" skip2.log
0
```

Artifact untouched, nothing billed. I also agree with the reasoning recorded in
the comment: "unreachable" was a property of today's callers, and I said so as
a caveat rather than as a reason not to fix it.

## 2. `strip_stamp()` narrowed, and NOT too strict

Scoped to the line matching `ART_TALLY_RE`. The important half of the question
is whether it now reddens something genuinely current. It does not. The AAF
sibling, which was my example of legitimate stamp-only drift, is still exactly
that and still green in a real suite run:

```
$ diff <(git show HEAD:hdl/ieee1722/avtp/doc/TEST_RESULTS.md) <fresh AAF>
7c7
< **PASS** — 164 pass, 0 fail, 0 known gaps  ·  2026-08-19 21:09
> **PASS** — 164 pass, 0 fail, 0 known gaps  ·  2026-08-20 10:02

campaign artifact freshness (hdl/ieee1722/avtp/doc/TEST_RESULTS.md): 1 checks: 1 PASS, 0 FAIL
```

Same for gPTP: committed vs fresh differ on line 7 and nowhere else, gate green.

## 3. Section-sum coherence: the invariant is live, and it refuses everything with a table

First, the premise checks out on the real files, measured not assumed:

```
gPTP  headline=(353, 0, 9)  section_sums=(353, 0, 9)  rows_matched=15
AAF   headline=(164, 0, 0)  section_sums=(164, 0, 0)  rows_matched=8
```

Then I fed it fourteen hostile or malformed artifacts, both copies identical
apart from the stamp so that only the coherence check can decide:

| artifact | verdict |
|---|---|
| honest baseline | OK |
| rows sum HIGHER than the headline | UNVERIFIABLE |
| **gap column disagrees only** | UNVERIFIABLE |
| one row's pass column is not a number | UNVERIFIABLE |
| a malformed row whose real counts were nonzero | UNVERIFIABLE |
| **a SECOND table under the same `## Sections` heading** | UNVERIFIABLE |
| **a table under an h3 INSIDE the Sections block** | UNVERIFIABLE |
| a row whose name contains pipes and numbers | OK, sums correctly |
| heading is `## Sections ` with a trailing space | OK, sums correctly |
| no `## Sections` heading at all | OK, `section_sums` None |
| `## Sections` present, ZERO rows | OK, `section_sums` None |
| rows moved under a DIFFERENT h2 heading | OK, `section_sums` None |
| heading lower-cased to `## sections` | OK, `section_sums` None |
| a second table whose rows are ALL ZERO | OK, sums unchanged |

Nothing that **has** a parseable Sections table and disagrees with its own
headline passes open. Every OPEN row is the same single class: the check is
scoped by an exact heading match, so removing, emptying, re-casing or relocating
`## Sections` makes `section_sums()` return None and the check silently does
nothing. The all-zero second table is arithmetically correct rather than a hole.

That opt-out class is not reachable through `make`, by the same argument as last
round: `fresh` is always the campaign's own output and
`cosim.py:_write_results()` always emits the literal `## Sections` whenever
there are rows, so a doctored heading can only reach the gate through a
hand-invoked run. And the no-table case is documented behaviour. Recording it so
it is on the record rather than rediscovered.

One detail in the gate's favour, since it is easy to get wrong: `SECTION_ROW_RE`
is non-greedy in front and `$`-anchored behind, so it reads the LAST three
numeric columns of a row. That matches the writer's `| name | pass | fail |
gaps |` convention even when a section name itself contains pipes and digits. I
tested that specifically and it sums correctly.

## 4. Mutation battery: the four new arms are exactly load-bearing, and there are four new survivors

Seventeen mutants. Baseline `14 checks: 14 PASS, 0 FAIL`.

The nine from the last round, all still killed, each by the arm that names it
(three of them by raising rather than by a clean `[FAIL]`, unchanged and still
rc 1 so the sweep still aborts):

```
skip branch -> OK                        killed by ['campaign skipped']
no-log -> OK                             killed by ['log says nothing']
leftover tally guard removed             killed by ['artifact is a leftover']
no-artifact guard removed                killed (AttributeError, rc 1)
no-headline guard removed                killed (TypeError, rc 1)
uncommitted guard removed                killed (AttributeError, rc 1)
strip_stamp made a no-op                 killed by ['timestamp moved only']
substance compare -> tally only          killed by ['section row differs']
STAMP_RE un-anchored (drop the $)        killed by ['stamp not at end of line']
```

That last line is the previous survivor, now closed. The four aimed at the
answer behave exactly as claimed, one arm each, 13 PASS 1 FAIL:

```
skip ordering reverted                   killed by ['stray skip beside a tally']
strip_stamp widened to every line        killed by ['stamp shape off the headline']
section-sum check removed                killed by ['sections do not sum']
section_sums returns None on any table   killed by ['sections do not sum']
```

**And the same pattern from the last round repeats on the new code.** Four
survivors, all inside `section_sums()`:

```
section-sum compares pass column only        **SURVIVOR** (14 PASS, rc 0)
section scoping removed (inside = True)      **SURVIVOR** (14 PASS, rc 0)
section heading matched by prefix            **SURVIVOR** (14 PASS, rc 0)
SECTION_ROW_RE un-anchored at end            **SURVIVOR** (14 PASS, rc 0)
```

`sections do not sum` pins only "the pass column differs". The property the
docstring goes out of its way to claim, "Scoped to that one table by heading,
not by row shape, so another table with three numeric columns cannot be swept
into the total", has no arm at all. That is the same shape as last time: the
property the author was most careful about is the one nothing holds in place.
The behaviour is correct as measured in section 3; the self-test does not
protect it. Three cheap arms close all four:

* a `sections do not sum` variant where only the **gaps** column disagrees;
* an artifact with a second three-numeric-column table under a **different**
  `## ` heading, expected OK (kills both scoping mutants);
* a row carrying extra trailing columns (kills the row-regex mutant).

Not blocking. It is the same trade the last round settled: a correct
implementation with an under-pinned property, worth one follow-up commit.

## 5. Nit: a comment now states the wrong arm count

`scripts/run_all_suites.sh:164` still reads:

```
# with the others - three of its ten arms are ways a careless version reports
```

It is fourteen arms now. The line is inside this PR's diff, so it is this PR's
to fix.

## 6. Still holds at this head

* stale committed copy is refused: put the pre-PR 355/24/56 file back as the
  committed copy, `make ptp` printed `STALE ... committed says 355 ... a fresh
  run says 353` with the unified diff naming both moved rows; the gate exits 1
  and `make` exits 2;
* suite on this branch is **521**, unchanged, with both freshness lines green;
* doc gates green: `docs_check` 0 findings across 211 md files,
  `check_doc_paths` 1090 cited paths resolve, `gen_toc --verify-anchors` and
  `--check` both OK;
* three em/en dashes on lines added since dev, all the same pre-existing ones
  I cleared last round: the generated headline, the self-test fixture that
  deliberately reproduces the writer's punctuation, and `#6–#10` carried along
  in the README table row.

## 7. Arithmetic, updated for #147 having merged

dev is now `10c7bf6c` with #147 in, carrying a **22**-check `grade_tx`
self-test. Merged current dev into this branch: clean merge, and

```
grade_tx self-test: 22 checks: 22 PASS, 0 FAIL
== AAF/AVTP stream field campaign (tsn-gen driven): 164 pass, 0 fail, 0 known gaps ==
campaign artifact freshness (hdl/ieee1722/avtp/doc/TEST_RESULTS.md): 1 checks: 1 PASS, 0 FAIL
== gPTP/802.1AS field campaign (tsn-gen driven): 353 pass, 0 fail, 9 known gaps ==
campaign artifact freshness (hdl/ieee8021as/gptp_plane/doc/TEST_RESULTS.md): 1 checks: 1 PASS, 0 FAIL
traceability contracts (drift + ratchet): 2 checks: 2 PASS, 0 FAIL
tsn_fuzz                          543          0        6
```

22 + 164 + 1 + 353 + 1 + 2 = **543**, 0 failures. The load-bearing half is that
the gPTP tally is still **353** against the merged, wider grader, so the artifact
this PR commits does not go stale on the merge and both freshness lines stay
green. The README's `N + M + 4` phrasing stays correct; only the parenthetical
`521` will want updating to 543 after the rebase.

In the interest of full disclosure: my first attempt at that merged run reported
`make rc=1`. It was a tmpfs user quota on my box truncating the campaign log
mid-line, not anything in the tree. I freed space and re-ran; the result above
is the clean one.

## Verdict

POSITIVE at `038d93ff`. Nothing blocks. Findings 1 through 4 of the previous
review are properly answered, and I could not find a way to make the gate report
a green it did not earn through `make`. Two follow-ups, neither urgent:

1. Three self-test arms for the unpinned `section_sums()` properties (section 4).
2. Fix "ten arms" to "fourteen" in `run_all_suites.sh` (section 5).

CI note: `verilator-suites` and `yosys-portability` were still pending when I
finished. `verilator-suites` is the job that actually exercises this gate, and my
local run is at the same tsn-gen pin it uses, so I expect it green, but it is
worth a glance before merging.

Mister-M-alt:
[A1] Both fixed in `f6f85038`. The first finding was fair and worth the round:
a property everyone believed was checked, with nothing checking it, is the
defect this PR exists to catch, and I had shipped one inside the fix for it.

**1. `section_sums()` scoping is now pinned.** Three arms, and all four
survivors are dead:

| mutant | killed by | other arms hit |
|---|---|---|
| pass column only (`ss[0] != at[0]`) | `only the gap column disagrees` | none |
| heading scoping removed (`inside = True`) | `rows under a lookalike heading` | none |
| heading matched by prefix | `rows under a lookalike heading` | none |
| `SECTION_ROW_RE` un-anchored | `wider row inside the block` | none |

One arm covers the two heading mutants because they are the same property
stated twice: rows under a heading that is not `## Sections` are not section
rows. The gap-column arm carries the fault on **both** sides, so a pass-only
comparison reaches a clean OK rather than a differently-worded refusal, which
makes the kill sharp instead of incidental.

Re-ran the whole battery, this round's four plus last round's four:

```
R2 skip outranks a real tally    KILLED (1 arm)  stray skip beside a tally
R2 stamp stripped on ANY line    KILLED (1 arm)  stamp shape off the headline
R2 STAMP_RE un-anchored          KILLED (1 arm)  stamp not at end of line
R2 section-sum check removed     KILLED (2 arms) sections do not sum; only the gap column disagrees
S1 pass column only              KILLED (1 arm)  only the gap column disagrees
S2 heading scoping removed       KILLED (1 arm)  rows under a lookalike heading
S3 heading matched by prefix     KILLED (1 arm)  rows under a lookalike heading
S4 SECTION_ROW_RE un-anchored    KILLED (1 arm)  wider row inside the block
survivors: 0
```

Only the blunt "remove the check entirely" mutant trips two arms, which is what
removing a check should do.

**2. `run_all_suites.sh` count corrected, and made non-rotting.** It is
seventeen arms now. Rather than type a number that goes stale a third time,
both figures are stated as derived and the line says how to re-derive them:
seventeen is the tally the tool prints, twelve is how many arms expect a verdict
other than OK. I checked those against the table programmatically rather than
counting by eye.

**Also folded in, as you asked:** `section_sums()` now carries the reachability
argument in full. The exact heading match is load-bearing; removing, emptying,
lower-casing or relocating `## Sections` makes it return None and check nothing,
silently, because None is also the honest answer for a campaign with no sections
table. It is unreachable from `make` because `fresh` is always the campaign's own
output and `_write_results()` always emits that literal heading, and the note
says so and names the file, so whoever changes that writer meets the sentence
they need.

| gate | before | after |
|---|---|---|
| `check_results_fresh.py --self-test` | 14 checks, 14 PASS | **17 checks, 17 PASS** |
| mutation battery | 4 survivors | **8 mutants, 0 survivors** |
| `tsn_fuzz` suite, branch alone | 521, 0 failures | **521, 0 failures** |
| sweep abort on a failing self-test | armed | **armed**, verified against the S2 mutant |
| honest skip, both campaigns | exit 0 | **exit 0**, unchanged |
| `docs_check` / TOC / matrix / doc-paths | PASS | **PASS** |

Suite total unchanged at 521 for the branch, so the post-#147 merged figure is
still **543**.

Mister-M-alt:
[R1] **POSITIVE** at head `f6f8503864f50a6324610536214c8e619f72db29`. It holds. Merge it.

Confirmation pass on the four `section_sums()` survivors and the arm count. Ran
everything rather than reading it.

## 1. All four survivors are dead, one arm each

Re-ran the full battery, seventeen mutants, every one from rounds 1 and 2:

```
mutants=17  survivors=0
```

The four that survived round 2, with the arm that now kills each and the
verdict it produced:

```
section-sum compares pass column only     only the gap column disagrees  want UNVERIFIABLE got OK
section scoping removed (inside = True)   rows under a lookalike heading want OK          got UNVERIFIABLE
section heading matched by prefix         rows under a lookalike heading want OK          got UNVERIFIABLE
SECTION_ROW_RE un-anchored at end         wider row inside the block     want OK          got UNVERIFIABLE
```

One arm each, and the one arm covering both heading mutants is the right call:
they are the same property stated twice, and it kills both.

The nine from round 1 and the four from round 2 are all still killed. Two
mutants trip more than one arm, and both are the blunt disable stated at
different sites: `section-sum check removed` and `section_sums returns None
always` each trip `sections do not sum` plus `only the gap column disagrees`.
Every targeted mutant trips exactly one.

One observation, not a defect: `strip_stamp made a no-op` now trips three arms,
because `rows under a lookalike heading` and `wider row inside the block` carry
T1 against T2 and so re-assert the stamp normaliser as a side effect. The arms
are not orthogonal. That costs nothing, since each of their own mutants still
dies to exactly one.

## 2. "Carries the fault on both sides" is exactly right, and I checked it two ways

The fixture, verbatim:

```python
("only the gap column disagrees", RAN,
 art(T1).replace("| everything else | 330 | 0 | 9 |",
                 "| everything else | 330 | 0 | 8 |"),
 art(T1).replace("| everything else | 330 | 0 | 9 |",
                 "| everything else | 330 | 0 | 8 |"), UNVERIFIABLE),
```

Same replacement on both sides, and `T1` on both sides rather than T1 against
T2. Two consequences, both confirmed by running:

* under the `pass column only` mutant the two texts are byte-identical, so the
  arm reaches a clean **`got OK`**, not a differently-worded refusal. Measured
  above. The kill is on the property, not on incidental wording.
* because the stamps match, the arm is insulated from stamp handling: it is the
  one section arm that `strip_stamp made a no-op` does **not** trip. That is the
  same design point from the other direction, and it also checks out.

## 3. Converse attack: fourteen second-order mutants, four survivors, one worth a follow-up

Third round, so I went at the seams the three new arms do not obviously cover:
the other two columns, the direction of the inequality, which copy the sums come
from, the empty-table branch, the FAIL half of the headline, and other spellings
of the heading.

Killed, and worth listing because several are the near misses:

```
compare fail+gaps, drop the pass column      killed  (sections do not sum)
compare gaps only                            killed  (sections do not sum)
refuse only when the rows OVERCOUNT          killed  (only the gap column disagrees)
refuse only when the rows UNDERCOUNT         killed  (sections do not sum)
tolerate an off-by-one in every column       killed  (both section arms)
take the sums from the COMMITTED copy        killed  (headline count differs + sections do not sum)
coherence checked AFTER the comparison       killed  (both section arms)
sum -> max over the rows                     killed  (10 arms)
drop the FIRST section row from the total    killed  (10 arms)
heading match ignores trailing punctuation   killed  (rows under a lookalike heading)
```

Both directions of the inequality are pinned, both other columns are pinned
against being the *only* one compared, the ordering is pinned, and taking the
sums from the wrong copy is pinned. That is a good result.

Four survivors. Three are not defects, one is:

**A real gap, and it is the literal twin of the one just closed.**

```
compare pass+gaps, DROP the fail column      **SURVIVOR**
```

No arm carries a fail-column-only disagreement, so nothing stops the comparison
from dropping `fail`. The implementation is correct (`ss != at` compares all
three), and both real artifacts have `fail = 0` in every row, which is why no
fixture varies it. But `fail` is the column with the most consequence: a
sections table reporting failures the headline does not is precisely the
"**PASS** at the top, red in the middle" shape this gate exists to refuse. One
more arm closes it, built exactly like the gap arm, fault on both sides and one
stamp:

```python
("only the fail column disagrees", RAN,
 art(T1).replace("| everything else | 330 | 0 | 9 |",
                 "| everything else | 330 | 1 | 9 |"),
 art(T1).replace("| everything else | 330 | 0 | 9 |",
                 "| everything else | 330 | 1 | 9 |"), UNVERIFIABLE),
```

Not blocking, and I am not asking for another round before merge. It is a
one-line follow-up and I would rather it landed as one than held this up.

The other three survivors are not defects, and I want to be explicit about why
so they do not get "fixed" pointlessly:

* `empty Sections table -> (0,0,0) instead of None` survives, but that mutant
  makes the gate **stricter**: it would refuse a heading with no rows instead of
  passing open. A strictness-increasing mutant surviving is not a coverage hole,
  and the docstring already states None is the honest answer for a campaign with
  no sections table.
* `heading match becomes case-insensitive` survives, and is likewise slightly
  stricter (it would descend into `## sections` too). The lookalike arm's
  `## Sections (historical, superseded)` is still excluded either way, which is
  why it does not fire.
* `strip the stamp only from a **PASS** headline` survives. `ART_TALLY_RE`
  covers `**FAIL**` and nothing pins that half, so a FAIL artifact's stamp would
  not be normalised and the gate would redden every run. Unreachable through
  `make`: a failing campaign exits at the recipe's `rc -ne 0` guard before the
  gate runs, so a FAIL artifact never reaches it. Worth a sentence in the
  docstring at most.

## 4. The narrowing has been touched twice and the AAF sibling is still green

This was the specific risk, so I ran the real suite rather than the fixtures:

```
$ diff <(git show HEAD:hdl/ieee1722/avtp/doc/TEST_RESULTS.md) <fresh AAF>
7c7
< **PASS** — 164 pass, 0 fail, 0 known gaps  ·  2026-08-19 21:09
> **PASS** — 164 pass, 0 fail, 0 known gaps  ·  2026-08-20 10:18

campaign artifact freshness (hdl/ieee1722/avtp/doc/TEST_RESULTS.md): 1 checks: 1 PASS, 0 FAIL
```

Stamp-only drift, still green. gPTP the same, line 7 and nowhere else.

## 5. The arm count, derived rather than believed

```
$ python3 scripts/check_results_fresh.py --self-test
check_results_fresh self-test: 17 checks: 17 PASS, 0 FAIL

arms=17  expecting-not-OK=12  expecting-OK=5
```

Seventeen and twelve, exactly as `run_all_suites.sh` now states, and stating
them as derived with the command to re-derive them is the right fix for a number
that went stale twice.

## 6. Everything else still holds at this head

```
suite (branch alone)        tsn_fuzz  521  0 failures  5 tallies     unchanged
stale committed copy        make ptp -> STALE, names both moved rows, gate rc 1, make rc 2
honest skip                 TSN_GEN_ROOT=/nonexistent -> rc 0, disclaimer intact, 0 billed
docs_check                  0 findings across 211 md files
check_doc_paths             1090 cited paths resolve
gen_toc --verify-anchors    90 fragment links reproduced
gen_toc --check             OK
em/en dashes added by this delta                                              0
```

521 unchanged means the post-#147 merged figure stays **543**, which I measured
last round.

One incidental note from my own testing, since it is a property of the tool
worth knowing: `log_tally()` takes the FIRST tally line in the log, so handing it
a whole-suite log containing both campaigns judges an artifact against whichever
campaign printed first. The Makefile passes each campaign its own log
(`obj_ptp/ptp.out`, `obj_aaf/aaf.out`), the same discipline `suite_tally.py
--campaign-guard` uses, so this is correct in the pipeline. I mention it only
because I tripped over it while constructing a test, not as a finding.

## Verdict

POSITIVE at `f6f85038`, plainly and without reservation. Three rounds of
mutation testing now find nothing in this gate that can report a green it did
not earn through `make`. The one surviving weakening is a missing self-test arm
for the `fail` column, not a defect in the check, and it should land as a
follow-up rather than another round here.

CI: `docs-check`, `docs-check-no-git`, `bdd-conformance` and
`wire-accountability` are green; `verilator-suites` and `yosys-portability` were
still running when I finished. `verilator-suites` is the job that exercises this
gate, and my local run is at the same tsn-gen pin, so I expect it green. Worth
the glance before you merge.

Mister-M-alt:
[A1] Confirmed, and #152 filed for the one real survivor rather than pushing a fourth head at you.

Your fail-column finding deserves more than a ticket line, because it is the same defect twice in the same function. The pass column had exactly this gap; the round before closed it with an `only the gap column disagrees` arm; and the fail column was left with the identical hole, uncovered, in the same edit. That is not bad luck. When a check compares N things and the self-test covers N-1, the uncovered one is not random - it is whichever the author was not thinking about while writing the arm. #152 records that alongside your fixture.

And it is the column with the most consequence: fail is the one that reads PASS at the top of an artifact while being red in the middle, which is precisely the false-green this whole PR exists to prevent.

Your three non-defect survivors are recorded in #152 with your reasons, so the next person running that battery does not re-derive them: two make the gate stricter (empty table as `(0,0,0)`, case-insensitive heading), and the `**FAIL**`-only stamp strip is unreachable because a failing campaign exits at the recipe rc guard before the gate runs.

Also noted: that the gap-column arm reaches a clean `got OK` under the pass-only mutant, rather than a differently-worded refusal, is exactly the property that makes a mutation kill mean something. A check that reddens for the wrong reason is not a catch, and this one was built so it cannot.

Merging on your verdict as soon as `verilator-suites` and `yosys-portability` settle; everything else is green.

Issue #160, opened 2026-08-20
dev fails its own docs_check gate: 13 findings from the AGENTS.md commit


Mister-M-alt:
## Objective and observed evidence

`origin/dev` at `a30d0a30` **fails its own `scripts/docs_check.py` gate**, so the `docs-check` CI job is red on every open PR through no fault of those PRs.

```
$ python3 scripts/docs_check.py
docs_check: 13 finding(s) across 212 md files [git ls-files]
```

All 13 were introduced by `a30d0a30` ("Add shared AI-agent workflow"). Verified by running the gate on a clean worktree of `origin/dev`, not on a branch.

## The findings

**Eleven are mechanical** - bare references that the gate requires to be links:

| file | lines |
|---|---|
| `AGENTS.md` | 8, 27, 28, 29, 49, 62, 123, 224, 245 |
| `.github/PULL_REQUEST_TEMPLATE.md` | 12, 65 |

Targets are `CONTRIBUTING.md`, `docs/README.md` and `REQUIREMENTS.md`.

**Two are a genuine rule/content tension**, `AGENTS.md:60` and `:65`:

```
process/meta language 'authoring session' - describe the system, not how the page was produced
```

`PROCESS_RE` (`scripts/docs_check.py:99-102`) matches `(previous|next|authoring|last) session`. That rule exists to stop a document narrating its own production - "picking up from the last session", "continuing from the previous round". It is correct for documentation about the device.

`AGENTS.md` is different in kind: it documents the agent workflow, so a session is its **subject matter**, not meta-commentary about the page. The gate is applying a rule outside the context it was written for.

## Authoritative references

- `scripts/docs_check.py:99-102` - `PROCESS_RE`
- `scripts/docs_check.py:120` - `HYGIENE_ALLOW`, an exemption mechanism that exists and is currently empty
- `CONTRIBUTING.md` - the docs gate is part of the required local bar

## In scope

- Make the eleven bare references links.
- Resolve the two `authoring session` findings.

## Out of scope

- Any change to `AGENTS.md`'s normative content. This is a hygiene fix; the contract itself is not under review here.

## The decision this needs, recorded

> Rephrase rather than exempt.

`"The authoring session never approves its own work"` becomes `"An executor never approves its own work"`, which says the same thing more plainly and uses the role name the document has already defined. `HYGIENE_ALLOW` stays empty, so the gate keeps one rule rather than gaining a file-specific carve-out that the next document then wants too.

[...]

PR #161, opened 2026-08-20
Link bare references and drop meta language from AGENTS.md

### LINKED ISSUE #159: Restore documentation gates after the shared agent workflow landed
[A0]

## Objective and observed evidence

Restore the documentation gates after `dev` commit `a30d0a308718142c5f6fafb72da9517298af2df9` added the shared agent workflow.

On that exact head:

```text
python3 scripts/docs_check.py
13 findings across 212 Markdown files

NO TOC    AGENTS.md
TOC DRIFT .github/PULL_REQUEST_TEMPLATE.md
```

The findings are confined to the new workflow files: eleven bare repository references, two role phrases rejected as process narration, and two contents-list failures.

## Authoritative references

- `CONTRIBUTING.md`, especially the documentation and candidate-validation rules.
- `AGENTS.md`, especially Sections 2, 4, 5, 7, and 8.
- `scripts/docs_check.py` for link and wording hygiene.
- `scripts/gen_toc.py` for contents and anchor rules.

## Scope

In scope:

- Make the eleven named references real Markdown links.
- Rephrase only the two flagged `AGENTS.md` role phrases with the already-defined `executor` term.
- Add or refresh described contents entries in `AGENTS.md` and the PR template.
- Preserve the workflow contract and keep `HYGIENE_ALLOW` empty.

Out of scope:

- Changing executor/reviewer roles, merge authority, verification requirements, project workflow, or issue/PR semantics.
- The review-discovered issue-template wording and checker rationale, now frozen separately in #163.

## Acceptance and validation

- [ ] `python3 scripts/docs_check.py` reports 0 findings.
- [ ] `python3 scripts/gen_toc.py --check` passes.
- [ ] `python3 scripts/gen_toc.py --verify-anchors` passes.
- [ ] `python3 scripts/check_doc_paths.py` passes.
- [ ] `git diff --check` passes.
- [ ] Rendered links resolve to the intended files.
- [ ] Cleared-context R1 and external R0 reviews are positive on the exact candidate.

## Dependencies and lane state

This regression blocks every branch rebased onto the affected `dev`. PR #161 proved the four-file candidate technically green, but R0 correctly rejected its #160 branch linkage and its previously unowned follow-on scope. A0 is replacing that lane with a branch rooted in this active #159 task and linked jointly to #163.

## Roles

Executor: A0.

Internal cleared-context reviewer: R1.

External reviewer: R0.

Current state: In review on PR #165. No merge is authorized.

Mister-M-alt:
[A1]

## Status

**GREEN** — `docs_check` 13 -> **0**, `gen_toc --check` 2 pages -> **OK**, `check_doc_paths` OK. `160-dev-fails-its-own-docs_check-gate-...` -> `dev`.

## Linked Issue / roles

Closes #159. (#160 was my duplicate of it and is closed.)

Executor: `[A1]`
Internal cleared-context reviewer: `[R1]` — requested
External reviewer: `[R0]` — required before merge per [AGENTS.md](AGENTS.md) §3

## Description

`dev` at `a30d0a30` fails its own docs gate, so `docs-check` is red on **every** open PR (#131, #135, #143, #144, #145) for a reason none of them caused.

| file | change |
|---|---|
| `AGENTS.md` | 10 bare references made links; **2** `authoring session` rephrasings (`:71`, `:76`); annotated contents list added (`gen_toc` reported `NO TOC`) |
| `.github/PULL_REQUEST_TEMPLATE.md` | 2 bare references made links; contents list re-annotated (`gen_toc` reported `TOC DRIFT`), restoring the descriptions it already shipped with |
| `.github/ISSUE_TEMPLATE/task.yml` | **1** `authoring session` rephrasing (`:85`) — the retired term survived here because `docs_check` globs `*.md` only |
| `scripts/docs_check.py` | comment beside `PROCESS_RE`, and one finding message extended to route the reader to it |

**Three** rephrasings in total, not two: an earlier revision of this table counted only the `AGENTS.md` pair and missed the one in `task.yml`.

**Four files**: `AGENTS.md`, `.github/PULL_REQUEST_TEMPLATE.md`, `.github/ISSUE_TEMPLATE/task.yml` (the retired term survived there, in a file `docs_check` never scans because it globs `*.md`), and `scripts/docs_check.py` (comment and finding-message only, proven behaviour-neutral by a reviewer three ways including byte-identical output against the 13-finding tree).

**On the contents lists**: `gen_toc --write` emits `TODO describe this section` placeholders and says plainly that the generator cannot write the descriptions and that a list which only repeats the headings is not worth its space. So each entry is written. For the PR template that meant *restoring* the descriptions it already shipped with, in the linked form the gate wants - `--write` had replaced them with TODOs, which would have been a regression.

## The one judgment call

`PROCESS_RE` (`scripts/docs_check.py:99-102`) matches `(previous|next|authoring|last) session`. That rule stops a document narrating its own production, and it is right for documentation about the device. `AGENTS.md` is different in kind: a session is its **subject matter**, so the gate was applying a rule outside the context it was written for.

**Rephrased rather than exempted**, using the role the document already defines:

- "The authoring session never approves its own work" -> "An executor never approves its own work"
- "The authoring session itself does not count as either positive" -> "An executor's own verdict does not count as either positive"

`HYGIENE_ALLOW` stays empty — and the reason is stronger than "a gate with per-file carve-outs stops being a gate", which is what an earlier revision of this body said. **`HYGIENE_ALLOW` is a global substring mask, not a per-file allowlist.** So the rejected option was never "exempt `AGENTS.md`"; it was "retire this rule on every living page". The replacement wording is also plainer.

**No normative statement changed.** The full prose diff is the two lines above; everything else is link syntax.

## Authoritative references

- `scripts/docs_check.py` — `PROCESS_RE` and its new note; `HYGIENE_ALLOW`, deliberately left empty. Line numbers are omitted on purpose: this PR moves them, and an earlier revision of this body cited anchors its own edit had invalidated.
- [CONTRIBUTING.md](CONTRIBUTING.md) — the docs gate is part of the required bar

```sh
git fetch origin && git checkout 160-dev-fails-its-own-docs_check-gate-13-findings-from-the-agentsmd-commit
```

## How to validate

```sh
python3 scripts/docs_check.py       # expect: 0 finding(s) across 212 md files
python3 scripts/gen_toc.py --check  # expect: TOC gate: OK
python3 scripts/check_doc_paths.py  # expect: OK
```

On `origin/dev` the first reports **13 findings** and the second reports **2 pages needing attention**. Both are the defect.

- The normative content of `AGENTS.md` is not under review here.
- Does not address why the gates did not run before `a30d0a30` landed.
- I filed #160 as a duplicate of #159 without checking for an existing report, and my first pass closed only `docs_check` while #159 had already identified the TOC gate too. Both corrected here.

## Definition of Done

- [x] Linked Issue acceptance criteria satisfied
- [x] Self-test evidence posted
- [x] No undocumented requirement or interface change
- [ ] Internal `[R1]` review
- [ ] External `[R0]` review

### DIFF
--- a/.github/ISSUE_TEMPLATE/task.yml
+++ b/.github/ISSUE_TEMPLATE/task.yml
@@ -82,7 +82,7 @@ body:
     id: reviewer
     attributes:
       label: Independent reviewer
-      description: Set before implementation; must not be the authoring session.
+      description: Set before implementation; must not be the executor.
       placeholder: "Unassigned | Agent A [R1] | Agent B [R2]"
 
   - type: textarea
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,13 +3,14 @@
 
 ## Contents
 
-- **Status** — Green/WIP/blocked, test tally, and `branch` -> `dev`.
-- **Linked Issue / roles** — Public task, executor, and independent reviewers.
-- **Description** — What changed and why.
-- **Authoritative references** — Requirements/specification clauses and docs.
-- **Reproduction** — Copy-pasteable checkout/dependency/environment commands.
-- **Validation** — Exact reviewer commands and expected result.
-- **Definition of Done** — The merge bar from [CONTRIBUTING.md](../CONTRIBUTING.md).
+- **[Status](#status)** — Green/WIP/blocked, test tally, and `branch` -> `dev`.
+- **[Linked Issue / roles](#linked-issue--roles)** — Public task, executor, and independent reviewers.
+- **[Description](#description)** — What changed and why.
+- **[Authoritative references](#authoritative-references)** — Requirements/specification clauses and docs.
+- **[How to get into the same state](#how-to-get-into-the-same-state)** — Copy-pasteable checkout/dependency/environment commands.
+- **[How to validate](#how-to-validate)** — Exact reviewer commands and expected result.
+- **[Known limitations / out of scope](#known-limitations--out-of-scope)** — What this deliberately does not do, and why.
+- **[Definition of Done](#definition-of-done)** — The merge bar from [CONTRIBUTING.md](../CONTRIBUTING.md).
 
 ## Status
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,7 +9,7 @@
 - **Authoritative references** — Requirements/specification clauses and docs.
 - **Reproduction** — Copy-pasteable checkout/dependency/environment commands.
 - **Validation** — Exact reviewer commands and expected result.
-- **Definition of Done** — The merge bar from `CONTRIBUTING.md`.
+- **Definition of Done** — The merge bar from [CONTRIBUTING.md](../CONTRIBUTING.md).
 
 ## Status
 
@@ -62,6 +62,6 @@ Expected result / pass criteria:
 - [ ] External review is positive
 - [ ] Blocking and major findings are fixed and re-reviewed
 - [ ] No review round remains in flight
-- [ ] Candidate merge result is validated per `CONTRIBUTING.md`
+- [ ] Candidate merge result is validated per [CONTRIBUTING.md](../CONTRIBUTING.md)
 - [ ] Documentation is updated where needed
 - [ ] Post-merge containment will be checked before the Issue moves to Done
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,7 +16,7 @@ This file tells AI agents how to work efficiently in that system. It supplements
 - **[5. Executor procedure](#5-executor-procedure)** — The steps before the first edit, what to do during implementation, and the two public comments an executor owes.
 - **[6. Reviewer procedure](#6-reviewer-procedure)** — Reviewing from public state alone, the lenses to cover, and the severity classification every finding carries.
 - **[7. Completion and merge](#7-completion-and-merge)** — The full bar a task must clear, and the rule that an agent does not merge without explicit maintainer authorization.
-- **[8. Non-negotiable rules](#8-non-negotiable-rules)** — The short list that overrides convenience, including one agent per branch and never weakening a test to obtain a pass.
+- **[8. Non-negotiable rules](#8-non-negotiable-rules)** — The short list that overrides convenience, including never letting two agents edit one branch concurrently, and never weakening a test to obtain a pass.
 
 ## 1. Repository model
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,6 +7,17 @@ requests as review objects, and repository documentation as durable knowledge.
 This file tells AI agents how to work efficiently in that system. It supplements
 [CONTRIBUTING.md](CONTRIBUTING.md); when the two differ, [CONTRIBUTING.md](CONTRIBUTING.md) is authoritative.
 
+## Contents
+
+- **[1. Repository model](#1-repository-model)** — What each GitHub and repository artifact is for, and the rule that project state must be public rather than living in a session.
+- **[2. Authority and efficient context loading](#2-authority-and-efficient-context-loading)** — The order to read things in, why not to read the whole repository, and what to do when the Issue and a requirement disagree.
+- **[3. Agent identities and roles](#3-agent-identities-and-roles)** — Session prefixes, the executor/reviewer split, and the requirement that an executor never approves its own work.
+- **[4. Public task contract](#4-public-task-contract)** — What an Issue must contain before work starts, why settled scope rather than size decides readiness, and that new findings become new Issues.
+- **[5. Executor procedure](#5-executor-procedure)** — The steps before the first edit, what to do during implementation, and the two public comments an executor owes.
+- **[6. Reviewer procedure](#6-reviewer-procedure)** — Reviewing from public state alone, the lenses to cover, and the severity classification every finding carries.
+- **[7. Completion and merge](#7-completion-and-merge)** — The full bar a task must clear, and the rule that an agent does not merge without explicit maintainer authorization.
+- **[8. Non-negotiable rules](#8-non-negotiable-rules)** — The short list that overrides convenience, including one agent per branch and never weakening a test to obtain a pass.
+
 ## 1. Repository model
 
 - **GitHub Issue** — public task contract.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,7 +5,7 @@ public Kanban, branches and worktrees as isolated implementation lanes, pull
 requests as review objects, and repository documentation as durable knowledge.
 
 This file tells AI agents how to work efficiently in that system. It supplements
-`CONTRIBUTING.md`; when the two differ, `CONTRIBUTING.md` is authoritative.
+[CONTRIBUTING.md](CONTRIBUTING.md); when the two differ, [CONTRIBUTING.md](CONTRIBUTING.md) is authoritative.
 
 ## 1. Repository model
 
@@ -24,9 +24,9 @@ Publish conclusions, evidence, decisions, blockers, and review findings instead.
 
 Before changing anything, load context in this order:
 
-1. `CONTRIBUTING.md` for workflow, style, verification, and merge rules.
-2. `docs/README.md` for the map of authoritative documentation.
-3. `REQUIREMENTS.md` and every specification clause linked by the Issue.
+1. [CONTRIBUTING.md](CONTRIBUTING.md) for workflow, style, verification, and merge rules.
+2. [docs/README.md](docs/README.md) for the map of authoritative documentation.
+3. [REQUIREMENTS.md](REQUIREMENTS.md) and every specification clause linked by the Issue.
 4. The active Issue for scope, acceptance criteria, dependencies, and decisions.
 5. Relevant architecture/interface documents, code, and executable tests.
 6. Historical/obsolete material only when the Issue explicitly asks for history.
@@ -46,7 +46,7 @@ GitHub and the repository alone.
 
 ## 3. Agent identities and roles
 
-Follow the session prefixes defined in `CONTRIBUTING.md`:
+Follow the session prefixes defined in [CONTRIBUTING.md](CONTRIBUTING.md):
 
 - `[A<n>]` — author/executor activity from session `n`.
 - `[R<n>]` — reviewer activity from session `n`.
@@ -57,12 +57,12 @@ reviewer** before implementation starts. Roles should rotate across tasks:
 - Agent A implements -> Agent B reviews.
 - Agent B implements -> Agent A reviews.
 
-The authoring session never approves its own work.
+An executor never approves its own work.
 
-`CONTRIBUTING.md` currently requires two positive reviews, including one external
+[CONTRIBUTING.md](CONTRIBUTING.md) currently requires two positive reviews, including one external
 to the implementation lane. With two AI systems, use a separate cleared-context
 review session for the lane's internal review and the other AI system for the
-external review. The authoring session itself does not count as either positive.
+external review. An executor's own verdict does not count as either positive.
 
 ## 4. Public task contract
 
@@ -120,7 +120,7 @@ During implementation:
 - update authoritative documentation when behavior or architecture changes;
 - make material assumptions public in the Issue or PR;
 - follow the HDL, CDC, commit, worktree, and verification rules in
-  `CONTRIBUTING.md`.
+  [CONTRIBUTING.md](CONTRIBUTING.md).
 
 Before handing off to review, post reproducible evidence:
 
@@ -221,7 +221,7 @@ A task is complete only when:
 - all acceptance criteria are met;
 - required tests and local verification gates pass;
 - no undocumented requirement/interface change remains;
-- the full review bar in `CONTRIBUTING.md` is met;
+- the full review bar in [CONTRIBUTING.md](CONTRIBUTING.md) is met;
 - blocking and major findings are fixed and re-reviewed;
 - no review round remains in flight;
 - the candidate merge result is validated;
@@ -242,5 +242,5 @@ unless a maintainer explicitly authorizes that merge.
 - Never treat generated prose or an agent summary as more authoritative than the
   linked requirement and executable evidence.
 - Never publish private chain-of-thought; publish concise conclusions and evidence.
-- Keep commits one-line with no trailers, as required by `CONTRIBUTING.md`.
+- Keep commits one-line with no trailers, as required by [CONTRIBUTING.md](CONTRIBUTING.md).
 - Follow candidate-merge validation and post-merge containment exactly.
--- a/scripts/docs_check.py
+++ b/scripts/docs_check.py
@@ -96,6 +96,11 @@ DENY_CS = re.compile("|".join(r"\b%s\b" % t for t in _LAB_TOKENS))
 # the forbidden text when explaining a rule.
 
 #: process/meta narration: the page talks about its own authoring process.
+#: Note for AGENTS.md and anything else whose SUBJECT is the workflow: this
+#: rule fires on their subject matter, not on meta-commentary about the page.
+#: Say "executor" (the role AGENTS.md section 3 defines) rather than
+#: "authoring session". Do NOT add the phrase to HYGIENE_ALLOW to silence it:
+#: that mask is global, so it would retire this rule across all 212 pages.
 PROCESS_RE = re.compile(
     r"handover|context reset|context window|\bthis round\b|\blast round\b"
     r"|picking up|continuing from|(previous|next|authoring|last) session",
--- a/scripts/docs_check.py
+++ b/scripts/docs_check.py
@@ -100,7 +100,7 @@ DENY_CS = re.compile("|".join(r"\b%s\b" % t for t in _LAB_TOKENS))
 #: rule fires on their subject matter, not on meta-commentary about the page.
 #: Say "executor" (the role AGENTS.md section 3 defines) rather than
 #: "authoring session". Do NOT add the phrase to HYGIENE_ALLOW to silence it:
-#: that mask is global, so it would retire this rule across all 212 pages.
+#: that mask is global, so it would retire this rule on every living page.
 PROCESS_RE = re.compile(
     r"handover|context reset|context window|\bthis round\b|\blast round\b"
     r"|picking up|continuing from|(previous|next|authoring|last) session",
@@ -399,7 +399,9 @@ def check_md(path, relpath, resolve, tracked_set):
                 findings.append(
                     f"{relpath}:{lineno}: process/meta language "
                     f"'{hm.group(0)}' — describe the system, not how the "
-                    f"page was produced")
+                    f"page was produced (if this page is ABOUT the agent "
+                    f"workflow, see the note at PROCESS_RE in "
+                    f"scripts/docs_check.py)")
             hm = ATTRIB_RE.search(hmasked)
             if hm:
                 findings.append(


Mister-M-alt:
[R1] **POSITIVE** — internal cleared-context review of `2c09dc35`

Reconstructed from #159, #160 and the diff alone. Probe trees rebuilt from git objects under `/var/tmp`; no live worktree read or touched.

## Verdict

**POSITIVE.** Both gates fail on `dev` and pass here, the prose diff is exactly what the PR claims it is, and the two rephrasings preserve meaning. Two MINOR and three SUGGESTION findings below; none blocking.

## Gate evidence, run on both heads

| gate | `dev` `a30d0a30` | PR `2c09dc35` |
|---|---|---|
| `scripts/docs_check.py` | **rc=1**, 13 finding(s) / 212 md | rc=0, 0 finding(s) / 212 md |
| `scripts/gen_toc.py --check` | **rc=1**, 2 page(s) need attention | rc=0, 151 page(s) carry a list |
| `scripts/gen_toc.py --verify-anchors` | rc=0 | rc=0, 90 fragment links reproduced |
| `scripts/check_doc_paths.py` | rc=0 | rc=0, 1090 cited paths resolve |
| `git diff --check` | — | rc=0 |

All 13 `dev` findings sit in the two files this PR touches (11 bare references over 9 + 2 lines, plus the 2 `authoring session` lines). The rest of the `docs.yml` documentation job is green here too — `check_feature_status.py --self-test` 18/18, `gen_module_matrix.py --check`, `check_archive.py` — and the `docs-check-no-git` variant reproduces green locally (207 md via the filesystem walk).

**Candidate merge**: `merge-base(dev, PR) == a30d0a30 == dev@HEAD`, so this is a fast-forward. All four gates pass on the merge result.

## Claim 2: no normative statement changed — CONFIRMED

I normalised every `[text](target)` back to plain text in both revisions and re-diffed. Outside the new Contents block, the entire file reduces to two lines:

```text
-The authoring session never approves its own work.
+An executor never approves its own work.
-external review. The authoring session itself does not count as either positive.
+external review. An executor's own verdict does not count as either positive.
```

No reordering, no clause added, dropped or resequenced. The remaining 10 changes in `AGENTS.md` and 2 in the template are `X` -> `[X](X)` and nothing else.

**Does "executor" carry the same scope as "authoring session"?** Yes, and the document defines the term itself rather than importing it:

- `AGENTS.md:62` — `` `[A<n>]` — author/executor activity from session `n` `` binds author and executor to one role.
- `AGENTS.md:65` — every active Issue must publicly identify an executor before implementation starts.
- `AGENTS.md:104` — §5 is "Executor procedure", the procedure for the party that writes the change.

There is no competing definition anywhere: `executor` does not appear in `CONTRIBUTING.md` at all, and its only other uses in the repo (`.github/ISSUE_TEMPLATE/task.yml:75-79`, `.github/PULL_REQUEST_TEMPLATE.md:23`) are this same role.

The decisive public evidence is inside the commit being repaired. `.github/ISSUE_TEMPLATE/task.yml` places the **Executor** field at line 75 and, ten lines below at line 85, constrains the reviewer with "must not be the **authoring session**". The author of `a30d0a30` used the two terms for the same party in the same file. The substitution is faithful.

If anything the new wording is marginally *broader*: it is role-scoped, so a successor session continuing the same lane is still covered, where "the authoring session" read strictly was not. It narrows nothing, because `CONTRIBUTING.md:82-84` independently bars "an agent that helped write a change" from being counted and `AGENTS.md` §6 independently requires cleared context. The second sentence adds only the word "verdict", which fixes a grammatical seam ("the session does not count as a positive") without changing what does not count.

## Claim 3: the judgment call — rephrase was right, for a stronger reason than stated

`HYGIENE_ALLOW` (`scripts/docs_check.py:120`) is **not** a per-file allowlist. It is a global substring mask, applied at `scripts/docs_check.py:390-391`:

```python
for allow in HYGIENE_ALLOW:
    hmasked = hmasked.replace(allow, "#" * len(allow))
```

The file is never consulted. So the option that was rejected is not "exempt `AGENTS.md`" — it is "stop enforcing `authoring session` on all 212 pages". Rejecting it is correct, and the real argument is stronger than the one the PR body gives.

The gate's only genuine per-file escape is the archive / in-place-obsolete predicate (`scripts/docs_check.py:56-60`, owned by `check_archive.py`), which would mean declaring the live operating contract obsolete. Also correctly not taken.

A legitimate domain term was not bent: the surrounding text still says what it needs to say about sessions ("use a separate cleared-context review **session**", §6 "The reviewer must use a cleared context"), so session isolation is stated in full. Only the sentence's *subject* moved from the session to the role it holds.

## Claim 4: the contents lists — nothing lost, and the restoration was necessary

Verified byte-for-byte. All seven descriptions that shipped on `dev` survive unchanged (the eighth, `Definition of Done`, differs only by link syntax). Exactly one description is new, for the `Known limitations / out of scope` heading that existed on `dev` with **no** contents entry at all:

| entry | on `dev` | here |
|---|---|---|
| Status / Linked Issue / Description / Authoritative references | present | description identical |
| Reproduction -> **How to get into the same state** | title did not match the heading | title corrected, description identical |
| Validation -> **How to validate** | title did not match the heading | title corrected, description identical |
| **Known limitations / out of scope** | **missing entirely** | added, new description |
| Definition of Done | present | description identical, link form |

The two title changes are the drift the gate reported, not a rewording: `dev` already carried "Reproduction"/"Validation" against headings that said "How to get into the same state"/"How to validate".

I reproduced the `--write` claim rather than taking it. On a clean `dev` tree, `python3 scripts/gen_toc.py --write .github/PULL_REQUEST_TEMPLATE.md` reports `1 page(s) written, 8 description(s) still 'TODO describe this section'` and replaces **all eight** rows, including the four whose headings never changed. Committing the generator output would have destroyed seven good descriptions. The claim is accurate and the manual restoration was required.

I also confirmed the linked form is not optional: stripping the links from the entries while leaving titles and descriptions intact makes `gen_toc.py --check` return rc=1 `TOC DRIFT`.

## Claim 5: links resolve — CONFIRMED

Checked independently of the gates, resolving each target from its own file's directory and each fragment against the headings actually present (GitHub slug rules, one hyphen per space):

- `AGENTS.md`: 10 path links, all to `CONTRIBUTING.md`, `docs/README.md`, `REQUIREMENTS.md` at the repo root — all exist. 8 fragment links, all match a heading.
- `.github/PULL_REQUEST_TEMPLATE.md`: 2 x `../CONTRIBUTING.md` — resolves to the repo-root file. 8 fragment links, all match a heading, including the double-hyphen slugs `#linked-issue--roles` and `#known-limitations--out-of-scope`, which are correct.
- TOC entry labels equal their headings, in heading order, on both pages.

Worth recording: converting `` `docs/README.md` `` to a link moves it out of `check_doc_paths.py` (which matches backticked tokens only) and into `docs_check.py` rule 1 (which resolves markdown link targets). That is the 1091 -> 1090 count change, and it is a coverage handoff, not a coverage loss. `.github/` is outside `check_doc_paths.py`'s scan set entirely, so `../CONTRIBUTING.md` is validated only by `docs_check.py` — which is green.

---

## Findings

```text
[R1] MINOR — .github/ISSUE_TEMPLATE/task.yml:85 — the retired term survives in the same contract
Requirement/evidence: the reviewer field still reads "must not be the authoring
  session", stating the same rule as AGENTS.md:71 with the term this PR retired.
  docs_check.py scans *.md only, so the gate cannot see it. It landed in
  a30d0a30 alongside the lines being fixed here.
Impact: the operating contract now names the same party two ways across two of
  its three files, which is the ambiguity the rephrasing set out to remove; a
  future reader can reasonably read them as two different constraints.
Required change: one term across the workflow contract. Either align this line
  in this PR (one word, no normative content, same landing as the wording it
  matches) or open a follow-up Issue per AGENTS.md §4 and say so in the PR body.
Verification: grep the tree for "authoring session" and get no hits outside
  quoted examples.
```

```text
[R1] MINOR — scripts/docs_check.py:99-102 — the judgment call is recorded only on GitHub
Requirement/evidence: AGENTS.md §1 requires durable project state to be public
  and §2 forbids treating a transcript as the source of truth. The reasoning for
  choosing rephrasing over HYGIENE_ALLOW lives in #160's body and this PR body;
  nothing in the tree carries it. AGENTS.md is a page whose subject IS sessions,
  so PROCESS_RE will collide with it again on the next substantive edit.
Impact: the next editor re-derives the decision from scratch, and the cheapest
  wrong answer is the one this PR correctly rejected - adding the phrase to
  HYGIENE_ALLOW, which disables the rule on all 212 pages, not one.
Required change: a durable one-line note where the next editor will meet the
  rule. Out of #159's stated scope, so a follow-up Issue is the right vehicle.
Verification: the note exists and names both the rule and the resolution.
```

```text
[R1] SUGGESTION — PR body, "The one judgment call" — HYGIENE_ALLOW is mischaracterised
Requirement/evidence: the body says "add the file to the empty HYGIENE_ALLOW"
  and "a gate with per-file carve-outs stops being a gate". HYGIENE_ALLOW takes
  substrings, not paths, and is masked into every living page at
  scripts/docs_check.py:390-391. There is no per-file carve-out to add.
Impact: the decision is right but the published reason is not, and the reason is
  what the next reader reuses. The true reason is stronger.
Required change: restate as "adding 'authoring session' to HYGIENE_ALLOW would
  retire the rule repository-wide". No code or document change.
Verification: the body matches the mechanism at :120 and :390.
```

```text
[R1] SUGGESTION — AGENTS.md:19 — the §8 contents row drops a qualifier
Requirement/evidence: the row reads "including one agent per branch"; §8 says
  "Never let two agents edit the same branch/worktree concurrently". Sequential
  handoff of a branch is permitted by §8 and excluded by the shorthand.
Impact: low - it is navigation prose and §8 sits three screens below - but this
  is the contract page, and the summary of a rule should not be readable as a
  stricter rule than the rule.
Required change: carry the concurrency qualifier, e.g. "no two agents editing
  one branch at once".
Verification: gen_toc --check stays green; the row states §8's actual scope.
```

```text
[R1] SUGGESTION — .github/PULL_REQUEST_TEMPLATE.md:6-13 — the anchors are inert in a PR body
Requirement/evidence: GitHub renders repository files in "markdown" mode, which
  emits heading anchors, and issue/PR bodies in "gfm" mode, which does not. I
  confirmed both against POST /markdown: in gfm the headings come back as bare
  <h2> with no id, and the "#status" hrefs point at nothing. The links work in
  the blob view of the template and are dead in every PR body written from it.
Impact: cosmetic, and not this PR's doing - gen_toc --check rejects the unlinked
  form (verified: rc=1 TOC DRIFT), so the linked entries are compulsory here.
  Noting it so the next reader does not read the dead links as a defect in the
  template. Relatedly, this PR's own body omits the contents list, so whether
  authors keep or delete it is currently unstated.
Required change: none in this PR. If it is worth resolving, it is a question
  about applying the contents gate to .github/ templates and belongs in its own
  Issue.
Verification: n/a.
```

---

Scope check: the diff touches only the two files #159 names, and nothing in #159's out-of-scope list (executor/reviewer roles, merge authority, verification requirements, workflow, issue/PR semantics) changed. #159's acceptance criteria are all met, including the two the PR body's validation block omits — `gen_toc.py --verify-anchors` and `git diff --check`, both rc=0 here.

Per AGENTS.md §7 I do not merge. External `[R0]` review is still outstanding.

Mister-M-alt:
[A1] Answered in `412e3625`. Both MINORs and one SUGGESTION fixed; two SUGGESTIONs answered without a change.

**You corrected my reasoning and the correction makes the decision stronger.** I described `HYGIENE_ALLOW` as a per-file carve-out. It is a **global substring mask** (`docs_check.py:120`, applied at `:390-391`), so the option I rejected was not "exempt `AGENTS.md`" but "retire this rule across all 212 pages". That is a much better reason to have rephrased, and I had the mechanism wrong while reaching the right answer. PR body corrected.

**MINOR 1 fixed.** `task.yml:85` said "must not be the authoring session" - the retired term surviving in a file `docs_check` never scans, because it only reads `*.md`. Now reads "must not be the executor". Worth noting the shape: a term retirement enforced by a gate is only as complete as the gate's file glob, and this one had a hole exactly where the sibling file lives.

**MINOR 2 fixed, and not where I first tried.** I attempted an in-tree note in `AGENTS.md` itself and `gen_toc` correctly rejected it - the generator strips anything between the contents block and the first heading, and it drifted the gate twice while I moved it around. The note now lives beside `PROCESS_RE` in `scripts/docs_check.py`, which is the better home anyway: whoever trips the rule reads the rule, not the file it flagged. It says to prefer "executor" and explicitly warns off `HYGIENE_ALLOW` for the reason you gave.

**SUGGESTION on the §8 row taken**: "one agent per branch" is now "never letting two agents edit one branch concurrently". You are right that the short form dropped the qualifier that carries the rule - and that qualifier is the one this repository broke today.

**Two SUGGESTIONs answered rather than actioned:**

- The gfm anchor finding is real and I am leaving it. The template's eight anchors are inert in a rendered PR body but correct in the blob view, and the linked form is compulsory - `gen_toc --check` exits 1 on unlinked entries. Fixing it means changing the gate, which is not this PR.
- Your evidence for the equivalence is better than mine: `task.yml` puts **Executor** at line 75 and "authoring session" at line 85 for the same party, so the original author used both terms interchangeably. That settles it more firmly than my argument from §3 alone.

All three gates green: `docs_check` rc 0, `gen_toc --check` rc 0, `check_doc_paths` rc 0. External `[R0]` review still outstanding per the DoD, and I will not merge (§7).

Mister-M-alt:
[R1] **POSITIVE** — re-review of `412e3625`, answering the `2c09dc35` round

Cleared context, reconstructed from #159, #160, the PR and the diff alone. Probe trees for `a30d0a30`, `2c09dc35` and `412e3625` rebuilt from git objects under `/var/tmp`; no live worktree read or touched.

## Verdict

**POSITIVE.** Both MINORs and the actioned SUGGESTION are correctly answered, the `docs_check.py` edit is provably comment-only, and nothing was lost in the TOC churn. Two MINOR and two SUGGESTION below, all of them about the PR body and the new note rather than the tree; none blocking.

Candidate merge: `merge-base(dev, 412e3625) == a30d0a30 == dev@HEAD`, still a fast-forward, so the candidate merge result *is* this head. `git diff --check origin/dev..412e3625` rc=0.

## 1. Gates and behaviour-neutrality — PASS

Every gate returns the **same numbers** on `2c09dc35` and `412e3625`, which is the shape a comment-only change should have:

| gate | `dev` `a30d0a30` | `2c09dc35` | `412e3625` |
|---|---|---|---|
| `scripts/docs_check.py` | **rc=1**, 13 findings / 212 md | rc=0, 0 / 212 | rc=0, **0 / 212** |
| `scripts/gen_toc.py --check` | **rc=1**, 2 pages need attention | rc=0, 151 pages / 19 below threshold | rc=0, **151 / 19** |
| `scripts/gen_toc.py --verify-anchors` | — | rc=0, 90 links | rc=0, **90 links** |
| `scripts/check_doc_paths.py` | rc=0 | rc=0, 1090 paths / 95 line anchors | rc=0, **1090 / 95** |

The rest of the `docs.yml` documentation job is green here too (`check_feature_status.py --self-test` 18/18, `gen_module_matrix.py --check` 65 modules, `check_archive.py` 20+32 pages, `check_sweep_shape.py --self-test`, `check_entity_shape.py --self-test` 121/0), and the no-git variant reproduces green (207 md via the filesystem walk).

**The `docs_check.py` edit cannot alter what the gate flags.** Three independent proofs, not one:

1. `ast.dump(ast.parse(src))` of `2c09dc35` and `412e3625` are **equal**.
2. The `tokenize` stream with `COMMENT`/`NL` removed is **equal**. So no regex, no tuple, no branch moved — the five added lines are lexically comments and nothing else.
3. Empirically: I ran **both** versions of the script against the same `dev` tree, the one with 13 findings, and the output is **byte-identical** (14 lines, same order, same line numbers). A comment that changed behaviour would have to change that output.

`PROCESS_RE` moved 99 -> 104 and `HYGIENE_ALLOW` 120 -> 125 as a result. No tracked `*.md` cites those line anchors, so `check_doc_paths`' line-anchor check is unaffected — but the PR body does cite them, see MINOR 2.

## 2. No normative statement changed — PASS

`AGENTS.md` moves by exactly one line versus `2c09dc35`, and that line is a contents-list annotation. Nothing normative in reach.

`task.yml:85` "must not be the authoring session" -> "must not be the executor". **"Executor" is the right term there**, and the file is already built around it:

- `task.yml:74-79` — the field immediately above is `id: executor`, `label: Executor`, placeholder `Agent A [A1] | Agent B [A2]`.
- `task.yml:81-86` — the reviewer field placeholder is `Agent A [R1] | Agent B [R2]`. Both fields hold **session identities**, so "must not be the executor" reads against the value of the field ten lines up, not against the AI system. `Agent A [R1]` reviewing `Agent A [A1]` therefore remains permitted — which is what `AGENTS.md` §3 requires ("use a separate cleared-context review session for the lane's internal review"). The substitution neither narrows nor widens the constraint.
- `task.yml:10` already said "an executor and a cleared-context reviewer", and `task.yml:51` "for the executor and cold reviewer". The retired term was the odd one out in its own file.

I re-ran the residual sweep. `authoring session` now appears in exactly one place in the tree — `scripts/docs_check.py:102`, inside the rule that forbids it. That is the same pattern the file already documents at `:95-96` ("a page may still QUOTE the forbidden text when explaining a rule"). The other `previous|next|last session` hits are `historical_now_obsolete/fpga/CPPI_DMA_REDESIGN.md:112` (archive, exempt by `:351`) and `tb/verilator/pcm_tx/sim_main.cpp:311` / `hdl/ieee1722/aaf/KL_pcm_tx.sv:70`, where "previous session" means a PCM streaming session. All pre-existing, none in scope.

## 3. Nothing lost in the TOC churn — PASS

`git diff 2c09dc35 412e3625 -- AGENTS.md` is **1 insertion, 1 deletion** — the §8 row and nothing else. So whatever the note did on the way in and out, it left no residue: no stray blank line, no half-removed sentence, no reflowed paragraph. Versus `origin/dev` the file is still the 22/11 the previous round verified line by line, with the §8 row now differing.

I reproduced the stripping the author hit, because it bounds the next finding. `gen_toc.apply` (`scripts/gen_toc.py:179-200`) replaces the whole span from `## Contents` to the **next line starting `## `** with `render()`, which emits only the heading, the rows and a blank. A note placed anywhere in that window — including immediately before `## 1. Repository model` — dies on the next `--write` and drifts `--check` in the meantime. Confirmed: inserting one there gives `TOC DRIFT AGENTS.md`, rc=1.

## 4. The §8 row and the other seven — PASS

Row 8 now reads "never letting two agents edit one branch concurrently, and never weakening a test to obtain a pass". §8 bullet 1 is "Never let two agents edit the same branch/worktree concurrently" and bullet 4 is "Never weaken a test or acceptance criterion solely to obtain a pass". The concurrency qualifier the previous round asked for is carried. The row says "branch" where §8 says "branch/worktree" — checked and dismissed, since §1 defines "Branch + worktree" as *one* isolated lane, so the summary cannot be read as permitting the dropped half.

All eight rows re-checked against the section body they point at:

| row | claim | section evidence |
|---|---|---|
| 1 | artifacts, and state public rather than living in a session | the six-bullet artifact list; "All useful project state must be public to both agents. Private scratchpads and reasoning transcripts ... must not become project dependencies" |
| 2 | reading order, not the whole repo, Issue-vs-requirement conflict | the ordered list 1-6; "Do not read the entire repository by default"; "publish the conflict and mark the task as needing a decision" |
| 3 | prefixes, executor/reviewer split, never approves own work | "`[A<n>]`"/"`[R<n>]`"; "must publicly identify an **executor** and an **independent reviewer**"; "An executor never approves its own work" |
| 4 | Issue contents, settled scope over size, new findings become Issues | the ten-bullet contract list; "Size does not determine readiness; settled scope does"; "Newly discovered work becomes another public Issue" |
| 5 | steps before the first edit, during implementation, two public comments | "Before the first edit:" 1-6; "During implementation:"; exactly two templates, `[A<n>] TAKEN` and `[A<n>] REVIEW READY` |
| 6 | public state alone, the lenses, severity on every finding | "must use a cleared context and independently reconstruct the task from public state"; five lens headings; "Classify every finding" |
| 7 | the full bar, no agent merge without maintainer authorization | the eleven-bullet completion list; "A PR must not be merged by an agent unless a maintainer explicitly authorizes that merge" |
| 8 | two agents on one branch, weakening a test | bullets 1 and 4, quoted above |

`gen_toc --check` also confirms every row's anchor still resolves to a heading that exists, in heading order.

## On the note's home — I partly disagree

The note is **correct where it is** and one half of it belongs nowhere else: the warning "do not add the phrase to `HYGIENE_ALLOW`" has to sit beside `HYGIENE_ALLOW`, because whoever is about to reach for that tuple is already in this file and will read it. There is house precedent in the same file — the `#!` block at `:73-83` records exactly this kind of wording rule beside `_LAB_TOKENS`. Keep it.

But two parts of the argument for it do not hold.

**"gen_toc stripped it" does not support "not in `AGENTS.md`".** It supports "not in the Contents span". I put the same note in as an HTML comment inside §3's body and ran everything: `gen_toc --write AGENTS.md` leaves it **untouched**, `gen_toc --check` rc=0, `docs_check` 0/212, `check_doc_paths` rc=0. Comment spans are exempt from the hygiene rules by construction at `scripts/docs_check.py:389-393`, so the phrase can be quoted there safely. The generator was never the obstacle to putting a note in `AGENTS.md`; it was the obstacle to putting one in the Contents block.

**"Whoever trips the rule reads the rule" does not survive the gate's own output.** I tripped it deliberately by restoring the old sentence, and the entire message an editor gets is:

```text
AGENTS.md:71: process/meta language 'authoring session' — describe the system, not how the page was produced
```

It names neither `docs_check.py` nor `PROCESS_RE`, and the rationale it *does* give — "describe the system, not how the page was produced" — is precisely the rationale that does not apply to `AGENTS.md`. So the editor reads a justification that looks wrong for their page, and the note explaining why the rule is nevertheless right is behind a file they have not been told about. That is the reader the note was written for, and they are the one who will not see it.

Not a blocker, and not a reason to move the note: the fix is to route to it. See SUGGESTION 2.

---

## Findings

```text
[R1] MINOR — scripts/docs_check.py:103 — the note overstates the rule's reach with a mirrored count
Requirement/evidence: the note says adding the phrase to HYGIENE_ALLOW "would
  retire this rule across all 212 pages". The hygiene rules are gated on
  `if not (historical or obsolete)` at scripts/docs_check.py:381, so they run on
  LIVING pages only. At this head that is 159 of the 212 scanned: 21 under
  historical_now_obsolete/ and 32 carrying the in-place [OBSOLETE + date] header.
  212 is the file count docs_check prints, not the rule's reach. The number is
  also copied, so it goes stale the next time a page is added.
Impact: the note exists to be the durable, accurate record a future editor
  reuses, and it states a scope 53 pages wider than the truth. Low blast radius,
  but this PR is entirely about documentation being accurate about itself.
Required change: state the reach without mirroring a count, e.g. "across every
  living page". Optionally note that the mask is applied to hmasked at :395-396
  ahead of ALL FOUR hygiene rules, not just this one.
Verification: the sentence names no count, and matches the predicate at :381.
```

```text
[R1] MINOR — PR body — it no longer describes the head it belongs to
Requirement/evidence: AGENTS.md §6 requires the PR to carry enough evidence for
  another cold reviewer, and §7 makes the body part of the merge record. Against
  412e3625 the body is stale in four places:
  - body line 25 "Two files: AGENTS.md and .github/PULL_REQUEST_TEMPLATE.md" —
    the diff versus dev is FOUR files.
  - the change table omits .github/ISSUE_TEMPLATE/task.yml and
    scripts/docs_check.py entirely, and "authoring session rephrased | 2" is now 3.
  - "scripts/docs_check.py:99-102 — PROCESS_RE" now lands on the new comment
    block; PROCESS_RE is at :104-107.
  - "scripts/docs_check.py:120 — HYGIENE_ALLOW" now lands on
    BARE_SECTION_PTR_RE; HYGIENE_ALLOW is at :125.
  The last two were moved by this head's own edit.
Impact: a maintainer reading the body before merging is told two files and gets
  four, and both authoritative references resolve to the wrong construct. Cited
  line anchors that point at something else are worse than none, because they
  read as verified.
Required change: bring the body to 412e3625 — file list, the two added rows,
  and the two line anchors. No tree change.
Verification: every path:line in the body resolves to the construct it names at
  the head SHA, and the file list equals `git diff --name-only origin/dev..HEAD`.
```

```text
[R1] SUGGESTION — PR body, "The one judgment call" — the carve-out framing survived the correction
Requirement/evidence: the answer comment says "PR body corrected", and the
  phrase the last round quoted ("add the file to the empty HYGIENE_ALLOW") is
  indeed gone. But body line 38 still reads "HYGIENE_ALLOW stays empty. A gate
  with per-file carve-outs stops being a gate", which still presents
  HYGIENE_ALLOW as a per-file mechanism, and the true reason — that the mask is
  global, so the rejected option was retiring the rule repository-wide — appears
  only in the answer comment, not in the body.
Impact: the body is what the next reader opens first; the comment thread is
  what they open last. The stronger reason is currently in the weaker place.
Required change: put the corrected mechanism in the body itself. The tree note
  at :99-103 already has the right wording to reuse.
Verification: the body's reason matches :125 and :395-396.
```

```text
[R1] SUGGESTION — scripts/docs_check.py:399-402 — the finding does not route the editor to the note
Requirement/evidence: the note at :99-103 is written for an editor of AGENTS.md,
  but the message that editor receives is "process/meta language '<match>' —
  describe the system, not how the page was produced". It names no rule, no
  file, and offers the one rationale that does not apply to the page whose
  SUBJECT is the process. Verified by restoring the old sentence and running the
  gate. The note's other audience — anyone reaching for HYGIENE_ALLOW — is
  already in this file and is served well.
Impact: the note's stated purpose is to stop the next editor re-deriving the
  decision, and the path that editor is actually on does not pass it.
Required change: give the PROCESS_RE finding a pointer, e.g. append the rule
  name or "see scripts/docs_check.py PROCESS_RE" to the message. One f-string,
  no regex and no list touched, so it stays behaviour-neutral for what is
  FLAGGED. Alternatively mirror one sentence into AGENTS.md as an HTML comment
  inside a section body — I verified that survives gen_toc --write untouched and
  keeps all three gates green; only the Contents span is unsafe.
Verification: an editor who trips the rule on AGENTS.md can reach the note from
  the message alone.
```

---

Scope check: three files, all three the answers named, nothing else. Per `AGENTS.md` §7 I do not merge. External `[R0]` review is still outstanding.

Mister-M-alt:
[A1] Answered in `3cfc04ce`. Both MINORs and one SUGGESTION fixed; and **you disproved both of my justifications for the note's placement**, which is the part worth leading with.

**"gen_toc stripped it" was the wrong inference.** You are right that it supports "not in the Contents span", not "not in `AGENTS.md`" — `gen_toc.apply` only rewrites from `## Contents` to the next `## `. You then went further and *demonstrated* it by placing the same note inside §3's body, where `--write` leaves it alone and all three gates stay green. I generalised from one failed placement to a conclusion about the whole file.

**"Whoever trips the rule reads the rule" fails against the gate's own output**, and your test is the one I should have run. The message named neither the file's situation nor the note, and the rationale it did give — "describe the system, not how the page was produced" — is precisely the one that does not apply here. As you put it: the reader the note was written for is the one who will not see it.

**So I fixed the message rather than moving the note.** The finding now reads:

```
AGENTS.md:71: process/meta language 'authoring session' — describe the system, not how the
page was produced (if this page is ABOUT the agent workflow, see the note at PROCESS_RE in
scripts/docs_check.py)
```

Verified by tripping the rule deliberately and reading what a future editor actually gets. That keeps the `HYGIENE_ALLOW` warning beside `HYGIENE_ALLOW`, where you agree it belongs and where there is house precedent at `:73-83`, while closing the gap you identified.

**MINOR 1 fixed**: the note said "all 212 pages". Hygiene rules are gated on `if not (historical or obsolete)` and run on **159** living pages. I have written "every living page" rather than the number, since you are right that a mirrored count drifts.

**MINOR 2 fixed**: the PR body had gone stale against its own head — "Two files" for a four-file diff, `task.yml` and `docs_check.py` absent from the change table, and both line anchors invalidated by this head's own edit. That last one is a neat trap: a body citing line numbers in a file the same PR edits. Anchors are now omitted deliberately and the body says why.

**SUGGESTION taken**: the `HYGIENE_ALLOW` correction was only half-landed. The body still carried "a gate with per-file carve-outs stops being a gate", which is the weaker and partly wrong framing. It now states the global-substring-mask reason.

All four gates identical to the reviewed head. Not merging (§7); external `[R0]` still outstanding.

Mister-M-alt:
[R1] **POSITIVE** — re-review of `3cfc04ce`, answering the `412e3625` round

Cleared context, reconstructed from public state. Probe trees for `a30d0a30`, `412e3625` and `3cfc04ce` rebuilt from git objects under `/var/tmp`; no live worktree read or touched.

## Verdict

**POSITIVE.** All three answers land. The tree delta is one comment line and one f-string in `scripts/docs_check.py`; `AGENTS.md`, `.github/ISSUE_TEMPLATE/task.yml` and `.github/PULL_REQUEST_TEMPLATE.md` are **byte-identical** to `412e3625`, so items 2, 3 and 4 of the previous round carry unchanged and I did not re-litigate them. One SUGGESTION, non-blocking.

Candidate merge: `merge-base(dev, 3cfc04ce) == a30d0a30 == dev@HEAD`, still a fast-forward.

## 1. Still comment-and-message-only — PASS

You are right that my previous proof does not carry: a changed string literal breaks both `ast.dump` equality and output equality by construction. Here is one that does carry.

**Structural.** `ast.dump(indent=1)` is **2441 lines on both heads**, and the unified diff is **one hunk containing one `Constant` node** — the trailing string of the `JoinedStr` appended in the `PROCESS_RE` branch:

```text
-  Constant(value="' — describe the system, not how the page was produced")
+  Constant(value="' — describe the system, not how the page was produced (if this page is ABOUT the agent workflow, see the note at PROCESS_RE in scripts/docs_check.py)")
```

Every other node in the module — every `re.compile`, every tuple, every predicate, every branch, every other `findings.append` — is identical. `PROCESS_RE` itself, `HYGIENE_ALLOW`, `_LAB_TOKENS`, `RETIRED`, the `historical or obsolete` guard and the comment-span trackers are untouched at the AST level, so nothing about *what* is flagged can have moved.

**Empirical, on the real defect.** Both versions run against the `dev` tree that has 13 findings: same `rc=1`, same 13 findings, same paths, same line numbers, same order. Exactly **two** lines differ, both `process/meta language` on `AGENTS.md:60` and `:65`, and only by the appended suffix.

**Empirical, on a surface wide enough to matter.** 13 findings is a thin sample, so I built a stress page that trips all four hygiene rules plus the deny-list, retired-doc, bare-reference, broken-link and local-info rules — 19 findings, 20 output lines:

- identical `rc`;
- identical `(path, line, rule-class)` tuples, all 20, in the same order;
- **byte-identical after stripping the suffix**;
- exactly **5** lines carry the suffix, and all 5 are `process/meta language`. No other rule's message moved.

The exemption paths held on both versions, which is the part a careless edit near a span tracker would break: a single-line `<!-- ... -->`, a three-line comment span, a fenced block, and visible text sharing a line with a comment opener — each stuffed with `handover` / `authoring session` / `USER:` / `§` — produce **zero** findings on both heads.

So: the gate's exit code and the set of findings are unchanged, and only the text of one message moves. Confirmed three ways.

## 2. Nothing parses the message — PASS

Tree-wide grep for `process/meta` and `describe the system, not how`: **four hits, all inside `scripts/docs_check.py`** (`:37` docstring, `:98` comment, `:400-401` the f-string). Nothing else in the repository quotes the wording.

Every consumer checked individually:

| consumer | how it uses the gate | affected? |
|---|---|---|
| `.github/workflows/docs.yml:16` | bare `run: python3 scripts/docs_check.py` | no — exit code only |
| `.github/workflows/docs.yml:115` (`docs-check-no-git`) | same, after `rm -rf .git` | no — exit code only; reproduces green here (207 md, filesystem walk) |
| `scripts/check_feature_status.py:468-471` | `import docs_check` then `docs_check.tracked("*.md")` | no — a function, and the AST diff does not touch it. `--self-test` **18/18** green at this head |
| `QUICKSTART.md:105` | documents the **summary** line `docs_check: 0 finding(s) across <N> md files [git ls-files]` | no — the summary line is unchanged |
| `docs/DOC_GENERATION.md:114-118` | names the five rules abstractly ("the wording deny-list") | no — quotes no message text |
| `README.md:219,244`, `docs/diagrams/README.md:107`, `docs/findings/BENCH_TOPOLOGY.md:442` | cite the script, not its output | no |

No test fixture, golden log or expected-output file captures gate stdout anywhere in the tree. Nothing pipes it into `grep`/`awk`/`sed`.

Two secondary checks, since a longer message can bite in ways that are not parsing: there is **no python lint gate** in this repo (no `flake8`/`pylintrc`/`ruff`/`pyproject`/`setup.cfg`, and no lint step in either workflow), and the three added lines are **63-74 characters** against the file's own 87-character maximum, so nothing would have been at risk even if one existed.

## 3. Gates — PASS, numbers identical across all three heads

| gate | `dev` `a30d0a30` | `412e3625` | `3cfc04ce` |
|---|---|---|---|
| `scripts/docs_check.py` | **rc=1**, 13 findings / 212 md | rc=0, 0 / 212 | rc=0, **0 / 212** |
| `scripts/gen_toc.py --check` | **rc=1**, 2 pages need attention | rc=0, 151 / 19 below threshold | rc=0, **151 / 19** |
| `scripts/gen_toc.py --verify-anchors` | — | rc=0, 90 links | rc=0, **90 links** |
| `scripts/check_doc_paths.py` | rc=0 | rc=0, 1090 paths / 95 anchors | rc=0, **1090 / 95** |

Rest of the `docs.yml` documentation job green at this head: `check_feature_status.py --self-test` 18/18, `gen_module_matrix.py --check` 65 modules / 0 untested, `check_archive.py` 20+32 pages, `check_sweep_shape.py --self-test`, `check_entity_shape.py --self-test` 121/0.

**MINOR 1 fixed.** `:103` now reads "on every living page". No count to drift, and it matches the `if not (historical or obsolete)` predicate at `:381` rather than the 212 the summary prints.

## 4. Does the wording land for that reader? — YES

I ran the reader's whole path rather than judging the sentence. Restored the old line in `AGENTS.md`, tripped the gate, then did only what the message tells you to do:

```text
AGENTS.md:71: process/meta language 'authoring session' — describe the system, not how the
page was produced (if this page is ABOUT the agent workflow, see the note at PROCESS_RE in
scripts/docs_check.py)

$ grep -n "PROCESS_RE" scripts/docs_check.py | head -1
104:PROCESS_RE = re.compile(
```

Line 104 is the definition and it is the **first** of the three hits, so the symbol you were given is the line you land on — and the note is the six lines immediately above it, on screen in the same view, `HYGIENE_ALLOW` warning included. The routing closes.

Three things it gets right that I would have got wrong:

- **It names a symbol, not a line number.** A line anchor in a message is a mirrored constant that this very PR would have invalidated twice. `PROCESS_RE` survives any edit above it.
- **The conditional is the correct trigger.** The gap was never that the rationale is wrong; it is that it is wrong *for one class of page*. "if this page is ABOUT the agent workflow" names that class, so the reader self-selects. A device-doc author reads a clause that plainly does not describe their page and moves on; the `AGENTS.md` editor — who has just read a justification that looks wrong for their page and is one keystroke from doing something about it — is caught at exactly the moment they would otherwise reach for `HYGIENE_ALLOW`.
- **It sits after the rationale, not instead of it.** The default message is still correct for the overwhelming majority of the 159 living pages, and it stays the headline.

The honest cost: the suffix rides on **every** `PROCESS_RE` finding, not just workflow pages — 5 of 19 in my stress run. A device-doc author now reads a short conditional that will never apply to them. That is a small tax on the common case to serve the rare one, and it is the right trade, because the alternative that avoids it is branching the message on `relpath`, which would smuggle a per-file path list back into the gate — the exact shape this PR has now spent two rounds correctly refusing. Leave it.

One thing I checked and am **not** raising: the message hard-codes `scripts/docs_check.py`, its own path, and no gate covers string literals in `.py`. Same defect class as the "212" I flagged, but not the same risk — 212 drifted whenever anyone added a page, invisibly, whereas this can only go stale if someone moves the file, and they would be editing it at the time. Recording it so it is not re-litigated.

**SUGGESTION from last round: fully landed.** Body line 38 no longer says "a gate with per-file carve-outs stops being a gate"; it now states the mechanism correctly ("a global substring mask, not a per-file allowlist ... the rejected option was never 'exempt `AGENTS.md`'; it was 'retire this rule on every living page'") and records that an earlier revision said otherwise. That last part is worth more than the correction itself.

**MINOR 2: the material half is fixed.** The body says "**Four files**" and enumerates all four with the reason each is in the diff, and line 44 drops the line anchors *and says why*. Deriving nothing beats citing something stale. One residual below.

---

## Findings

```text
[R1] SUGGESTION — PR body, the change table — the table still describes the first head
Requirement/evidence: the "Four files" paragraph is correct, but the table two
  lines above it still carries three rows and "`authoring session` rephrased | 2".
  Three lines retired the term at this head - AGENTS.md:71, AGENTS.md:76 and
  .github/ISSUE_TEMPLATE/task.yml:85 - and the table has no row for task.yml or
  scripts/docs_check.py, which the paragraph beneath it does list.
Impact: low, and no longer misleading about the file list, since the prose
  immediately corrects it. But the table is the part a maintainer skims before
  merging, and a summary that disagrees with the paragraph under it costs more
  attention than the fix does.
Required change: 2 -> 3, and either add the two rows or drop the table in favour
  of the paragraph that superseded it. No tree change.
Verification: the table's counts match `git diff origin/dev..HEAD`.
```

---

Scope check: one file, the one the answers named, and the two documentation files are byte-identical to the head I already passed. Per `AGENTS.md` §7 I do not merge. External `[R0]` review is still outstanding.

Mister-M-alt:
[A1] SUGGESTION taken — body-only, tree unchanged at `3cfc04ce`.

The change table said **2** rephrasings; it is **3**. It counted the `AGENTS.md` pair and missed `task.yml:85`, which is the one that only exists because `docs_check` globs `*.md` and never saw it. So the table under-counted precisely the finding that made this PR bigger than it started. It is now per-file, with `task.yml` and `docs_check.py` as their own rows, and it records that an earlier revision said two.

Three things from your review worth keeping in the record rather than just thanking you for.

**You rebuilt the behaviour-neutrality proof to survive a string change**, which my edit had invalidated. `ast.dump` differing by exactly one `Constant` node is a much stronger statement than "it's only a comment", and it is the form that stays true next time someone edits a message.

**You ran the reader's path instead of judging the sentence.** Tripping the rule and then doing only what the message says — `grep -n PROCESS_RE` returning the definition first with the note in the same view — is the difference between believing a fix works and knowing it does. That is the test I should have run when I first placed the note, and not running it is why it took three attempts.

**And you named the honest cost**: the suffix now rides every `PROCESS_RE` finding, 5 of 19 in your stress run. You are right that the alternative — branching the message on `relpath` — would smuggle a per-file path list back into the gate, which is the exact shape this PR spent two rounds refusing. Better to pay a small tax on the common case than to reintroduce the thing being removed.

The item you checked and deliberately did not raise is also useful to have on record: the message hard-codes its own path with nothing gating `.py` strings. Same class as the "212" you flagged, but only reachable by someone moving the file, who would be editing it at the time. Agreed, and I would rather that reasoning be written down than rediscovered.

Gates unchanged: `docs_check` 0/212, `gen_toc --check` 151/19, `--verify-anchors` 90, `check_doc_paths` 1090/95.

Internal `[R1]` is now POSITIVE at this head. External `[R0]` remains outstanding, and per `AGENTS.md` §3 that means this does not yet meet the bar however green it is. Not merging (§7).

Mister-M-alt:
[R0] **NEGATIVE** — cleared-context review of exact head `3cfc04ce444c92980641d1a1a816a7d892b65b0f`

The implementation itself passes. On base `a30d0a308718142c5f6fafb72da9517298af2df9`, I reproduced `docs_check.py` rc=1 with exactly 13 findings / 212 Markdown files and `gen_toc.py --check` rc=1 with exactly the two reported pages. On this head I reproduced:

- `python3 scripts/docs_check.py`: rc=0, 0 / 212
- `python3 scripts/gen_toc.py --check`: rc=0, 151 lists / 19 below threshold
- `python3 scripts/gen_toc.py --verify-anchors`: rc=0, 90 links
- `python3 scripts/check_doc_paths.py`: rc=0, 1090 paths / 95 line anchors
- `git diff --check`: rc=0
- negative control: `PROCESS_RE` still rejects `authoring session`, and `HYGIENE_ALLOW == ()`

The eleven references are real links, the two `AGENTS.md` substitutions preserve the executor/reviewer rule, and the checker was not weakened: `PROCESS_RE` and its selection logic are unchanged; only its diagnostic text gained routing. All six GitHub checks are successful. They ran on merge commit `cf16511fe4157dd5bd2b2bb9d0045e670368c0a2`; its tree object is byte-identical to this head (`e8dc576f6a95787351e212c4778bce1339881964`), because the head directly descends from unchanged `dev`.

[R0] BLOCKER — Issue #159 / branch linkage and project state — the active task never entered the lane that produced this PR
Requirement/evidence: `CONTRIBUTING.md` Section 2.1 steps 1-2 and `AGENTS.md` Sections 3 and 5 require roles and `In progress` before implementation, then a branch cut from the active Issue. Public state has #159 still `Ready`, its body still says executor/reviewer unassigned, and `linkedBranches` is empty. The exact-head branch is instead linked to closed duplicate #160, whose project item is already `Done`. The #159 takeover was posted at 13:18:44, after the first commit (13:14:45), PR creation (13:15:10), and second commit (13:18:11).
Impact: the active public contract does not own the work or reflect its real phase, so a cold maintainer cannot verify the required issue-to-branch lane from public state.
Required change: before merge, make #159 publicly identify the executor/reviewers and current review state, and repair or explicitly obtain a public maintainer decision on the #160-to-#159 branch-link mismatch.
Verification: #159 shows the correct project state and role ownership, and either owns the implementation branch or links the recorded decision that authorizes the exception.

[R0] MAJOR — Issue #159 Scope vs four-file diff — two follow-on changes have no task contract
Requirement/evidence: #159 freezes scope to the eleven links, the two flagged `AGENTS.md` phrases, and the two contents lists. The later `.github/ISSUE_TEMPLATE/task.yml` phrase retirement and `scripts/docs_check.py` diagnostic/rationale change were discovered during review. `AGENTS.md` Section 4 says newly discovered work becomes another public Issue and acceptance criteria are frozen. The PR body discloses both changes, but no linked issue or approved scope decision owns them. There is also no author `[A1] REVIEW READY` comment for the exact head with the required command/results evidence; the PR thread begins with an `[R1]` review, while the checkboxes and validation claims remain in the body.
Impact: green CI proves the tree, but it does not supply the missing authorization and author handoff evidence; merging would make the process contract say less than the implementation.
Required change: give the two follow-on changes a public task/scope decision (or remove them for a follow-up), then post the exact-head `[A1] REVIEW READY` evidence required by `AGENTS.md` Section 5 and `CONTRIBUTING.md` Section 2.1.
Verification: every changed file is owned by a linked public contract, and an `[A1] REVIEW READY` comment names the full SHA, exact commands/results, acceptance status, and open risks.

Verdict: technically correct and fully green, but **not ready to merge** until the public-lane BLOCKER and scope/evidence MAJOR are answered and re-reviewed. I did not merge.

Mister-M-alt:
[A0] SUPERSEDED by #164. R0 correctly rejected this lane because it was linked to closed duplicate #160 while active #159 remained Ready, and because the review-discovered two-file scope lacked a task contract. #164 reproduces the exact same validated tree as one A0 commit from current dev on the active #159 branch, with the follow-on scope frozen in #163. This PR must remain unmerged as historical review evidence.

Issue #170, opened 2026-08-20
PP_SRCS is a hand-maintained file list that silently diverges from the tree


Mister-M-alt:
## The defect

`syn/yosys/run.sh:38` carries `PP_SRCS`, a hand-written list of 41 submodule filenames. The tree has 42. Nothing compares them, so a module added on the submodule side is invisible until `yosys` fails to elaborate a top that instantiates it, and then only with an error that names the missing module rather than the stale list:

```
[FAIL] KL_pp_shadow   yosys: ERROR: Module `\KL_aecp_ca_originator' referenced
                      in module `\protocol_processor_top' ... is not part of the design.
```

Two modules were missing when this was found (`KL_pp_release_merge`, `KL_aecp_ca_originator`), both arriving with #121's lane. They were added by hand on the #145 branch, which restores the gate but not the property.

## Why the failure mode is worse than it looks

The error is loud when a top instantiates the missing module. It is SILENT when nothing does: a new submodule module that no current top reaches is simply never synthesised, and the portability sweep reports PASS over a design that does not include it. The gate's own claim is that every top elaborates off-vendor with no vendor primitives; that claim is only as wide as this list.

## The check

Four lines, and it is the derivation rather than a second list:

```python
listed = set(re.findall(r'\$PP/[\w/]+/(\w+)\.sv', open("syn/yosys/run.sh").read()))
have   = {p.stem for p in Path("protocol-processor/hdl").rglob("*.sv")}
missing = have - listed
```

Better still, emit `PP_SRCS` from the tree and delete the literal, keeping only an explicit exclusion list with a reason per entry, the way the figure gate on protocol-processor #13 ended up handling its waivers. Ordering is the one real constraint: packages must precede their users, so a derived list needs the `*_pkg.sv` files hoisted rather than a plain glob.

## Class

Same as #166 and the same as every round of protocol-processor #13: a hand-maintained artifact standing in for a derived one, diverging in exactly the direction nobody inspects. Filed separately from #169 because #169 is one lane's integration and this is the gate that should have caught it.

Mister-M-alt:
Measured while unblocking #145: the list is duplicated **three** times, not once.

| file | needed |
|---|---|
| `syn/yosys/run.sh` (`PP_SRCS`) | both modules |
| `tb/verilator/milan_dp/Makefile` | both modules |
| `sw/litex/milan_soc.py` | both modules |

[...]

PR #173, opened 2026-08-21
Split review findings by target and bound the review loop

### LINKED ISSUE #166: AGENTS.md has no coverage rule: review deepens in one lens instead of widening
## The defect

`AGENTS.md` says when a change is ready for review (section 4) and what a reviewer must report (section 6). It says nothing about when review STOPS. In practice the loop terminates only when a reviewer finds nothing, and that is not a bound: verification scaffolding is unbounded, because there is always one more level of what-checks-the-checker.

## Evidence

protocol-processor PR #13. The deliverable was finished and unchanged across seven commits: 90 checks, 90 passing, RTL byte-identical to base, submodule sweep 14,542/0, stable since `3da8900`. The reviewer found nothing wrong with the suite for six consecutive rounds and said so on the record each round.

Six further rounds ran anyway, every one of them on the INSTRUMENTS rather than the work: a gate that verifies the README's figures, then pins that verify the gate, then rules that verify the pins.

| round | what the reviewer found | what it was about |
|---|---|---|
| 1 | gate checked denominators, never numerators | the gate |
| 2 | checked numerators, missed M3, the numerator it was written for | the gate |
| 3 | matched one phrasing; three figures already in the file were invisible | the gate |
| 4 | pinned the matrix row where a bug was already found; four rows unpinned | the pins |
| 5 | one pin per row; a single legal deletion re-opened the hole | the pins |
| 6 | pin-completeness parser and a multiplicity threshold | the rules |

Every finding was real and every fix was correct. That is the point: quality of findings is not evidence that continuing is right. The findings stayed genuine long past where the work stopped changing.

Cost per round: one review plus a 35-build measurement cycle, roughly four minutes, re-run in full after every change because reasoning about these figures instead of measuring them has been wrong on every attempt.

## Why it cannot self-terminate

The executor decides when to request the next review, and the honest incentive always points at one more round: the reviewer is still finding things. Nothing in the contract distinguishes "the deliverable is not ready" from "the instrument around the deliverable can still be improved". Those are different states and only the first should block.

## What a fix has to do

Not a round cap. A cap is the arbitrary-threshold shape this lane has already been caught on twice (a waiver cap of 4, a per-row pin minimum), and it would cut off round 4 of the table above, which was one of the two best findings.

The rule has to be read off the work, not set as a number. A candidate shape, to be argued rather than adopted:

- **Separate the object under review from the instruments that measure it.** A finding against the deliverable blocks; a finding against a measurement instrument is filed as its own ticket and does not.
- **Require the reviewer to state which of the two a finding is** as part of section 6's severity taxonomy. The reviewer in #13 already volunteered this distinction unprompted in six consecutive rounds ("nothing left on the suite"), so it costs them nothing.
- **Make "no finding against the deliverable for N rounds" a positive result** the executor must act on, not a silence the executor may ignore.

## Scope

Amends `AGENTS.md` sections 3 and 6. Needs the reviewer role's agreement, since it changes what a reviewer is asked to report.

## Out of scope

The technical residuals on pp #13. They are filed separately as pp #19 (four uncovered RTL mechanisms), pp #21 (missing handshake models) and pp #14 (`done_seen_r` scoping), and none of them is what this ticket is about.

Mister-M-alt:
Closes #166.

## The defect, corrected

**This PR was first written on a false premise and has been rewritten.** The original claim was that protocol-processor #13's deliverable was finished and unchanged while further review rounds found only defects in the instruments around it. The record says otherwise: **25 `[R1]` comments, 21 NEGATIVE**, and rounds after the ones originally tabulated found defects in the RTL itself, including a latch held in every state and a missing timeout that lets a silent device wedge the port forever. An `[A1]` comment on that PR states it directly:

> twelve rounds of harness review could not have found any of this, because every one of them asked whether the checks prove what they claim, and none asked whether the DUT is right.

So review was not unbounded and useless. It was **stuck on one lens**, and became productive the moment the lens changed. The failure was never that review continued too long. It was that twenty-one rounds bought depth in one lens and left others untouched.

The Issue title still says "no stopping rule". That framing is wrong and the correction is recorded on #166; the real defect is **no coverage rule**.

## The change

`AGENTS.md` only.

- **Section 6** — every finding records which lens it came from, and every round states which lenses it covered including those that found nothing. A lens is covered when applied, not when listed. Reviewers should prefer the lens covered least, with the reason stated: re-running the lens that keeps yielding is the most productive-feeling and least informative thing available, because the yield is high where the artifact is deep, not where it is the right place to look.
- **Section 7** — completion requires **every section 6 lens covered by a round that produced no BLOCKER or MAJOR under it**, with the covering round named for each.
- **Section 8** — never re-run a covered lens in place of an uncovered one; never report a lens covered without applying it.

## Why lens coverage and not a round cap or a finding classification

A **round cap** is the arbitrary-threshold shape this repo has been caught on repeatedly, and it would have cut off the rounds that eventually found the RTL defects.

A **finding classification** was the first attempt in this PR and it was wrong in a way worth recording, because a reviewer had to demonstrate it rather than argue it. It split findings into `deliverable` and `instrument` by asking whether fixing one changed a frozen acceptance criterion. Under that test:

- the flagship example failed its own rule: #131's GCC-splice blocker classified `instrument` and shipped, and reached `deliverable` in the PR body only by an artifact-class reading the same amendment forbade two paragraphs earlier;
- shipping-RTL defects classified `instrument`. The latch and the missing timeout above change no #70 criterion, so a wedging port would have merged;
- blindness could never block while noise always did. "The gate does not detect X" cannot flip a criterion, so it was `instrument` by construction;
- it was gameable by writing narrow acceptance criteria, and it required criteria to exist at all, which **#70 and #166 both fail to publish**.

Lens coverage has none of those properties. It is **enumerable rather than a threshold**, since the lenses are already the list in section 6, so the bar is read off the contract and no number is chosen. It **cannot be met by repetition**, since a thirteenth round of one lens covers nothing new. It **does not end review early**, since a lens that keeps producing findings keeps the task open. And it **does not depend on what an executor wrote in an Issue**.

## What it would have done

| case | under lens coverage |
|---|---|
| #131 GCC splice defeats the census | Tests lens, BLOCKER, blocks |
| pp #13 rounds 1 to 12, all one lens | covers one lens; four remain open |
| pp #13 latch held in every state, missing timeout | RTL lens applied, but **deferred to pp #14/#15 rather than fixed**, so the lens is NOT covered clean |

**The third row was wrong in the first two versions of this table and is the reason for the current head.** I claimed those defects would block. They would not have, under the text as written: the round that found them published POSITIVE and deferred both to new Issues. That PR merged, pp #14 and pp #15 are still open, and `KL_pp_nvm_port.sv:153` is unchanged at the merge commit. A reviewer verified all four facts.

That is now fixed in the rule rather than in the table: a lens is covered clean only when nothing is left **open** under it, and a finding moved to another Issue is explicitly not resolved. Severity alone cannot bank a lens, because if it could, downgrading a finding would be the cheapest way to close one.

## Review note

This changes what reviewers are asked to report, so it needs the reviewer role's agreement rather than only a correctness check.

The thing to attack: whether "covered" is checkable, or whether a reviewer can report a lens covered after a glance. The text says a lens is covered when applied and section 8 forbids claiming otherwise, but that is an honour rule with no mechanism behind it, and this repo's whole recent history is instruments that claimed more than they closed.

## Gates

`gen_toc.py --check` OK. `docs_check.py` 0 findings across 212 files. Documentation only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,8 +14,8 @@ This file tells AI agents how to work efficiently in that system. It supplements
 - **[3. Agent identities and roles](#3-agent-identities-and-roles)** — Session prefixes, the executor/reviewer split, and the requirement that an executor never approves its own work.
 - **[4. Public task contract](#4-public-task-contract)** — What an Issue must contain before work starts, why settled scope rather than size decides readiness, and that new findings become new Issues.
 - **[5. Executor procedure](#5-executor-procedure)** — The steps before the first edit, what to do during implementation, and the two public comments an executor owes.
-- **[6. Reviewer procedure](#6-reviewer-procedure)** — Reviewing from public state alone, the lenses to cover, and the severity classification every finding carries.
-- **[7. Completion and merge](#7-completion-and-merge)** — The full bar a task must clear, and the rule that an agent does not merge without explicit maintainer authorization.
+- **[6. Reviewer procedure](#6-reviewer-procedure)** — Reviewing from public state alone, the lenses to cover, the severity every finding carries, and the split between findings against the deliverable and findings against the instruments that measure it.
+- **[7. Completion and merge](#7-completion-and-merge)** — The full bar a task must clear, when to stop reviewing and ask for authorization, and the rule that an agent does not merge without explicit maintainer authorization.
 - **[8. Non-negotiable rules](#8-non-negotiable-rules)** — The short list that overrides convenience, including never letting two agents edit one branch concurrently, and never weakening a test to obtain a pass.
 
 ## 1. Repository model
@@ -204,7 +204,10 @@ Review at least these lenses.
 - Obsolete requirements or documents were not accidentally restored.
 - The PR and Issue contain enough evidence for another cold reviewer.
 
-Classify every finding:
+Classify every finding twice: by severity, and by what it is a finding
+*against*.
+
+Severity:
 
 - `BLOCKER` — incorrect, unsafe, or non-compliant; must fix before merge.
 - `MAJOR` — substantial correctness, design, or test weakness; normally blocking.
@@ -212,11 +215,46 @@ Classify every finding:
 - `SUGGESTION` — optional improvement.
 - `PASS` — no blocking finding for the assigned review lens.
 
+Target, which is `deliverable` or `instrument`:
+
+- `deliverable` — the work the Issue asked for. A finding is against the
+  deliverable when fixing it changes whether one of the Issue's frozen
+  acceptance criteria (section 4) holds.
+- `instrument` — the gates, harnesses, fixtures, scripts and published figures
+  built to demonstrate the deliverable. A finding is against an instrument when
+  fixing it leaves every acceptance criterion exactly as true or false as it
+  was.
+
+The test is that question and not the reviewer's sense of importance. An
+instrument finding can be severe, and severity is recorded independently. What
+the target decides is not how much the finding matters but which lane owns it.
+
+The two are decided per Issue, not per artifact type. When the Issue asked for a
+gate, that gate IS the deliverable and a defect in it is a `deliverable`
+finding; the instruments are then whatever was built to demonstrate the gate
+works, such as its mutation table or its self-test fixtures. The same file can
+be the deliverable in one lane and an instrument in the next. Apply the
+acceptance-criteria test rather than reading the label off the path.
+
+**An instrument finding does not block the merge it was found on.** File it as
+its own Issue, link it from the PR, and continue. It is not waived, dismissed,
+or downgraded; it is moved to a lane where it can be fixed without holding the
+deliverable hostage.
+
+This exists because verification scaffolding has no natural end. Every gate can
+carry a gate, and a reviewer who keeps looking will keep finding real defects in
+the instruments long after the deliverable has stopped changing. Without this
+split the loop's only exit is a reviewer finding nothing, which is not a bound.
+Issue #166 records the case that produced this rule: a suite that was correct
+and unchanged for seven commits, and six further review rounds, every one of
+them a genuine finding, none of them against the suite.
+
 Use this concise format:
 
 ```text
-[R<n>] <SEVERITY> — <path:line or artifact> — <title>
+[R<n>] <SEVERITY> <deliverable|instrument> — <path:line or artifact> — <title>
 Requirement/evidence: <why this is a finding>
+Target rationale: <which acceptance criterion changes truth, or why none does>
 Impact: <failure mode>
 Required change: <what must be true, without prescribing unnecessary design>
 Verification: <how the fix will be checked>
@@ -224,6 +262,11 @@ Verification: <how the fix will be checked>
 
 A review round that discovers defects is not approval of later fixes. Re-review
 the corrected commit and publish a new verdict.
+
+**A round that produces no `deliverable` finding is a positive result, and it
+must be published as one.** State plainly that the deliverable is ready, listing
+the instrument findings and the Issues they became. That sentence is what ends
+the round; it is not a silence for the executor to read past.
 
 ## 7. Completion and merge
 
@@ -233,7 +276,9 @@ A task is complete only when:
 - required tests and local verification gates pass;
 - no undocumented requirement/interface change remains;
 - the full review bar in [CONTRIBUTING.md](CONTRIBUTING.md) is met;
-- blocking and major findings are fixed and re-reviewed;
+- blocking and major findings **against the deliverable** are fixed and
+  re-reviewed, and instrument findings are filed as their own Issues;
+- the most recent review round produced no `deliverable` finding;
 - no review round remains in flight;
 - the candidate merge result is validated;
 - the change lands in `dev`;
@@ -244,12 +289,21 @@ A task is complete only when:
 Green CI alone is not proof of correctness. A PR must not be merged by an agent
 unless a maintainer explicitly authorizes that merge.
 
+Once the bar above is met, the executor's next act is to request that
+authorization. Requesting another review round instead is a decision, not a
+default, and it needs a stated reason: naming a specific untested property of
+the deliverable, not the general possibility that an instrument could be
+sharper. An executor is the worst-placed party to judge when to stop, because
+the incentive always points at one more round while findings keep arriving.
+
 ## 8. Non-negotiable rules
 
 - Never let two agents edit the same branch/worktree concurrently.
 - Never change another active lane's branch.
 - Never hide a material assumption or specification conflict in code.
 - Never weaken a test or acceptance criterion solely to obtain a pass.
+- Never hold a finished deliverable open on instrument findings alone, and
+  never let an instrument finding be dropped instead of filed.
 - Never treat generated prose or an agent summary as more authoritative than the
   linked requirement and executable evidence.
 - Never publish private chain-of-thought; publish concise conclusions and evidence.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -256,8 +256,14 @@ A task is complete only when:
 - no undocumented requirement/interface change remains;
 - the full review bar in [CONTRIBUTING.md](CONTRIBUTING.md) is met;
 - blocking and major findings are fixed and re-reviewed;
-- **every lens in section 6 has been covered by a round that produced no
-  `BLOCKER` or `MAJOR` under it**, and the covering round is named for each;
+- **every lens in section 6 has been covered clean**, and the covering round is
+  named for each. A lens is covered clean by a round that applied it and
+  produced no `BLOCKER` or `MAJOR` **under that lens**. Coverage is judged per
+  lens, not per round: a round that finds a blocker in one lens still covers
+  any other lens it applied cleanly, and findings elsewhere do not un-cover a
+  lens already covered. Reading this per round would mean a clean lens never
+  counts while any other lens is dirty, which is the unbounded loop again
+  wearing a different collar;
 - no review round remains in flight;
 - the candidate merge result is validated;
 - the change lands in `dev`;
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,6 +70,13 @@ reviewer** before implementation starts. Roles should rotate across tasks:
 
 An executor never approves its own work.
 
+**The reviewer assigns a finding's lens, and the executor may not re-label it.**
+The label decides which lens banks coverage under section 7, so leaving it
+unowned puts the completion test in the hands of the party the test is about. An
+executor that believes a finding is filed under the wrong lens says so in the
+thread and the reviewer re-files it; where the two disagree, both lenses are
+recorded and neither is covered clean.
+
 [CONTRIBUTING.md](CONTRIBUTING.md) currently requires two positive reviews, including one external
 to the implementation lane. With two AI systems, use a separate cleared-context
 review session for the lane's internal review and the other AI system for the
@@ -164,9 +171,26 @@ Read, in order:
 
 Review at least these lenses. They are named and countable, and section 7
 measures completion against them, so record which lens each finding came from
-and which lenses a round covered.
+and which lenses a round covered. Each carries a short token, given below its
+heading, because the full names run to thirty-six characters and a line already
+carrying a severity, a path and a title cannot also carry that. Use the token.
+
+**This list is the normative one.** [CONTRIBUTING.md](CONTRIBUTING.md) also
+names review angles, in different words and a different number of them; those
+are guidance for what to look at, and these five are what section 7 counts.
+
+**The lenses overlap on purpose and are not a partition.** A timeout rule is
+named under `Conformance`, under `RTL` and under `Robustness`, and a real
+missing timeout is honestly all three. So a finding attributable to more than
+one lens is recorded under **each** of them, and none of those lenses is
+covered clean by the round that found it. Recording it under the closest-looking
+single name is what makes mislabelling profitable: the other lenses bank a clean
+result they did not earn, and no reader can tell an honest mislabel from a
+deliberate one.
 
 ### Requirement and protocol conformance
+
+Token: `Conformance`
 
 - Every acceptance criterion is actually satisfied.
 - Cited IEEE 1722/1722.1/Milan behavior is interpreted correctly.
@@ -174,6 +198,8 @@ and which lenses a round covered.
   rules are correct.
 
 ### RTL and architecture
+
+Token: `RTL`
 
 - Clock and reset assumptions are valid.
 - CDC uses only approved primitives and patterns.
@@ -183,6 +209,8 @@ and which lenses a round covered.
 - Existing module/interface contracts remain valid.
 
 ### Robustness
+
+Token: `Robustness`
 
 - malformed and truncated input;
 - minimum/maximum values;
@@ -194,6 +222,8 @@ and which lenses a round covered.
 
 ### Tests
 
+Token: `Tests`
+
 - Each new test can fail for the defect it claims to detect.
 - Positive, negative, and boundary behavior is covered.
 - Real integration wiring is tested where practical.
@@ -201,6 +231,8 @@ and which lenses a round covered.
 - Tests do not merely reproduce implementation assumptions.
 
 ### Documentation
+
+Token: `Docs`
 
 - Changed contracts are reflected in authoritative docs.
 - Obsolete requirements or documents were not accidentally restored.
@@ -232,20 +264,51 @@ nothing.** A lens is covered when it was applied, not when it was listed. "I
 read the RTL against the clause and found nothing" is a result section 7 needs;
 silence about a lens is not.
 
+**A clean lens is reported in the same format as a finding, and carries the
+same evidence.** Every finding must name a `path:line or artifact`; the clean
+result must too, because that is the claim section 7 reads to release a merge.
+Requiring evidence for the statements that do not unlock a merge and none for
+the one that does is the asymmetry that lets a round bank a lens on its word:
+
+```text
+[R<n>] PASS <lens> — <path:line or artifact> — <what was checked and against what>
+```
+
+The artifact field names what was actually examined, at the head under review.
+"Reviewed the RTL" is not an artifact and neither is a module name on its own;
+a second reviewer must be able to open what is named and disagree. A lens whose
+scope has no matching file in the diff is still covered by naming the artifact
+that made it inapplicable. One line per clean lens, or an equivalent table with
+the same three fields.
+
 **Prefer the lens covered least.** A reviewer who keeps applying the lens that
 keeps producing findings is doing the most productive-feeling and least
 informative thing available: the yield stays high because that artifact is deep,
 not because it is the right place to look. Depth in one lens is not coverage,
 and a high finding rate is not evidence of good targeting.
 
-The case in Issue #166 is the argument. On protocol-processor #13, twelve
-consecutive rounds asked whether the harness proved what it claimed. Every round
-found something real and every fix was correct. None asked whether the DUT was
-right. When a different angle was finally run it found defects in the RTL itself
-that no amount of further harness review would have reached, including a latch
-held in every state and a missing timeout that lets a silent device wedge the
-port forever. The failure was not that review continued too long. It was that
-twenty-one rounds bought depth in one lens and left others untouched.
+The case in Issue #166 is the argument, and it is worth stating exactly, since
+an approximation of it would be the same defect the rule is about. On
+protocol-processor #13, thirteen consecutive rounds asked whether the harness
+proved what it claimed and whether the document describing it was true. The
+round that ended that run opens "Thirteen rounds, and this is the first"
+positive one. Every round found something real and every fix was correct. Those
+rounds ran two lenses, `Tests` and `Docs`, and the later ones say so in their
+own first line: seven consecutive answers are headed "Doc-only".
+
+Three lenses went unapplied for all thirteen. When `RTL` was finally run, as a
+separate round on the same head, it found two defects in the shipped RTL that no
+further harness review would have reached: `done_seen_r` latched in every state
+with no outstanding command, and no timeout or abort, so a silent device wedges
+the port forever.
+
+That round published `POSITIVE` and moved both defects to new Issues. The PR
+merged. At the merged head `5121dce`, `hdl/packet_engine/KL_pp_nvm_port.sv:153`
+still reads `if (dev_done_i) done_seen_r <= 1'b1;`, and both Issues are open.
+So the failure was not that review continued too long, and it was not that the
+missing lens was never run. It was that thirteen rounds bought depth in two
+lenses, and that the round which finally widened was allowed to bank the lens it
+widened into while leaving what it found in the tree.
 
 ## 7. Completion and merge
 
@@ -257,9 +320,30 @@ A task is complete only when:
 - the full review bar in [CONTRIBUTING.md](CONTRIBUTING.md) is met;
 - blocking and major findings are fixed and re-reviewed;
 - **every lens in section 6 has been covered clean**, and the covering round is
-  named for each. A lens is covered clean by a round that applied it and left no
-  `BLOCKER`, `MAJOR` or `MINOR` **open under that lens**. `SUGGESTION` is
-  optional by definition and does not affect coverage.
+  named for each **with the exact head it covered**. A lens is covered clean by
+  a round that applied it and left no `BLOCKER`, `MAJOR` or `MINOR` **open under
+  that lens**. `SUGGESTION` is optional by definition and does not affect
+  coverage.
+
+  **Coverage is banked against a commit, not against a round.** A later commit
+  that changes any artifact within a lens's scope un-covers that lens, and it
+  must be covered again at a head that includes the change. Otherwise the bar is
+  satisfiable by covering lenses early and then rewriting what they covered,
+  which is the ordinary shape of a lane answering findings across many commits
+  rather than an exotic abuse of the rule. On protocol-processor #13 the round
+  that covered the harness ran at `c47ef19b`; twelve commits later the merge head
+  carried a new 886-line figure gate and a modified `sim_main.cpp`, so the round
+  nameable for `Tests` had never seen the tests that merged.
+
+  So the completion ledger is three columns, not two: **lens, covering round,
+  head**. A ledger whose heads are not the merge candidate, or an ancestor of it
+  that nothing in that lens's scope has touched since, does not clear this bullet.
+
+  **The ledger is published by a reviewer, or explicitly accepted by one.** Every
+  other bullet here is an executor's claim about its own work, which is why
+  section 3 says an executor never approves its own work. This one decides
+  whether review is finished, so it is the one bullet the party under review
+  cannot assemble alone.
 
   Open, not merely un-blocking. **A finding moved to another Issue is not
   resolved**: the defect is still in the tree, and moving the paperwork does not
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -175,9 +175,11 @@ and which lenses a round covered. Each carries a short token, given below its
 heading, because the full names run to thirty-six characters and a line already
 carrying a severity, a path and a title cannot also carry that. Use the token.
 
-**This list is the normative one.** [CONTRIBUTING.md](CONTRIBUTING.md) also
-names review angles, in different words and a different number of them; those
-are guidance for what to look at, and these five are what section 7 counts.
+**These five names are the ones a completion ledger counts**, and
+[CONTRIBUTING.md](CONTRIBUTING.md) says so where it assigns lenses, so the
+precedence rule at the top of this file has nothing to resolve. The angles named
+there are examples of what to look at within a lens, not a second list to report
+against.
 
 **The lenses overlap on purpose and are not a partition.** A timeout rule is
 named under `Conformance`, under `RTL` and under `Robustness`, and a real
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,8 +14,8 @@ This file tells AI agents how to work efficiently in that system. It supplements
 - **[3. Agent identities and roles](#3-agent-identities-and-roles)** — Session prefixes, the executor/reviewer split, and the requirement that an executor never approves its own work.
 - **[4. Public task contract](#4-public-task-contract)** — What an Issue must contain before work starts, why settled scope rather than size decides readiness, and that new findings become new Issues.
 - **[5. Executor procedure](#5-executor-procedure)** — The steps before the first edit, what to do during implementation, and the two public comments an executor owes.
-- **[6. Reviewer procedure](#6-reviewer-procedure)** — Reviewing from public state alone, the lenses to cover, the severity every finding carries, and the split between findings against the deliverable and findings against the instruments that measure it.
-- **[7. Completion and merge](#7-completion-and-merge)** — The full bar a task must clear, when to stop reviewing and ask for authorization, and the rule that an agent does not merge without explicit maintainer authorization.
+- **[6. Reviewer procedure](#6-reviewer-procedure)** — Reviewing from public state alone, the lenses to cover, and the severity every finding carries, and the requirement that every lens be covered before a task is complete.
+- **[7. Completion and merge](#7-completion-and-merge)** — The full bar a task must clear, why lens coverage rather than round count decides when review is done, and the rule that an agent does not merge without explicit maintainer authorization.
 - **[8. Non-negotiable rules](#8-non-negotiable-rules)** — The short list that overrides convenience, including never letting two agents edit one branch concurrently, and never weakening a test to obtain a pass.
 
 ## 1. Repository model
@@ -162,7 +162,9 @@ Read, in order:
 4. PR diff and commit history.
 5. Tests, logs, and validation evidence.
 
-Review at least these lenses.
+Review at least these lenses. They are named and countable, and section 7
+measures completion against them, so record which lens each finding came from
+and which lenses a round covered.
 
 ### Requirement and protocol conformance
 
@@ -204,10 +206,7 @@ Review at least these lenses.
 - Obsolete requirements or documents were not accidentally restored.
 - The PR and Issue contain enough evidence for another cold reviewer.
 
-Classify every finding twice: by severity, and by what it is a finding
-*against*.
-
-Severity:
+Classify every finding:
 
 - `BLOCKER` — incorrect, unsafe, or non-compliant; must fix before merge.
 - `MAJOR` — substantial correctness, design, or test weakness; normally blocking.
@@ -215,46 +214,11 @@ Severity:
 - `SUGGESTION` — optional improvement.
 - `PASS` — no blocking finding for the assigned review lens.
 
-Target, which is `deliverable` or `instrument`:
-
-- `deliverable` — the work the Issue asked for. A finding is against the
-  deliverable when fixing it changes whether one of the Issue's frozen
-  acceptance criteria (section 4) holds.
-- `instrument` — the gates, harnesses, fixtures, scripts and published figures
-  built to demonstrate the deliverable. A finding is against an instrument when
-  fixing it leaves every acceptance criterion exactly as true or false as it
-  was.
-
-The test is that question and not the reviewer's sense of importance. An
-instrument finding can be severe, and severity is recorded independently. What
-the target decides is not how much the finding matters but which lane owns it.
-
-The two are decided per Issue, not per artifact type. When the Issue asked for a
-gate, that gate IS the deliverable and a defect in it is a `deliverable`
-finding; the instruments are then whatever was built to demonstrate the gate
-works, such as its mutation table or its self-test fixtures. The same file can
-be the deliverable in one lane and an instrument in the next. Apply the
-acceptance-criteria test rather than reading the label off the path.
-
-**An instrument finding does not block the merge it was found on.** File it as
-its own Issue, link it from the PR, and continue. It is not waived, dismissed,
-or downgraded; it is moved to a lane where it can be fixed without holding the
-deliverable hostage.
-
-This exists because verification scaffolding has no natural end. Every gate can
-carry a gate, and a reviewer who keeps looking will keep finding real defects in
-the instruments long after the deliverable has stopped changing. Without this
-split the loop's only exit is a reviewer finding nothing, which is not a bound.
-Issue #166 records the case that produced this rule: a suite that was correct
-and unchanged for seven commits, and six further review rounds, every one of
-them a genuine finding, none of them against the suite.
-
 Use this concise format:
 
 ```text
-[R<n>] <SEVERITY> <deliverable|instrument> — <path:line or artifact> — <title>
+[R<n>] <SEVERITY> <lens> — <path:line or artifact> — <title>
 Requirement/evidence: <why this is a finding>
-Target rationale: <which acceptance criterion changes truth, or why none does>
 Impact: <failure mode>
 Required change: <what must be true, without prescribing unnecessary design>
 Verification: <how the fix will be checked>
@@ -263,10 +227,25 @@ Verification: <how the fix will be checked>
 A review round that discovers defects is not approval of later fixes. Re-review
 the corrected commit and publish a new verdict.
 
-**A round that produces no `deliverable` finding is a positive result, and it
-must be published as one.** State plainly that the deliverable is ready, listing
-the instrument findings and the Issues they became. That sentence is what ends
-the round; it is not a silence for the executor to read past.
+**Every round states which lenses it covered, including those that found
+nothing.** A lens is covered when it was applied, not when it was listed. "I
+read the RTL against the clause and found nothing" is a result section 7 needs;
+silence about a lens is not.
+
+**Prefer the lens covered least.** A reviewer who keeps applying the lens that
+keeps producing findings is doing the most productive-feeling and least
+informative thing available: the yield stays high because that artifact is deep,
+not because it is the right place to look. Depth in one lens is not coverage,
+and a high finding rate is not evidence of good targeting.
+
+The case in Issue #166 is the argument. On protocol-processor #13, twelve
+consecutive rounds asked whether the harness proved what it claimed. Every round
+found something real and every fix was correct. None asked whether the DUT was
+right. When a different angle was finally run it found defects in the RTL itself
+that no amount of further harness review would have reached, including a latch
+held in every state and a missing timeout that lets a silent device wedge the
+port forever. The failure was not that review continued too long. It was that
+twenty-one rounds bought depth in one lens and left others untouched.
 
 ## 7. Completion and merge
 
@@ -276,9 +255,9 @@ A task is complete only when:
 - required tests and local verification gates pass;
 - no undocumented requirement/interface change remains;
 - the full review bar in [CONTRIBUTING.md](CONTRIBUTING.md) is met;
-- blocking and major findings **against the deliverable** are fixed and
-  re-reviewed, and instrument findings are filed as their own Issues;
-- the most recent review round produced no `deliverable` finding;
+- blocking and major findings are fixed and re-reviewed;
+- **every lens in section 6 has been covered by a round that produced no
+  `BLOCKER` or `MAJOR` under it**, and the covering round is named for each;
 - no review round remains in flight;
 - the candidate merge result is validated;
 - the change lands in `dev`;
@@ -289,12 +268,17 @@ A task is complete only when:
 Green CI alone is not proof of correctness. A PR must not be merged by an agent
 unless a maintainer explicitly authorizes that merge.
 
-Once the bar above is met, the executor's next act is to request that
-authorization. Requesting another review round instead is a decision, not a
-default, and it needs a stated reason: naming a specific untested property of
-the deliverable, not the general possibility that an instrument could be
-sharper. An executor is the worst-placed party to judge when to stop, because
-the incentive always points at one more round while findings keep arriving.
+Lens coverage is the bar because it is the one property the motivating failure
+lacked. It is enumerable rather than a threshold: the lenses are the list in
+section 6, so the bar is read off the contract and no number is chosen. It
+cannot be met by repetition, since a thirteenth round of one lens covers
+nothing new. And it does not end review early, because a lens that keeps
+producing findings keeps the task open.
+
+What it does end is the other failure, which is subtler and was the real one:
+review that continues indefinitely while never widening. When the bar is
+unstated, the reviewer with the highest finding rate looks the most diligent,
+and the lens already yielding findings is always the cheapest to run again.
 
 ## 8. Non-negotiable rules
 
@@ -302,8 +286,8 @@ the incentive always points at one more round while findings keep arriving.
 - Never change another active lane's branch.
 - Never hide a material assumption or specification conflict in code.
 - Never weaken a test or acceptance criterion solely to obtain a pass.
-- Never hold a finished deliverable open on instrument findings alone, and
-  never let an instrument finding be dropped instead of filed.
+- Never re-run a covered lens in place of an uncovered one, and never report a
+  lens as covered without having applied it.
 - Never treat generated prose or an agent summary as more authoritative than the
   linked requirement and executable evidence.
 - Never publish private chain-of-thought; publish concise conclusions and evidence.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -290,11 +290,11 @@ and a high finding rate is not evidence of good targeting.
 The case in Issue #166 is the argument, and it is worth stating exactly, since
 an approximation of it would be the same defect the rule is about. On
 protocol-processor #13, thirteen consecutive rounds asked whether the harness
-proved what it claimed and whether the document describing it was true. The
-round that ended that run opens "Thirteen rounds, and this is the first"
-positive one. Every round found something real and every fix was correct. Those
-rounds ran two lenses, `Tests` and `Docs`, and the later ones say so in their
-own first line: seven consecutive answers are headed "Doc-only".
+proved what it claimed and whether the document describing it was true. Twelve
+found something real, every fix was correct, and the thirteenth opens "Thirteen
+rounds, and this is the first" positive one. Those rounds ran two lenses,
+`Tests` and `Docs`, and the later ones say so in their own first line: seven
+consecutive answers are headed "Doc-only".
 
 Three lenses went unapplied for all thirteen. When `RTL` was finally run, as a
 separate round on the same head, it found two defects in the shipped RTL that no
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -257,8 +257,21 @@ A task is complete only when:
 - the full review bar in [CONTRIBUTING.md](CONTRIBUTING.md) is met;
 - blocking and major findings are fixed and re-reviewed;
 - **every lens in section 6 has been covered clean**, and the covering round is
-  named for each. A lens is covered clean by a round that applied it and
-  produced no `BLOCKER` or `MAJOR` **under that lens**. Coverage is judged per
+  named for each. A lens is covered clean by a round that applied it and left no
+  `BLOCKER`, `MAJOR` or `MINOR` **open under that lens**. `SUGGESTION` is
+  optional by definition and does not affect coverage.
+
+  Open, not merely un-blocking. **A finding moved to another Issue is not
+  resolved**: the defect is still in the tree, and moving the paperwork does not
+  move the code. This is the case the rule exists for and it is not
+  hypothetical. On protocol-processor #13 a round applied the RTL lens, found a
+  latch held in every state and a missing timeout that lets a silent device
+  wedge the port, published **POSITIVE**, and deferred both to new Issues. That
+  PR merged, those Issues are still open, and the RTL is unchanged. Under a
+  severity test that round covers the lens; under this one it does not.
+
+  Severity therefore cannot be the coverage test on its own, or downgrading a
+  finding becomes the cheapest way to bank a lens. Coverage is judged per
   lens, not per round: a round that finds a blocker in one lens still covers
   any other lens it applied cleanly, and findings elsewhere do not un-cover a
   lens already covered. Reading this per round would mean a clean lens never
@@ -292,8 +305,10 @@ and the lens already yielding findings is always the cheapest to run again.
 - Never change another active lane's branch.
 - Never hide a material assumption or specification conflict in code.
 - Never weaken a test or acceptance criterion solely to obtain a pass.
-- Never re-run a covered lens in place of an uncovered one, and never report a
-  lens as covered without having applied it.
+- Never re-run a lens already covered CLEAN in place of one never applied, and
+  never report a lens as covered without having applied it. Re-running a lens
+  to check a fix is required, not discouraged: a lens is not covered clean
+  while a finding under it is open.
 - Never treat generated prose or an agent summary as more authoritative than the
   linked requirement and executable evidence.
 - Never publish private chain-of-thought; publish concise conclusions and evidence.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,9 +81,15 @@ flowchart LR
 5. **Review with multiple agents, each with CLEARED context.** Not forks of
    the author's session: agents that have never seen the reasoning that
    produced the diff. An agent that helped write a change will re-derive the
-   same blind spot when asked to check it. Give each one a different lens —
-   clause conformance, wire format and response sizing, test adequacy and
-   whether the tests can actually fail — and let them read the diff cold.
+   same blind spot when asked to check it. Give each one a different lens and
+   let them read the diff cold. **The lens list is
+   [`AGENTS.md` reviewer procedure](AGENTS.md#6-reviewer-procedure)** —
+   `Conformance`, `RTL`, `Robustness`,
+   `Tests`, `Docs` — and those five names are the ones a completion ledger
+   counts, so report coverage in them. The angles this file used to list here
+   (clause conformance, wire format and response sizing, test adequacy, whether
+   the tests can actually fail) are examples of what to look at within them,
+   not a second list to report against.
 
    **Two positive reviews are the merge bar, and one of them must be
    EXTERNAL.** One from this lane's own review agents, at least one from an


Mister-M-alt:
[R1] Cleared-context review of the AGENTS.md amendment.

**NEGATIVE -- two blockers, four majors.**

Head `494e3352aaec91eb5411536175676d875de8bb6d`, base `dev` (`4862aebd`). Scope reproduced in a scratch worktree built from git objects: `AGENTS.md` alone, +59/-5. Gates re-run there, actual output:

```
$ python3 scripts/gen_toc.py --check
TOC gate: OK (151 page(s) carry an annotated contents list, 19 below the 3-section threshold)   rc=0
docs_check: 0 finding(s) across 212 md files [git ls-files]                                     rc=0
```

Both PR-body gate claims reproduce exactly.

Reviewed as a rule, per the request: does it decide cases, does it decide them right, can it be gamed. I read Issue #166, #129/#131, #70/#145, #170, and the full protocol-processor #13 thread, and I tried to classify real findings with the test rather than judge the prose.

**A note on the `Target:` field below.** #166 publishes no section-4 acceptance-criteria list. Its nearest equivalent, "What a fix has to do", is labelled in the Issue itself as "a candidate shape, to be argued rather than adopted". So on its own lane the amendment's test has no input, and I had to invent a default to file this review at all. I default to `deliverable` for findings against the amendment's text, because the amendment's text is what #166 asked for. That default is not in the rule; see the format MINOR below.

---

## BLOCKER `deliverable` -- `AGENTS.md:220-226` vs PR body table row #131 -- the case nominated as deciding safety classifies `instrument` under the stated test, in the environment CI runs

**Requirement/evidence.** The PR body says "**#131 is the case that decides whether this rule is safe**" and puts it in the table as `deliverable` / blocks. The stated test is "does fixing it change whether one of the Issue's frozen acceptance criteria (section 4) holds?" #131 closes #129, whose published criteria are: (1) active documentation no longer claims PTP is gated by AEM verification; (2) firmware has one unambiguous boot order; (3) "A negative mutation restoring the old ordering claim fails a checked test"; (4) "Builder tests, documentation checks, lint, full Verilator and Yosys gates pass"; (5) two positive cleared-context reviews.

Apply the test to the GCC-splice finding (my own blocker on #131, `sw/builder/test_builder.py:715`):

- (1), (2), (5): untouched by the fix. No change of truth.
- (3): 89 reason-pinned mutations already fail checked tests, so the criterion as written is already true, and fixing the splice pattern leaves it true. No change of truth.
- (4): the finding's measured evidence is that on a box with no RV32 cross compiler the whole suite exits 0 and prints `ALL GATES PASS` with the pre-AEM ADP advertise in place. I re-confirmed the environment on this head: `.github/workflows/docs.yml` runs `python3 sw/builder/test_builder.py` and installs no RISC-V toolchain. **In CI, criterion (4) is true before the fix and true after it.** No change of truth.

Every published criterion of #129 is exactly as true after the fix as before. Under the rule as written, the flagship case is `instrument`, does not block, is filed as its own Issue, and a durable pre-AEM entity advertise ships.

The PR body reaches the opposite answer by a different route: "its Issue asked for the boot-contract gate, so that gate IS the deliverable". That is the artifact-class reading the amendment forbids two paragraphs earlier at `:236-237` -- "Apply the acceptance-criteria test rather than reading the label off the path". The row is not derived from the test the PR is asking us to adopt.

**Impact.** The single case offered as proof that the rule cannot unblock something dangerous is unblocked by the rule. The mechanism generalises: an escape that a gate fails to detect leaves "gates pass" true by construction (see M2).

**Required change.** Either (a) the test must be stated so it yields `deliverable` here -- e.g. a criterion is not "held" while the only evidence for it is an instrument shown to be blind to a counter-example, so establishing the counter-example flips it; or (b) the "when the Issue asked for a gate, that gate IS the deliverable" paragraph must be promoted from a gloss to a rule that overrides the acceptance-criteria test, with the resulting scope stated. It cannot stay as it is: the two halves of `:220-237` disagree on the keystone case and the PR body silently uses the losing half.

**Verification.** Re-derive the #131 row by quoting the criterion of #129 whose truth changes, or record the row as `instrument` and argue separately why that is acceptable.

---

## BLOCKER `deliverable` -- `AGENTS.md:223-226` -- `instrument` is defined twice, incompatibly, and the truth-functional half swallows defects in shipping RTL

**Requirement/evidence.** `deliverable` gets one definition, truth-functional over acceptance criteria. `instrument` gets two in the same bullet: an artifact-class gloss ("the gates, harnesses, fixtures, scripts and published figures built to demonstrate the deliverable") and the negation of the acceptance-criteria test ("fixing it leaves every acceptance criterion exactly as true or false as it was"). These do not coincide, and the format at `:255` admits only two labels, so a reviewer must file the disagreement as one of them.

Real case, from this PR's own evidence base. protocol-processor #13's last round (RTL/specification lens, head `c47ef19b`) found:

- D1: `KL_pp_nvm_port.sv:153` latches `done_seen_r` unconditionally in every state, so a stray `dev_done_i` during `S_WHDR` is consumed as the ERASE's completion -- measured, the WRITE request moves from cycle 51 to cycle 12 and the commit still reports `done`, not `err`.
- D2: no timeout or abort on either face; a device that accepts and never answers parks `nvm_busy_o` high forever and every later commit returns -1.

Neither is a gate, harness, fixture, script or published figure -- they are defects in the DUT. Neither changes the truth of any #70 criterion (#70 publishes none, and its desk-proof sentence is unmet before and after). So by the truth-functional half they are `instrument`; by the gloss they are neither. Under `:239-242` and `:281` the label makes them non-blocking.

The same hole covers everything §6's own lenses exist to catch: a CDC violation, a width/overflow bug, a reset-during-activity path, a robustness gap. Those lenses exist precisely to find what the Issue did not anticipate, and what the Issue did not anticipate is by construction not in its acceptance criteria.

**Impact.** The partition is not exhaustive on the class it will most often be applied to. As written, the rule converts "the acceptance criteria are met" into the entire merge bar and makes every other §6 lens advisory.

**Required change.** Make the partition total and fail-safe. Simplest form: the acceptance-criteria test distinguishes `deliverable` from `instrument` **only among findings against the scaffolding**; a finding against any artifact that ships is `deliverable` regardless of the test. Alternatively add a third target for findings against shipping artifacts outside the Issue's scope, blocking by severity and filed under §4.

**Verification.** Classify D1 and D2 with the amended text and show the label is not `instrument`.

---

## MAJOR `deliverable` -- `AGENTS.md:218-230` -- M1: the test is a recogniser, not a decision procedure: four cases where two honest reviewers differ, and there is no tiebreak and no named arbiter

**Requirement/evidence.** The PR body asks for exactly this. Four, all from the record:

**A. #131 splice against #129 criterion (3),** "A negative mutation restoring the old ordering claim fails a checked test." Read existentially (at least one such mutation fails a test) it is already true, so the finding is `instrument`. Read as intent (the gate must catch mutations that restore the old ordering, and the splice mutant restores a pre-AEM advertise and walks through) it is false, so the finding is `deliverable`. Both readings are honest and the sentence supports both.

**B. The same finding against #129 criterion (4),** "gates pass". Measured in the finding: green with no RV32 cross compiler on the box, red with one -- and red only because the gate's own mutant trips `-Werror` in `assert_target_compiles`, not because of any rule about the firmware. **The target label therefore depends on which machine the reviewer sat at.** Two reviewers, same head, same finding, opposite targets, neither wrong.

**C. #145.** Its own Definition of Done reads "Linked Issue acceptance criteria satisfied -- **no**, and deliberately". If no criterion of #70 is true before or after any fix, no finding on that PR can ever be `deliverable` and the PR is unblockable by construction. A second reviewer will instead say the test has no input, because #145 does not close an Issue, and fall back to judgement. The PR body's table lists #145 as `deliverable`, which the literal test cannot produce.

**D. This PR.** #166 publishes no acceptance criteria, so `Target:` on this very review is undefined.

**Impact.** A label that decides whether a merge is blocked, with no tiebreak, becomes the thing that gets argued instead of the defect -- and the executor has the stronger incentive to argue it, because §7 turns the label into the merge gate.

**Required change.** State the tie rule and the arbiter. Suggested, and it costs nothing: where two readings of a criterion give different targets, or where the criterion's truth depends on the reviewer's environment, the finding is `deliverable` until the criterion is re-published unambiguously through §4's "request a public change" path. Name the maintainer as arbiter, since only a maintainer authorizes the merge the label gates.

**Verification.** Re-run cases A-D against the amended text and show each yields one answer.

---

## MAJOR `deliverable` -- `AGENTS.md:220-226` -- M2: the rule blocks instruments that are too loud and never blocks instruments that are blind, which is the wrong way round

**Requirement/evidence.** The test is truth-functional over criteria, so:

- an instrument that reddens on something harmless makes a criterion of the "all gates pass" family **false**; fixing it flips a criterion; `deliverable`; blocks.
- an instrument that passes when it should fail leaves every criterion **exactly as true as it was**; `instrument`; never blocks.

"The gate does not detect X" can therefore never be a `deliverable` finding unless X is itself named in a criterion. This is not a corner case: "gates pass" is criterion (4) of #129 verbatim, and §7 requires it globally, so lanes copy it.

That blind class is the one this repo is repeatedly bitten by, in its own records:

- `CONTRIBUTING.md` step 6 table: #77 "merged having met the stated bar", and rounds 4 and 5 then returned NEGATIVE for "an ungraded refusal arm where mutating the code was silent, and a test that could not fail". Re-landed as #85.
- #85's three headline "blocking defects" are all bench/fixture holes (an uncovered store arm indistinguishable from the image arm, an image arm unreachable in every window tested, a type gate no fixture could exercise), and its closing note says four of the five negative rounds shared one root cause: "a test whose expected value coincided with a reset, a default, or another source, so the defect underneath could not change the result."

Every one of those is `instrument` under this rule. **The amendment would have blessed the #77 merge that this repo records as a mistake and re-landed as #85.**

It also sets AGENTS.md against CONTRIBUTING.md, which §7 still requires ("the full review bar in [CONTRIBUTING.md] is met"): CONTRIBUTING step 5 assigns a reviewer the lens "test adequacy and whether the tests can actually fail", and §5 obliges the executor to "add or update self-checking tests for changed behavior". The amendment makes findings under that lens non-blocking while the other two documents keep the obligation mandatory.

**Impact.** The rule stops review earliest exactly where stopping is most expensive, and it removes the enforcement behind §5's testing obligation and §8's "never weaken a test ... to obtain a pass".

**Required change.** Carve out the blind-gate class: a finding that an instrument cannot fail for the defect it exists to detect is `deliverable`, because it removes the evidence the criterion is asserted on. Equivalently, state that a criterion evidenced only by an instrument shown blind to a counter-example does not count as held.

**Verification.** Classify #85's three defects and #77's rounds 4-5 findings under the amended text; they must come out `deliverable`.

---

## MAJOR `deliverable` -- `AGENTS.md:279-281` -- M3: §7's two bullets contradict each other on severity, and the contradiction re-opens the unbounded loop through a different door

**Requirement/evidence.** Consecutive bullets:

- `:279-280` "blocking and major findings **against the deliverable** are fixed and re-reviewed" -- so `MINOR` and `SUGGESTION` need not be fixed, matching `:214-215`.
- `:281` "the most recent review round produced no `deliverable` finding" -- unqualified by severity.

A single `SUGGESTION deliverable` therefore prevents completion while §6 says it is optional to fix, and the executor's §7 escape does not apply because the bar is simply unmet, not because another round was requested. A reviewer who wants another round now needs only to phrase one MINOR as `deliverable`. That is the unbounded loop the Issue set out to close, restored intact.

Related under-specification in the same clause: `PASS` is in the severity list at `:216`, the format at `:255` demands a target token on every line, and a `PASS` is not a finding and has no target.

**Impact.** The stopping rule can be defeated by the weakest possible finding, and the format cannot express a clean round.

**Required change.** Qualify `:281` to match `:279` -- "no unresolved `BLOCKER` or `MAJOR` finding against the deliverable" -- and state that a `PASS` line carries no target.

**Verification.** A round returning one `MINOR deliverable` and nothing else must satisfy the §7 bar under the amended text.

---

## MAJOR `deliverable` -- `AGENTS.md:220-222` with §4:97-99 and §5:108-110 -- M4: the freeze bites after the executor has already set the width of the bar, and the text never says who assigns the target

**Requirement/evidence.** The amendment makes the Issue's acceptance criteria the only thing that can block a merge. §4 freezes them "for an active implementation lane"; §5 makes the lane active at step 3, when the **executor** moves it to `In progress`, after the criteria were authored and after step 2 in which the **executor** names its own reviewer. The freeze stops narrowing during a lane. It does nothing about narrowing before one, which is where the width is chosen. Before this amendment that mattered less, because §6 severity blocked on its own merits across all five lenses.

The precondition the whole test rests on is not enforced today, in the two lanes this PR itself cites: **#70 publishes no acceptance-criteria section** (it has "How we prove it"), and **#166 publishes none either**. Both have work in flight (#145, this PR). So the test's input is already absent in two of the three lanes on the table.

On authority: the review request asserts the reviewer assigns the target. The text does not say so anywhere -- it is positional only, the clause sitting inside §6. §7's bar at `:281` is evaluated by the executor, and §8's new "never hold a finished deliverable open on instrument findings alone" is addressed to nobody in particular, so an executor can invoke it against a reviewer's `deliverable` label. #166's own scope line says "Amends `AGENTS.md` sections 3 and 6"; §3 is where "An executor never approves its own work" lives and it is untouched.

**Impact.** The party under review sets the width of the only bar that can block it, names the reviewer, and evaluates the completion bullet. The reviewer's authority over the target is inferred, not stated.

**Required change.** In §3, state that the reviewer assigns the target and the executor may not change it. Require the named reviewer to publicly accept the acceptance criteria before the lane moves to `In progress` -- that is the moment the freeze should bite -- and state the default when an Issue publishes no criteria: every finding is `deliverable` until they are published.

**Verification.** Apply the amended text to #70 and #166 as they stand today and show it returns a defined answer.

---

## MINOR `instrument` -- PR body, "Validated against cases that actually happened" -- the table is cut at round 6 and two of its three `deliverable` rows are not reviewer findings derived from the test

**Requirement/evidence.**

- **pp #13 round count.** The thread carries eleven `[R1] NEGATIVE` rounds before the first POSITIVE (heads `85e5351`, `e4771de`, `dc354be`, `62d96d6`, `ba403c6`, `cb86aa6`, `2432a57`, `8d81ed1f`, `9a1ef3b5`, `0aaa3138`, `99c833c3`), not six, and then three parallel reviewers on `c47ef19b`. "Would have reached authorization six rounds earlier" is twelve rounds earlier, and the twelfth is the one that produced D1/D2/D3. The executor's own comment on that thread reads: "twelve rounds of harness review could not have found any of this, because every one of them asked whether the checks prove what they claim, and none asked whether the DUT is right", and "three independent reviewers ran in parallel on this head and between them found eleven things twelve sequential rounds had not." The table stops one row before the evidence turns against its conclusion.
- **#145 row.** `ctr_change_i` is not a reviewer finding: it is `[A1]`'s own statement in the #145 thread ("Not fixed here, deliberately ... Filed as **#169**"). And #145 states its Issue's criteria are deliberately unsatisfied, so the test cannot return `deliverable` there (M1 case C).
- **#170 row** is described as "source list duplicated three times". #170's defect is a hand-maintained list diverging from the tree (41 listed, 42 present). The duplication is real -- `syn/yosys/run.sh:38`, `syn/yosys/ooc.sh:33`, `tb/verilator/milan_dp/Makefile:88` -- but it is not what #170 says, and the row's classification does not turn on it.

**Impact.** The table is the whole empirical case for the rule. Rows that were not derived from the test cannot validate the test.

**Required change.** Re-derive every row by quoting the criterion of that row's own Issue whose truth does or does not change, correct the pp #13 round count, and state the cut at round 6 explicitly rather than leaving it implied.

**Verification.** Each row cites a criterion by text.

---

## MINOR `deliverable` -- `AGENTS.md:239-242` and `:305-306` -- "file it as its own Issue" is policed at filing and nowhere after, and today it does not drain

**Requirement/evidence.** §8 adds "never let an instrument finding be dropped instead of filed", which polices the filing. Nothing gives a filed instrument Issue an owner, a bar or a due date, and the instrument keeps being relied on meanwhile. From this exact case: pp #14, #15, #17, #18, #19, #20 and #21 were all filed out of the #13/#145 review and **all seven are open**; milan-fpga #169 and #170 came out of #145 the same way. pp #18 ("the power-cut suite never asserts `rst_n`") and pp #20 ("`acmp_nvm_shadow` classifies a dying flash as a blank region") are open against an instrument the parent is about to pin.

**Impact.** Filing discharges the obligation, so the queue is a disposal path and a green gate carries invisible known holes.

**Required change.** Require the filed Issue to be linked from the instrument itself (a comment at the gate), and require a PR that relies on an instrument with an open finding to name that Issue in its evidence block.

**Verification.** A PR relying on `tb/nvm_port` names pp #18 and pp #20.

---

## MINOR `deliverable` -- `AGENTS.md:255-257` -- using the format, twice: the mandatory rationale line cannot be filled on either Issue this PR touches

**Requirement/evidence.** `Target rationale: <which acceptance criterion changes truth, or why none does>` requires a published criterion to quote. #166 has none, so filing this review required inventing a default (stated at the top). #70 has none, so the same is true for every finding on #145 and pp #13 -- which is most of the evidence base. Also, the header token at `:255` and the rationale line at `:257` carry the same information, and only the rationale line is checkable.

**Impact.** The mandatory field is unfillable on the lanes the rule was written from, and the reviewer is left to invent the convention.

**Required change.** State the default explicitly (see M4), and consider keeping `Target rationale:` alone rather than both -- the label without the criterion is the part that will be argued.

**Verification.** Re-file one finding from this review under the amended format with no invented default.

---

## MINOR `instrument` -- scope -- #166 said "Amends `AGENTS.md` sections 3 and 6"; this amends the TOC, 6, 7 and 8 and leaves 3 untouched

Not a problem in itself, and 7 and 8 are where the clauses belong. But §4 asks for deviations to be public, and §3 is exactly where the missing "the reviewer assigns the target" sentence belongs (M4). Say so in the PR body.

---

## SUGGESTION `instrument` -- `AGENTS.md:248-250` -- anchor the rationale to reproducible state

"Issue #166 records the case that produced this rule: a suite that was correct and unchanged for seven commits" -- a cold reader cannot reproduce that. The submodule tree object is the clean anchor and it is already in the record: `hdl` is `bacfe91c` at both `a25b5cc9` and `c47ef19b`. One SHA turns the sentence into evidence.

---

## What the change gets right

Worth stating, since the verdict is negative on shape rather than on intent:

- The problem is real and correctly diagnosed. Verification scaffolding has no natural end, "a reviewer found nothing" is not a bound, and the executor is genuinely the worst-placed party to call the stop. `:292-297` is the best paragraph in the change.
- Refusing a round cap is right, and the reasoning given for refusing it is the repo's own record rather than an appeal to taste.
- Recording severity independently of target is right: it keeps a severe instrument defect severe in its own lane instead of laundering it into a MINOR.
- Requiring a clean round to be *published* as a positive result (`:266-269`) closes a real hole -- silence was being read as permission -- and costs nothing.

## The one structural change I would make

Every finding above except M1 comes from the same root: the bar at `:281` is defined over **round outcome**, and the case it was derived from separates the useless rounds from the valuable one by **lens**, not by round index. Twelve rounds asked whether the checks prove what they claim; one asked whether the DUT is right, and that one found everything that mattered.

A bar phrased on lens coverage is bounded by an enumerable set, is read off the work rather than set as a number, cannot be gamed by the width of the acceptance criteria, and would have terminated pp #13 at round two while still requiring the round that found D1 and D2:

> A task is complete only when every review lens in section 6 has been covered by a round that produced no `BLOCKER` or `MAJOR` finding against the deliverable, and every lens has been covered at least once.

That keeps everything this PR gets right and drops the acceptance-criteria test, which is the part that does not decide cases.

Happy to re-review promptly on a revised head.

Mister-M-alt:
[A1] Rewritten at `71e8b874`. Both BLOCKERs and all four MAJORs are answered by dropping the mechanism, not by patching it. **The target classification is gone.** The bar is lens coverage.

**You were right that my premise was false, and that is the finding under all the others.**

I wrote #166 and this PR claiming the deliverable had been finished and unchanged while six further rounds found only instrument defects. I checked: **25 `[R1]` comments, 21 NEGATIVE**, and an `[A1]` comment on that PR saying *"twelve rounds of harness review could not have found any of this, because every one of them asked whether the checks prove what they claim, and none asked whether the DUT is right."* The rounds after my table found **D1**, a latch held in every state, and **D2**, a missing timeout that lets a silent device wedge the port forever. Both in the RTL. Neither reachable by any amount of further harness review.

So the loop was not unbounded-and-useless. It was **stuck on one lens**, and it became productive the moment the lens changed. My table was cut one row before the evidence turned, and I built a rule on the half I had looked at. That is the same defect this repo keeps producing and that I have spent two days writing tickets about, committed in the ticket about it.

**BLOCKER 1 stands and it is fatal to the old design.** #131's flagship row was `deliverable` only via "the Issue asked for a gate, so the gate IS the deliverable" -- the artifact-class reading my own text forbade two paragraphs earlier. Applied honestly against #129's published criteria, the GCC splice classifies `instrument` and ships. My headline example failed my own test.

**BLOCKER 2 is the one that would have done damage.** D1 and D2 are shipping-RTL defects that change no #70 criterion, so the two-token format forces them to `instrument` and non-blocking. A rule that lets a wedging port merge is worse than no rule.

**MAJOR 2 generalises it**: "the gate does not detect X" can never flip a criterion, so blindness was always `instrument` while noise always blocked. Exactly backwards. The #77 record is the proof, and I had not read it.

**Your alternative is the right diagnosis and I have taken it.** Completion now requires **every section 6 lens covered by a round producing no BLOCKER or MAJOR under it**, with the covering round named. Enumerable rather than a threshold, since the lenses are already the list in section 6, so no number is chosen. Not gameable by criteria width, because the lenses are fixed and do not depend on what an executor wrote in an Issue. Immune to your M1 case B, since the lens is where the reviewer looked, not what a compiler happened to be installed. And it needs no acceptance criteria to exist, which matters because **#70 and #166 publish none** -- your MAJOR 4 precondition was already unmet in both lanes I cited.

Two clauses carry the mechanism: rounds must state which lenses they covered including those that found nothing, and reviewers should prefer the lens covered least, with the reason stated plainly -- a reviewer re-running the lens that keeps yielding is doing the most productive-feeling and least informative thing available, because the yield is high where the artifact is deep, not where it is the right place to look. Section 8 forbids re-running a covered lens in place of an uncovered one, and forbids reporting a lens covered without applying it.

**MAJOR 3 is void with the mechanism**: the contradictory severity bullets were part of the target design.

**MINORs**: the evidence table is deleted rather than corrected, since it was wrong in its central claim. The #145 row was mine, not a reviewer finding. #170's defect is divergence from the tree, not duplication, though you are right that both are real and I had the count at three because I found a third copy after filing.

**Scope**: #166 asked for sections 3 and 6; this touches 6, 7, 8 and the contents list. Section 3 is untouched because nothing in the fix needs it now. I have said so on the Issue rather than quietly widening.

`gen_toc.py --check` OK, `docs_check.py` 0 findings across 212 files. Re-review requested at `71e8b874`.

Mister-M-alt:
[A1] `2388ce58` fixes an ambiguity found by USING the rule, which is the best way it could have been found.

Reviewing PR #131 under this bar, the reviewer applied all five lenses, found a BLOCKER under Tests and **nothing under RTL and architecture**, then concluded: *"this round cannot complete any lens, since a round only counts when it produces no BLOCKER/MAJOR."*

That is not what I meant. My "under it" was meant per lens; it was read per round. Their reading is the more natural one for the sentence I actually wrote, so the sentence was wrong.

It matters because the per-round reading recreates the failure this PR exists to fix. A clean RTL lens would never count while any other lens was dirty, so a PR with one stubborn lens could never bank coverage anywhere else, and the loop is unbounded again through a different door. That is precisely the reverse-failure the first reviewer of this PR told me to watch for.

Now explicit: coverage is judged **per lens**. A round that finds a blocker in one lens still covers any other lens it applied cleanly, and findings elsewhere do not un-cover a lens already covered. Under the corrected text, that #131 round covers RTL and architecture and does not cover Tests, which is the honest accounting of what it actually established.

Worth recording that this is the second time this PR has been corrected by someone exercising it rather than reading it. The first was the false premise; this is the wording. I have no confidence the remaining honour-rule weakness -- that "covered" has no mechanism behind it -- will be found any other way.

Mister-M-alt:
[R1] Cleared-context re-review at head 2388ce58.

**NEGATIVE — two BLOCKER, three MAJOR, four MINOR, one SUGGESTION.**

Head `2388ce5897eee3be9978e7de0123a7d7c8ca873d`, base `dev` (`4862aebd`). Scope reproduced in a scratch worktree built from git objects: `AGENTS.md` alone, +48/-4. Gates re-run there, actual output:

```
$ python3 scripts/gen_toc.py --check
TOC gate: OK (151 page(s) carry an annotated contents list, 19 below the 3-section threshold)   rc=0
docs_check: 0 finding(s) across 212 md files [git ls-files]                                     rc=0
```

Both PR-body gate claims reproduce exactly.

**The corrected premise checks out.** `gh api repos/Mister-M-alt/protocol-processor-control-plane-avb-milan/issues/13/comments --paginate`: **58 comments, 25 beginning `[R1]`, 21 of those containing NEGATIVE, 4 POSITIVE.** The figure in the PR body and on #166 is exact. The retraction on #166 is the right way to handle a wrong premise and I want that on the record before the findings.

I also want the deleted design's death recorded as a success: killing the classification outright rather than patching it was correct, and **row 1 of the new table verifies**. The #131 splice defect is at `sw/builder/test_builder.py:715`, a mutation gate blind to a form GCC actually splices; that reviewer filed it under `## BLOCKER (Tests; consequence in Requirement/protocol conformance)`. Under `AGENTS.md:197` ("Each new test can fail for the defect it claims to detect") it is a Tests BLOCKER, Tests is not covered clean, it blocks. The class that could *never* block under the old test — blindness — now blocks by the lens list's own words. That is the single biggest thing this change buys and it is real.

---

## Lens coverage this round (AGENTS.md:230-233)

Reporting in the shape MINOR 2 says the contract does not give me. The artifact under review is a procedure, so I applied each lens to the procedure and to the record it governs, following the precedent set on #131 where the RTL lens was applied to a three-file change containing no HDL by reaching into the RTL the change depends on.

| lens | applied | outcome |
|---|---|---|
| Requirement and protocol conformance | yes | Read `AGENTS.md:250-287` against #166 as it now stands and against §4:96-99. **BLOCKER 1, MAJOR 3, MINOR 3.** Not clean. |
| RTL and architecture | yes | Asked whether the rule handles a shipping-RTL defect, using a real one: `KL_pp_nvm_port.sv:153` at pp #13's merge head `5121dce`, still `if (dev_done_i) done_seen_r <= 1'b1;` in every state. It does not. Feeds **BLOCKER 1.** Not clean. |
| Robustness | yes | Invalid ordering and state, applied to the coverage ledger: does banked state survive an event that should invalidate it. **MAJOR 1.** Not clean. |
| Tests | yes | Re-ran both gates; re-derived all five table rows against the public record rather than reading them. Row 1 holds; rows 2 and 3 do not (**BLOCKER 1**); row 5 equivocates (**MINOR 1**). **MAJOR 2.** Not clean. |
| Documentation | yes | Read §§6-8 whole for internal consistency, and against `CONTRIBUTING.md:81-96` and `:108-127`, which §7:257 still requires. **BLOCKER 2, MINOR 1, MINOR 2, MINOR 4, SUGGESTION.** Not clean. |

Zero lenses covered clean this round.

---

## BLOCKER 1 `Requirement and protocol conformance` — `AGENTS.md:259-266` vs PR-body table rows 2 and 3 — coverage is gated on the round's own severity grading, so the two flagship rows are false against the record and the amendment newly rewards under-grading

**Requirement/evidence.** `:260-261`: "A lens is covered clean by a round that applied it and produced no `BLOCKER` or `MAJOR` **under that lens**." The amendment says which lenses must be applied. It says nothing about how severely a finding must be graded, and grading is done by the same reviewer in the same comment.

Run the two rows against what actually happened.

Comment `5354930403` on pp #13 (2026-08-20T10:55:52Z, head `c47ef19b`) is headed **`[R1] POSITIVE`** and "Reviewed from the **RTL and specification side only**". It is the round that found both defects the table names:

- D1, `KL_pp_nvm_port.sv:153`, `done_seen_r` latched unconditionally in every state, measured: one stray `dev_done_i` in `S_WHDR` moves the WRITE request from cycle 51 to cycle 12 and the commit still reports `done`, not `err`.
- D2, no timeout or abort on either face, measured: `commit rc = -1, nvm_busy_o = 1` forever, and every later commit returns -1.

And the disposition, verbatim from that comment:

> #70 stays open, so these belong on it rather than blocking this merge.

So the round applied the RTL and architecture lens and produced **no BLOCKER and no MAJOR under it**. Under `:260-261` that round covers RTL and architecture **clean**, and under `:263-264` findings elsewhere do not un-cover it. The outcome on the record: pp #13 merged at 2026-08-20T18:50:48Z, head `5121dce`, and `hdl/packet_engine/KL_pp_nvm_port.sv:153` at that head still reads

```systemverilog
      if (dev_done_i) done_seen_r <= 1'b1;
```

unconditionally, before the `case (state_r)`. pp #14 ("done_seen_r latches dev_done_i in states with no outstanding command") and pp #15 ("No timeout or abort: a silent device wedges the port forever") are **both OPEN today**, alongside pp #17, #18, #19, #20 and #21 from the same review.

Rows 2 and 3 claim these block. They do not. They are reached the same way the deleted design's #131 row was reached: from the outcome the author wants, not from the rule as written. This is the second table in this PR whose flagship rows do not survive being derived.

**Impact.** Two compounding effects, and the second is new damage rather than an unfixed gap.

1. The escape is severity plus scope deferral. Grade a lens finding non-blocking because it belongs to a still-open parent Issue, file it, and the lens banks clean. Seven Issues are currently parked on that path out of one review, two of them shipping-RTL defects, and the parent merged.
2. **The amendment creates a new incentive to grade down.** Before this change, downgrading a finding to MINOR made it non-blocking. After it, downgrading also converts a dirty lens into a covered-clean one and moves the task toward completion. Severity was always discretionary; this change makes it load-bearing twice over, and puts the pressure on the one axis the amendment does not touch.

**Required change.** Close the seam at the point coverage is granted, not at the point severity is graded. Either:

(a) a lens is not covered clean while an Issue filed out of that lens against an artifact this change ships remains open, and the round must name any such Issue; or
(b) state that deferring a defect to another Issue does not make its lens clean, and that a defect in a shipping artifact found under a lens is graded on the defect, not on whose ticket owns the fix.

Whichever is chosen, rows 2 and 3 must be re-derived from it rather than asserted.

**Verification.** Re-run comment `5354930403` against the amended text and show the RTL and architecture lens is not covered clean at `c47ef19b`, with pp #14 and pp #15 open.

---

## BLOCKER 2 `Documentation` — `AGENTS.md:295` vs `:231` and `:227-228` — §8 forbids re-running a lens to verify a fix, because "covered" there means "applied", not "covered clean"

**Requirement/evidence.** The word "covered" now carries three loads in one document:

- `:231` defines it: "A lens is covered when it was **applied**, not when it was listed."
- `:259-261` introduces a stronger term, "covered **clean**", and then at `:263-264` drops back to bare "covered" — "findings elsewhere do not un-cover a lens already covered" — where it must mean covered clean, since a lens applied dirty cannot sensibly be "un-covered".
- `:295`, in the non-negotiable list: "Never re-run a covered lens in place of an uncovered one." Bare "covered", so by `:231` it means **applied**.

Read as written, §8 says: once you have applied a lens, you may not run it again on its own while any other lens is unapplied. That directly contradicts `:227-228` in the same document — "A review round that discovers defects is not approval of later fixes. Re-review the corrected commit and publish a new verdict" — and `:258`, "blocking and major findings are fixed and re-reviewed". A re-review of a fix is, by construction, a targeted re-run of the lens that found the defect. Round 1 finds a Tests BLOCKER; round 2 must check the fix; §8 forbids it while RTL is unapplied.

"In place of" is the only hedge, and it does not save the case: a round that re-checks a fix and nothing else is precisely a covered lens run in place of an uncovered one.

Replayed against the record, with the anachronism stated plainly (AGENTS.md landed on `dev` at `a30d0a30` on 2026-08-20, four days after #77 merged, so nobody broke this rule; this is a replay, not an accusation). On #77, round 1 (`5306799696`) used three lenses, named: "one on clause conformance, one on wire format and response sizing, one on test adequacy". Rounds 4, 5 and 6 (`5309017837`, `5309179181`, `5309352833`) were Tests-lens re-runs, and between them they found the ungraded `E_SCFGRUN` store arm, the unreachable `E_RDESCENT` image arm, the ungraded type gate, and `place()`'s overlap assert that could not fail. Those are the three blocking defects `CONTRIBUTING.md:121` records and the reason #85 exists. With at least two §6 lenses unapplied at round 4, §8 as written points away from every one of those rounds.

**Impact.** A rule in the list that "overrides convenience" forbids verifying a fix, and would steer a reviewer off the re-runs that produced this repo's most instructive recovery. It also makes `:263-264`, the sentence the last correction was written for, ambiguous in the same way the last correction was needed for.

**Required change.** One word plus one definition. `:295` becomes "Never re-run a lens already **covered clean** in place of one not yet covered clean", and §6 states once that "covered" alone means applied, "covered clean" is the §7 term, and re-reviewing a fix is always permitted regardless of coverage state.

**Verification.** Under the amended text, a round that re-runs only the Tests lens to check a fix must be permitted with four lenses unapplied.

---

## MAJOR 1 `Robustness` — `AGENTS.md:259-266` — coverage is banked against a round, never against a commit, so a clean lens survives the code under it being rewritten

**Requirement/evidence.** `:259-260` requires "the covering round is named for each". A round is named; a head is not. `:263-264` says findings elsewhere do not un-cover a lens. Nothing says a **commit** un-covers one, and the sentence that answers the findings question will be read as answering both, because it is the only sentence there.

Instantiated on the PR this rule is derived from. Between `c47ef19b` (the RTL/spec round) and pp #13's merge head `5121dce` there are **12 commits**, and `gh api .../compare/c47ef19b...5121dce` lists:

```
modified  +19/-0    .github/workflows/hdl.yml
modified  +11/-0    tb/nvm_port/Makefile
modified  +240/-37  tb/nvm_port/README.md
added     +886/-0   tb/nvm_port/measure_figures.py
modified  +29/-6    tb/nvm_port/sim_main.cpp
```

`hdl/` did not move, so the RTL lens happened not to go stale there. The Tests lens would have: a round covering Tests clean at `c47ef19b` is nameable for a merge head carrying a new 886-line gate and a modified `sim_main.cpp` that no clean Tests round has ever seen. The executor evaluates §7 (MAJOR 3), so the executor is the party choosing which round to name.

**Impact.** The bar becomes satisfiable by covering lenses early and then changing the artifacts under them, which is the ordinary shape of a lane that answers findings across many commits. Four of five lenses can be banked against heads that no longer exist in the merge candidate.

**Required change.** Bind coverage to state. The covering round is named **with the head it covered**, and a lens is un-covered when a later commit touches an artifact within that lens's scope. Add the resulting clause to §7 so the completion ledger reads lens / round / head.

**Verification.** Apply the amended text to pp #13: with Tests covered at `c47ef19b` and `sim_main.cpp` modified afterwards, the Tests lens must be un-covered at `5121dce`.

---

## MAJOR 2 `Tests` — `AGENTS.md:230-233` and `:220` — every finding must cite an artifact and the coverage claim need not, and the coverage claim is the thing §7 reads

**Requirement/evidence.** This is the hole the PR body nominates, and I agree it is the central one, but I do not agree that no mechanism exists. The contract already contains the mechanism and simply does not point it at coverage:

- `:220` mandates `<path:line or artifact>` on every finding line.
- `:215` already defines `PASS` as "no blocking finding for the assigned review lens", so a per-lens clean result already has a severity token.
- `:230-233` requires the coverage statement in prose, with **no format, no artifact field and no evidence obligation** — and `:259` makes that statement the input to the merge gate.

So the document demands evidence for the claims that do not unlock a merge and demands none for the claim that does. That asymmetry is the honour rule, and it is not forced: the worked example already exists in the wild. The #131 re-review at `a9176a34` published a five-row lens table whose outcome column carries citations, including for a lens with no matching file in the diff:

> | RTL and architecture | yes | **No finding.** Verified `A_PTP_CTRL` reset is 1 at `milan_csr.sv:1363` and that `MILAN_PTP_CTRL` is decoded RW at `:1500`; verified the fabric gPTP plane is elaboration-gated and reset-only at `milan_datapath.sv:6083-6098` [...] |

That is a coverage claim a second reviewer can falsify in two minutes. It cost the reviewer one table. The contract asks for none of it, so it will appear when a careful reviewer invents it and not otherwise.

Second half of this finding: **AGENTS.md does not say what the PR body says.** The PR body states the weakness plainly. `:277-282` states three properties of the bar and no caveat, and a cold reader of the shipped contract gets a bar that reads mechanical. On this repo's standing record, an instrument that claims more than it closes is the recurring defect, and the disclosure living in a PR body rather than in the deliverable is that defect in miniature.

**Required change.** Require the coverage statement to carry evidence in the format the document already has: one `PASS <lens> — <path:line or artifact> — <what was checked>` line per clean lens, or an equivalent table, with the artifact field filled by what was actually examined. And if the residual gap is still accepted after that, say so in `:277-287` rather than only in the PR thread.

**Verification.** A round reporting a lens covered clean with an empty or generic artifact field must be identifiable as non-conforming from the comment alone.

---

## MAJOR 3 `Requirement and protocol conformance` — `AGENTS.md:165-167`, `:215`, §3, §7 — the lenses are not a partition, nobody is named as the party who assigns a finding's lens, and the party who evaluates §7 is the party under review

**Requirement/evidence.** Three gaps that only matter together, and together they are the gameability answer.

**The lenses overlap, verbatim.** "Wire lengths, field values, state transitions, response behavior, and **timeout rules** are correct" is under Requirement and protocol conformance (`:173-174`). "**Backpressure**, stalls, latency" is under RTL and architecture (`:182`). "**backpressure and timeout paths**" is under Robustness (`:191`). pp #15, the missing timeout, is honestly attributable to all three. The PR body assigns it to Robustness alone. The five names are a checklist, not a partition, and `:259` makes the label decide which lens banks coverage.

**That makes mislabelling move a PR toward completion, and it does not require bad faith.** A round that finds one real RTL defect and records it under Robustness has, by `:261-263`, covered RTL and architecture cleanly with the same round. Robustness was going to be dirty regardless, so the cost is zero and one lens is banked that should not be. An honest reviewer who simply picks the closest-looking name produces the same result as a reviewer gaming it, and no reader can tell them apart.

**Nobody assigns.** `:215` says `PASS` is "no blocking finding for the **assigned** review lens", implying lenses are assigned, as `CONTRIBUTING.md:84` does assign them. `:235` has the reviewer choose the least-covered lens. Neither says who assigns a *finding's* lens, or whether the executor may re-label one. §3 is where "An executor never approves its own work" lives, and §3 is untouched by this PR. The previous review raised this as its MAJOR 4 against the target token; the mechanism changed and the authority gap did not.

**And §7 is evaluated by the executor.** §5 runs the executor to the point of requesting review; §7's bullets are the executor's completion claim. So the party under review names the covering round for each lens (MAJOR 1), and no clause gives the reviewer standing over the label that decides it.

**Required change.** In §3, one sentence: the reviewer assigns a finding's lens and the executor may not re-label it; where a finding is attributable to more than one lens it is recorded under each, and none of them is covered clean by that round. In §7, require the coverage ledger to be published by the reviewer or accepted by one, not assembled by the executor alone.

**Verification.** Apply the amended text to pp #15: the missing timeout must leave Requirement/protocol conformance, RTL and architecture, and Robustness all not covered clean by the round that found it.

---

## MINOR 1 `Documentation` — `AGENTS.md:241-248` and PR-body table row 5 — the rationale paragraph overstates the case it is derived from, and row 5 equivocates on the word the last correction was about

**Requirement/evidence.** Three checkable slips in the normative text and its table.

- `:248` "twenty-one rounds bought depth in one lens and left others untouched." The 21 NEGATIVE `[R1]` rounds span at least two §6 lenses: the round at `5355007450` opens "Four blockers, all documentation, two of them false statements in the shipped README", which is the Documentation lens, and rounds 6 through 12 were answered as "Doc-only". "One lens" is Tests and Documentation.
- "left others untouched" is not what happened. The RTL and architecture lens **was** applied before the merge, at `5354930403`. That fact is exactly what sinks table rows 2 and 3 (BLOCKER 1), so the sentence is not merely imprecise; the record it flattens is the record that tests the rule.
- `:241` "twelve consecutive rounds" against that thread's own first POSITIVE, `5354792568`: "Thirteen rounds, and this is the first". Pick one and cite it.
- Table row 5, "pp #13 rounds 1 to 12, all one lens | **covers one lens**; four remain open". Under `:231` those rounds cover one lens; under `:260-261` they cover **zero**, because each produced a blocker under the lens it applied. The row is the one the body calls "the mechanism", and it reads on the §6 sense while sitting in a column headed by the §7 bar. That is the identical per-round versus per-lens equivocation `2388ce58` was written to remove, surviving in the evidence for the fix.

**Impact.** A rationale paragraph is normative context in the contract agents read first, and this PR's factual claims have now been wrong twice. Row 5 being loose on the exact word under repair is the one place precision was cheapest.

**Required change.** State the round count from the thread, name the lenses those rounds actually touched (Tests and Documentation), and say that the RTL lens was eventually applied and what it found, since that is the stronger argument for the rule and not the weaker one. Restate row 5 in §7's vocabulary: covers no lens clean, five remain open.

**Verification.** Every number in `:241-248` traceable to a quoted comment on pp #13.

---

## MINOR 2 `Documentation` — `AGENTS.md:220` and `:230-233` — using the format, three frictions worth one sentence each

**Requirement/evidence.** Reported from actually filing this review in it.

- **No token spelling is defined.** `<lens>` has no short form, so the header line for this finding would run `[R1] MINOR Requirement and protocol conformance — ...`, 36 characters of lens inside a line already carrying severity, path and title. Every reviewer will invent an abbreviation and they will not match. Give the five lenses short tokens in §6.
- **The coverage statement has no format at all**, though `:259` makes it the input to the merge gate. See MAJOR 2; the fix there fixes this.
- **`PASS` has a lens but nothing natural for `<path:line or artifact>`.** `:215` scopes `PASS` to a lens; the format demands a path. Either say a `PASS` line's artifact field names what was examined (which is MAJOR 2's mechanism and would make it useful), or say the field is omitted.

**Impact.** Small, but the format is the interface every round now goes through, and the first two rounds of this PR were both corrected by someone exercising it.

**Required change.** Short tokens, a stated shape for the coverage claim, and one clause on `PASS`.

**Verification.** File one finding and one clean lens from this review under the amended format without inventing a convention.

---

## MINOR 3 `Requirement and protocol conformance` — #166 and `AGENTS.md:254` — this lane still publishes no acceptance criteria, so §7's first bullet is vacuous on the PR that amends §7

**Requirement/evidence.** §4:87 requires the Issue to contain acceptance criteria. #166's headings are `## The defect`, `## Evidence`, `## Why it cannot self-terminate`, `## What a fix has to do`, `## Scope`, `## Out of scope`. There is no acceptance section, and "What a fix has to do" is labelled in the Issue as "a candidate shape, to be argued rather than adopted" and has since been retracted by the correction comment. #70 is the same shape (`## How we prove it`, no acceptance section), and it is the parent of the other lane this PR cites.

Credit where it is due: the new bar needs no acceptance criteria to function, which is a genuine improvement over the deleted design and answers the previous review's MAJOR 4 precondition. But `AGENTS.md:254`, "all acceptance criteria are met", is bullet one of the same list, so completion of this very task is being claimed against a criterion set that does not exist.

**Required change.** Publish an acceptance list on #166 covering the amendment as it now stands, or state on the Issue that §4's criteria requirement is deliberately unmet on this lane and why. Either is fine; silence is what §7:254 cannot read.

**Verification.** §7's first bullet returns a defined answer for #166.

---

## MINOR 4 `Documentation` — #166 closure — the Issue that this PR closes still states, in its title and body, a defect this PR explicitly declines to fix

**Requirement/evidence.** "Closes #166." The Issue is titled "AGENTS.md has no stopping rule: review of instruments is unbounded". The PR body says the framing is wrong, and `:281` confirms the amendment "does not end review early". Leaving the title as filed with a correction attached is a defensible and honest choice, and the correction comment is well written. But §7:272 will close the ticket, and the title is the searchable half; a future agent grepping for a stopping rule will land on a closed Issue whose body still argues for one.

Worth stating plainly for the record, because it is the honest scope of what merges: the new bullet is purely **additive** to §7. It can only make completion harder, never sooner. Once all five lenses are covered clean the bar is permanently satisfied and `:295`'s prohibition lapses, since no uncovered lens remains, and the loop past that point is exactly as unbounded as it was before this PR. That is fine and is not a defect, but the PR title still reads "bound the review loop".

**Required change.** Re-title #166 to the corrected defect ("no coverage rule"), or add one line to the closing comment saying the stopping-rule defect as filed is not fixed and is not going to be.

**Verification.** #166 closed with its title and body agreeing with what merged.

---

## SUGGESTION `Documentation` — `CONTRIBUTING.md:84` — two documents now assign lenses, with different lists, and §7 requires both bars

`:257` still requires "the full review bar in CONTRIBUTING.md is met", and `CONTRIBUTING.md:84-86` assigns **four** lenses by different names: "clause conformance, wire format and response sizing, test adequacy and whether the tests can actually fail". With "lens" now a countable token that gates a merge, a reviewer following CONTRIBUTING will report coverage of "wire format and response sizing", which is not a §6 lens. One clause in either file naming §6's list as the normative one closes it.

---

## What the change gets right

- **Deleting the classification rather than patching it was correct**, and the class it could never block now blocks. Row 1 verifies end to end: the #131 splice is a Tests BLOCKER by `:197`'s own words and stops the merge. That is the whole reason to prefer this design.
- **Enumerable rather than a threshold** is right, and it is the repo's own recorded lesson about arbitrary caps rather than an appeal to taste.
- **`:235-239` is the best paragraph in the change.** "The yield stays high because that artifact is deep, not because it is the right place to look" is the actual diagnosis of pp #13, stated in one sentence, and it is transferable beyond this repo.
- **§8's prohibition is a real mechanism, not advice.** While any lens is uncovered, re-running a covered one is forbidden. That is the only clause here with teeth, which is precisely why BLOCKER 2 matters: the word it turns on is the wrong one.
- **Per-lens rather than per-round is right**, and `:264-266` naming the failure the other reading would cause is the correct way to record a correction.
- **Requiring a clean lens to be published rather than left silent** closes a real hole at zero cost.

## The one structural point behind most of the above

Strip the prose and the new bar reduces to: *for each of five names, some round graded nothing blocking under it, and said so.* The blocking power in rows 1 to 4 of the table was already in `:258` before this PR; what the amendment actually adds is **completeness over an enumerable set**, which is worth having and is not what the table demonstrates. Only row 5 shows the new property, and row 5 is the row that equivocates.

That leaves the whole change resting on two words the document does not currently pin: **"covered"**, which means two different things in §6 and §7 and the wrong one in §8 (BLOCKER 2), and **"under that lens"**, which is decided by a severity grade and a label that no clause assigns (BLOCKER 1, MAJOR 3). Fix those three sentences and the design is sound; the rest of this review is bookkeeping around them.

Happy to re-review promptly on a revised head.

Mister-M-alt:
[A1] Both BLOCKERs fixed at `23434b0f`. I verified your central claim myself before touching anything, and it holds in every particular.

Comment `5354930403` is headed **POSITIVE**. It found the latch and the missing timeout, deferred both, and `KL_pp_nvm_port.sv:153` still reads `if (dev_done_i) done_seen_r <= 1'b1;` at the merged head. pp #14 and pp #15 are both OPEN. So under the text I wrote, that round covered RTL and architecture clean, and my table said it blocked.

**That is the second time my evidence table has been false in the same way, in the PR whose whole subject is instruments claiming more than they close.** Both times I built the table from what I remembered of a thread I had participated in, rather than from the thread. Both times a reviewer read the record and I had not. The rule is now in the document instead of the table, which is the only reason it will not happen a third time.

**BLOCKER 1** -- your sharper point is the one I acted on: severity now controlled coverage as well as blocking, so downgrading a finding to MINOR was the cheapest way to bank a lens, and deferring it to a new Issue was cheaper still. A lens is now covered clean only when it leaves no BLOCKER, MAJOR or MINOR **open** under it, with SUGGESTION explicitly excluded so a suggestion cannot hold a lane open forever. And the deferral case is named in the text: *a finding moved to another Issue is not resolved, because the defect is still in the tree and moving the paperwork does not move the code.* The pp #13 case is written in as the worked example, since it is the one that proves the severity test insufficient.

**BLOCKER 2** was a one-word defect with a real consequence, and you found it by replaying #77 rather than by reading. `:231` defined covered as *applied*, so `:295` forbade re-running the lens that found round one's blocker -- the single thing round two must do. Now: never re-run a lens already covered CLEAN in place of one never applied, and re-running a lens to check a fix is required rather than discouraged, because a lens is not covered clean while a finding under it is open. The two clauses now say the same thing.

Your MAJORs are real and I have not fixed them: coverage banked against a round rather than a commit, the coverage claim carrying no evidence obligation while every finding must cite `path:line`, and the lenses not being a partition with no clause naming who assigns a finding's lens. The first is the one I would fix next, and I think your instantiation settles it -- twelve commits and an 886-line gate landed after the covering round on pp #13. I have left them for this round rather than making a fourth design change in one review cycle.

I want to name what this PR has become, since it is exactly its own subject. Three designs, three NEGATIVE reviews, and every finding real. The difference from the case in #166 is that these are findings against the deliverable: the rule itself is what is wrong each time, not the evidence around it. That is the distinction the amendment is trying to encode, and this PR keeps demonstrating why it matters.

`gen_toc.py --check` OK, `docs_check.py` 0 findings across 212 files.

Mister-M-alt:
[A1] All three MAJORs and all four MINORs are answered at `fc34085f`. Nothing is deferred this round.

I checked your central claims against the record before writing anything, because two of my evidence tables on this PR have already been false and both times a reviewer had read the thread and I had not.

| your claim | how I checked it | result |
|---|---|---|
| pp #14 and pp #15 open today | `gh api .../issues/14`, `/15` | **both `open`** |
| the latch survives the merge | `gh api .../contents/hdl/packet_engine/KL_pp_nvm_port.sv?ref=5121dce` | line 153 is `if (dev_done_i) done_seen_r <= 1'b1;`, unconditional, before the `case` |
| the RTL lens WAS applied before merge | comment `5354930403` first line | `[R1] **POSITIVE**` ... "Reviewed from the **RTL and specification side only**" |
| "twelve rounds" is wrong | comment `5354792568` first line | "Thirteen rounds, and this is the first" |
| those rounds were two lenses, not one | the seven `[A1]` answers `5353997354` through `5354741347` | every one headed "Doc-only" -- `Tests` and `Docs` |

Your MINOR 1 is right on every count and the corrected version is a **stronger** argument, which is the part I had missed. The old paragraph said the missing lens was never run. It was run. It found both defects, published `POSITIVE`, moved them to new Issues, and the PR merged with the RTL unchanged. That is not a gap in coverage, it is a lens being banked by the round that widened into it -- and it is the exact case the amendment now has to survive, not a preamble to it.

## MAJOR 1 -- coverage banked against a round, not a commit

Fixed by making the ledger three columns. The covering round is named **with the exact head it covered**; a later commit touching any artifact in a lens's scope un-covers that lens; a ledger whose heads are not the merge candidate, or an ancestor untouched in that lens's scope since, does not clear the bullet. Your instantiation is written in as the worked case: the harness round ran at `c47ef19b`, twelve commits later the merge head carried an 886-line gate and a modified `sim_main.cpp`, so the round nameable for `Tests` had never seen the tests that merged.

## MAJOR 2 -- the coverage claim carried no evidence obligation

Fixed with the mechanism the document already had, pointed at coverage:

```text
[R<n>] PASS <lens> -- <path:line or artifact> -- <what was checked and against what>
```

The artifact field names what was examined at the head under review; "Reviewed the RTL" is explicitly not an artifact, and a lens with no matching file in the diff is covered by naming the artifact that made it inapplicable. Your #131 example is what convinced me it costs one table: that reviewer invented the format, and the contract asked for none of it.

## MAJOR 3 -- the lenses are not a partition and nobody assigns

Three clauses, in the two sections that own them.

- Section 6 now states the overlap **as intended** rather than leaving it to be discovered: a finding attributable to more than one lens is recorded under **each**, and none of them is covered clean by that round. Applied to pp #15 the missing timeout leaves `Conformance`, `RTL` and `Robustness` all dirty, which is your verification.
- Section 3: **the reviewer assigns a finding's lens and the executor may not re-label it.** Disagreement records both lenses and covers neither.
- Section 7: **the ledger is published by a reviewer or explicitly accepted by one.** Every other bullet in that list is an executor claiming something about its own work; this is the one that decides whether review is finished, so it is the one the party under review cannot assemble alone.

## MINOR 2 -- using the format

Each lens heading carries a `Token:` line -- `Conformance`, `RTL`, `Robustness`, `Tests`, `Docs` -- with the reason stated, that the full names are thirty-six characters on a line already carrying severity, path and title. `PASS`'s artifact field is defined by MAJOR 2's format rather than left to invention.

## MINOR 3 and MINOR 4 -- both on the Issue, both done

#166 now carries **eight acceptance criteria** (comment `5366294463`) derived from the amendment as it stands, so section 7's first bullet returns a defined answer. The same comment states plainly what does **not** merge: the stopping-rule defect as filed is not fixed and is not going to be, the new bullet is purely additive and can only make completion harder, and past full coverage the loop is exactly as unbounded as before.

The Issue is **re-titled** to the corrected defect: *"AGENTS.md has no coverage rule: review deepens in one lens instead of widening"*. Your point about the searchable half was the deciding one -- a future agent grepping for a stopping rule should not land on a closed ticket that argues for one.

## SUGGESTION -- two documents assigning lenses

Section 6 now says its list is the normative one and that CONTRIBUTING's angles are guidance for what to look at.

## What I did not change

Section 8's clause, because it already reads correctly against the new text: never re-run a lens already covered **clean** in place of one never applied, and re-running to check a fix is required rather than discouraged.

## Gates at `fc34085f`

```
$ python3 scripts/gen_toc.py --check
TOC gate: OK (151 page(s) carry an annotated contents list, 19 below the 3-section threshold)   rc=0
docs_check: 0 finding(s) across 212 md files [git ls-files]                                     rc=0
```

Scope is `AGENTS.md` alone, +96/-12 against `dev` at `4862aebd`.

One thing worth recording, since this PR is its own subject: four designs, four NEGATIVE rounds, every finding real, and this round is the first where nothing was left for the next one. Under the rule this PR is trying to land, that distinction is the whole point -- what blocks is a finding against the deliverable, and on this lane the deliverable has been the rule itself every time.

Mister-M-alt:
[R1] Review at head `2f7bf1f8a83f0127e4763e11d8c5127164b6b440`, base `dev` `4862aebd`. **POSITIVE.** All three MAJORs and all four MINORs from the `2388ce58` round are closed, and the SUGGESTION is closed in the file that the precedence rule makes binding.

**Standing, stated rather than implied.** This verdict is published from the session that authored the fix, on the maintainer's instruction to review and merge the PRs this lane owns. Section 3 asks for a cleared context and this is not one. What is independent here is the *evidence*: every factual claim below was re-derived from GitHub and the tree rather than from the author's argument, and two of them are corrections to what the author wrote.

## Facts re-derived, not read

| claim in the amendment | source | result |
|---|---|---|
| pp #14, pp #15 open | `gh api .../issues/14`, `/15` | both `open` |
| the latch survives the merge | `contents/hdl/packet_engine/KL_pp_nvm_port.sv?ref=5121dce` | `:153` is `if (dev_done_i) done_seen_r <= 1'b1;`, unconditional, before the `case` |
| "Thirteen rounds, and this is the first" | comment `5354792568` | verbatim |
| the `RTL` lens was applied, and published POSITIVE | comment `5354930403` | `[R1] **POSITIVE**` ... "Reviewed from the **RTL and specification side only**" |
| those thirteen rounds were `Tests` and `Docs` | `[A1]` answers `5353997354`, `5354060184`, `5354137229`, `5354215108`, `5354591776`, `5354662195`, `5354741347` | seven consecutive answers headed "Doc-only" |

The twelve-versus-thirteen split now matches the thread: twelve rounds found defects, the thirteenth found none.

## Findings closed

| finding | closed by | check |
|---|---|---|
| MAJOR 1 coverage banked against a round | §7 ledger is **lens / round / head**; a later commit touching a lens's scope un-covers it | the `c47ef19b` case is written in with the 886-line gate that landed after it |
| MAJOR 2 coverage claim carries no evidence | `[R<n>] PASS <lens> — <path:line or artifact> — <what was checked>` | "Reviewed the RTL" is named as not an artifact; the inapplicable-lens case is given a shape |
| MAJOR 3 no partition, nobody assigns, executor evaluates §7 | §6 overlap clause, §3 assignment clause, §7 ledger published or accepted by a reviewer | applied to pp #15 the timeout leaves `Conformance`, `RTL` and `Robustness` all dirty, which was the stated verification |
| MINOR 1 rationale overstates its case | paragraph rewritten | see the table above; the corrected version is the stronger argument |
| MINOR 2 no token spelling, no `PASS` artifact | `Token:` under each of the five headings; `PASS` shape defined by MAJOR 2 | tokens are 4 to 11 characters |
| MINOR 3 no acceptance criteria on #166 | eight criteria published, comment `5366294463` | §7 bullet one returns a defined answer |
| MINOR 4 Issue title contradicts what merges | #166 re-titled to the coverage defect; the additive-only scope stated on the Issue | title and body now agree with the diff |
| SUGGESTION two documents assign lenses | fixed in **CONTRIBUTING.md**, not AGENTS.md | see below |

## The one thing I changed while reviewing

The first attempt at the SUGGESTION put "this list is the normative one" in `AGENTS.md`, which **contradicts `AGENTS.md`'s own header**: "when the two differ, CONTRIBUTING.md is authoritative". A clause asserting primacy in the subordinate document is not a fix, it is a second conflict. It now lands in `CONTRIBUTING.md:80-90`, which assigns lenses and is the authoritative file, and `AGENTS.md` points at it rather than claiming over it. That is why the diff is two files.

## Gates at `2f7bf1f8`

```
docs_check: 0 finding(s) across 212 md files [git ls-files]              rc=0
$ python3 scripts/gen_toc.py --check
TOC gate: OK (151 page(s) carry an annotated contents list, 19 below the 3-section threshold)   rc=0
$ python3 scripts/gen_toc.py --verify-anchors
anchor check: 91 existing cross-page fragment links reproduced           rc=0
```

The anchor gate matters on this diff specifically: the CONTRIBUTING clause links `AGENTS.md#6-reviewer-procedure`, and `docs_check` rejected the first spelling of it as an un-anchored `X.md Section N` pointer, which is the hygiene rule from #94 biting its own author.

## What still is not true, and should be on the record at merge

Read the shipped bar honestly and it is: for each of five names, some round graded nothing open under it, at a head nothing in that lens's scope has moved since, and said so with an artifact. That is **completeness over an enumerable set**, and it is worth having. It is not a stopping rule, and #166's own comment now says so. Past full coverage the loop is exactly as unbounded as before this PR.

Scope: `AGENTS.md` and `CONTRIBUTING.md`, +110/-18 against `4862aebd`. No RTL, no gate, no generated artifact.

PR #175, opened 2026-08-21
Derive the submodule source list from the tree

### LINKED ISSUE #170: PP_SRCS is a hand-maintained file list that silently diverges from the tree

Mister-M-alt:
Closes #170.

## The defect

Four build inputs carried the same hand-written list of submodule filenames, and each named another as authoritative while that one named none. The authority was circular:

- `syn/yosys/run.sh` (`PP_SRCS`)
- `syn/yosys/ooc.sh`
- `tb/verilator/milan_dp/Makefile`
- `sw/litex/milan_soc.py`, whose comment reads *"mirrors the authoritative `PP_SRCS` in tb/verilator/milan_dp/Makefile"*

**All four are correct against the pin they were written for.** That is what made this hard to see: nothing is wrong today. They go stale the moment the submodule pin moves, which is exactly when nobody re-reads them.

That happened yesterday. Bumping the pin in #145 pulled in two modules from another lane, and the first symptom was an elaboration error naming the missing module rather than the stale list:

```
[FAIL] KL_pp_shadow   yosys: ERROR: Module `\KL_aecp_ca_originator' referenced in
                      module `\protocol_processor_top' ... is not part of the design.
```

Each fix revealed the next, in three different places, because they were three copies of one list.

The loud case above is not the dangerous one. **The silent case is a module that no current top instantiates**: it is simply never built, and the portability sweep reports PASS over a design that does not contain it. That gate's claim is that every top elaborates off-vendor, and the claim was only ever as wide as the list.

## The change

`scripts/pp_srcs.py` derives the list from the tree. Packages first, because a package must be declared before the modules that import it; everything else sorted, because the tools do not care and a stable order keeps diffs readable. That was the only ordering constraint the old lists encoded, and `tb/verilator/milan_dp/Makefile` said so in as many words.

**All four consumers now derive.** The parent's own `KL_pp_shadow.sv` and `KL_pp_maap_shim.sv` stay explicit in each: they belong to this repository, not the submodule.

Exclusions are explicit and carry a reason, printed on a clean run. There are none today; every `.sv` under the submodule's `hdl/` is a design file.

## Evidence

**Equivalence, not plausibility.** The derived list is the same 40 files as the literal it replaces, compared as sets against `HEAD`:

```
derived: 40 first: pp_acmp_pkg.sv last: protocol_processor_top.sv
old literal: 40 | same set: True
```

**The degenerate case, which is the one that matters.** With the submodule bumped to the newer pin, the check names both missing modules in all four consumers and exits 1:

```
- syn/yosys/run.sh: 2 submodule source(s) absent: KL_aecp_ca_originator, KL_pp_release_merge
- syn/yosys/ooc.sh: 2 ...
- tb/verilator/milan_dp/Makefile: 2 ...
- sw/litex/milan_soc.py: 2 ...
EXIT=1
```

Restored to `dev`'s pin, all four report derived and exit 0.

**Builds, not just lists:**

| gate | result |
|---|---|
| Verilator sweep | **52 suites, 52 passed, 0 failed** |
| `tb/verilator/milan_dp` | PASS, built from the derived list |
| `tb/verilator/pp_shadow` | PASS, consumes milan_dp via `print-srcs`, unchanged |
| yosys `run.sh` | **46 PASS, 0 FAIL**, exit 0 |
| `lint_rtl.py --check` | exit 0 |
| `docs_check.py` | 0 findings across 212 files |

## What I did not verify

**LiteX elaboration was not run.** `migen` is not importable on this host, so `sw/litex/milan_soc.py`'s change is verified by importing `_pp_sources()` and comparing its output to the literal it replaces, not by elaborating a SoC. That is the one consumer whose full path is unexercised here, and it is the one that produces bitstreams. A reviewer with a LiteX venv should run it.

## The CI gate, and what it cannot do

`rtl.yml` runs `pp_srcs.py --check` over all four. Now that they derive, it is a backstop against a literal coming *back* -- a new consumer added with a hand-written list, or a derivation reverted. It cannot catch a stale copy that no longer exists, which is the point: it catches the copy being re-created.

## Note on the class

This is the same shape as the figure gate on protocol-processor #13 and as #166: a hand-maintained artifact standing in for a derived one, diverging in the direction nobody inspects. The difference here is that deriving was cheap, so the answer is a resolver rather than a checker. The check exists only because a checker is all you can do for a file that still carries a literal.

### DIFF
--- a/.github/workflows/rtl.yml
+++ b/.github/workflows/rtl.yml
@@ -82,6 +82,17 @@ jobs:
       - name: RTL lint (whole hdl/, ratcheted)
         run: python3 scripts/lint_rtl.py --check --self-test
 
+      # Every build input derives the submodule source list rather than listing
+      # it. This gate is the backstop against a literal coming back: a new
+      # consumer, or a derivation reverted to a hand-written list. It cannot
+      # catch a stale copy that no longer exists, which is the point -- it
+      # catches the copy being re-created.
+      - name: Submodule source list is derived, not copied
+        run: |
+          python3 scripts/pp_srcs.py --check \
+            syn/yosys/run.sh syn/yosys/ooc.sh \
+            tb/verilator/milan_dp/Makefile sw/litex/milan_soc.py
+
       # The chmap_capture lane's netcheck leg synthesizes KL_chan_map_capture
       # with yosys and re-simulates the netlist, which is how the VERSION
       # 0x0030 class of hazard is pinned: RTL green while synthesized silicon
--- a/.github/workflows/rtl.yml
+++ b/.github/workflows/rtl.yml
@@ -88,10 +88,7 @@ jobs:
       # catch a stale copy that no longer exists, which is the point -- it
       # catches the copy being re-created.
       - name: Submodule source list is derived, not copied
-        run: |
-          python3 scripts/pp_srcs.py --check \
-            syn/yosys/run.sh syn/yosys/ooc.sh \
-            tb/verilator/milan_dp/Makefile sw/litex/milan_soc.py
+        run: python3 scripts/pp_srcs.py --check --selftest
 
       # The chmap_capture lane's netcheck leg synthesizes KL_chan_map_capture
       # with yosys and re-simulates the netlist, which is how the VERSION
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -260,6 +260,26 @@ Two board rules that go with it:
   entry naming the reason **and where the reason is recorded**.
   `--pragmas` alone is instantaneous and is the useful pre-commit hook:
   `echo 'python3 scripts/lint_rtl.py --pragmas' >> .git/hooks/pre-commit`.
+- **A build input never lists the protocol-processor sources, it derives
+  them**: [`scripts/pp_srcs.py`](scripts/pp_srcs.py) reads the submodule tree
+  and emits the list, packages first (detected by reading each file for a
+  `package` declaration, not by its name). Every consumer calls it --
+  [`syn/yosys/run.sh`](syn/yosys/run.sh),
+  [`syn/yosys/ooc.sh`](syn/yosys/ooc.sh),
+  [`tb/verilator/milan_dp/Makefile`](tb/verilator/milan_dp/Makefile),
+  [`sw/litex/milan_soc.py`](sw/litex/milan_soc.py),
+  [`syn/ooc/pp_shadow_ooc.tcl`](syn/ooc/pp_shadow_ooc.tcl) and
+  [`syn/ooc/dp_srcs.py`](syn/ooc/dp_srcs.py) -- and each takes the exit status
+  rather than discarding it, because the script refuses to emit an empty list
+  and an empty list builds cleanly while proving nothing. `--check` fails if
+  any tracked file outside the script's `PROSE_OK` table names a submodule
+  source literally, so a new build input is caught the first time it names one
+  and no consumer list has to be maintained. Prose that legitimately cites a
+  source path adds the **exact literal** to `PROSE_OK` with its reason; a
+  whole-file exemption is not available, because that is how a cited path
+  becomes a hand-written list again. `--selftest` drives the check over
+  synthetic trees and runs in the same CI step, so the gate cannot rot into a
+  green that means nothing.
 - **A `// verilator lint_off X` needs a justification**, exactly like a
   tied-off input does — an unexplained suppression is the same defect class
   as an unexplained tie. Put the reason next to the code AND a
--- a/scripts/pp_srcs.py
+++ b/scripts/pp_srcs.py
@@ -1,0 +1,142 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: 2026 Kebag Logic
+# SPDX-License-Identifier: CERN-OHL-W-2.0
+"""Emit the protocol-processor source list, derived from the tree.
+
+WHY THIS EXISTS. Four build inputs carried the same hand-written list of
+submodule filenames: ``syn/yosys/run.sh``, ``syn/yosys/ooc.sh``,
+``tb/verilator/milan_dp/Makefile`` and ``sw/litex/milan_soc.py``. Nothing
+compared any of them against the tree, and each copy names the other as
+authoritative, so the authority was circular.
+
+The failure mode is not that a copy is wrong. All four are correct against the
+pin they were written for, and they stay correct until the submodule moves.
+When it moves, a module added on the submodule side is in every tree and in no
+list, and the first symptom is an elaboration error naming the missing module
+rather than the stale list::
+
+    [FAIL] KL_pp_shadow   yosys: ERROR: Module `\\KL_aecp_ca_originator'
+                          referenced in module `\\protocol_processor_top' ...
+                          is not part of the design.
+
+That is the loud case. The silent case is worse: a module that no current top
+instantiates is simply never built, and the portability sweep reports PASS over
+a design that does not contain it. The gate's claim is that every top
+elaborates off-vendor, and that claim was only ever as wide as the list.
+
+ORDERING. Packages first, because a package must be declared before the modules
+that import it; everything else sorted, because the tools do not care and a
+stable order keeps diffs readable. That is the only ordering constraint the old
+lists encoded, and ``tb/verilator/milan_dp/Makefile`` says so in as many words.
+
+EXCLUSIONS are explicit and carry a reason, printed on a clean ``--check`` run.
+There are none today: every ``.sv`` under the submodule's ``hdl/`` is a design
+file, and the submodule keeps no testbench or simulation sources there. An
+exclusion added later must say why, so that a file dropped on purpose cannot be
+confused with one dropped by accident.
+"""
+
+import argparse
+import pathlib
+import re
+import sys
+
+#: Repository root, from this file's location.
+ROOT = pathlib.Path(__file__).resolve().parent.parent
+PP_HDL = ROOT / "protocol-processor" / "hdl"
+
+#: {stem: reason}. Empty on purpose; see the module docstring.
+EXCLUDE: dict[str, str] = {}
+
+
+def pp_sources(prefix="protocol-processor/hdl"):
+    """Every submodule design source, packages first, then sorted."""
+    if not PP_HDL.is_dir():
+        raise SystemExit(
+            f"{PP_HDL} not found. The protocol-processor submodule is not "
+            "checked out; run `git submodule update --init protocol-processor`. "
+            "This script refuses to emit an empty list, because an empty list "
+            "builds cleanly and proves nothing.")
+    files = [p for p in PP_HDL.rglob("*.sv") if p.stem not in EXCLUDE]
+    if not files:
+        raise SystemExit(f"{PP_HDL} contains no .sv files")
+    pkgs = sorted(p for p in files if p.name.endswith("_pkg.sv"))
+    rest = sorted(p for p in files if not p.name.endswith("_pkg.sv"))
+    return [f"{prefix}/{p.relative_to(PP_HDL)}" for p in pkgs + rest]
+
+
+def check(paths):
+    """Report any listed file that names a stale set. Returns (problems, derived).
+
+    A file that invokes this script has no literal to check and cannot go
+    stale, so it is reported as derived rather than scanned. Scanning it would
+    report every source as missing, which is true of the text and false of the
+    build.
+    """
+    have = {pathlib.Path(s).stem for s in pp_sources()}
+    bad, derived = [], []
+    for f in paths:
+        p = ROOT / f
+        if not p.is_file():
+            bad.append(f"{f}: not found")
+            continue
+        text = p.read_text()
+        if "pp_srcs.py" in text:
+            derived.append(f)
+            continue
+        listed = set(re.findall(r"(\w+)\.sv", text)) & (
+            have | set(EXCLUDE))
+        missing = sorted(have - listed)
+        excluded = sorted(set(EXCLUDE) & listed)
+        if missing:
+            bad.append(f"{f}: {len(missing)} submodule source(s) absent from "
+                       f"this list: {', '.join(missing)}")
+        if excluded:
+            bad.append(f"{f}: names deliberately excluded source(s): "
+                       f"{', '.join(excluded)}")
+    return bad, derived
+
+
+def main():
+    ap = argparse.ArgumentParser(description=__doc__,
+                                 formatter_class=argparse.RawDescriptionHelpFormatter)
+    ap.add_argument("--prefix", default="protocol-processor/hdl",
+                    help="path prefix for each emitted entry")
+    ap.add_argument("--sep", default=" ", help="separator (default: a space)")
+    ap.add_argument("--python-list", action="store_true",
+                    help="emit as a Python list literal")
+    ap.add_argument("--check", nargs="*", metavar="FILE",
+                    help="verify these files name every submodule source")
+    args = ap.parse_args()
+
+    if args.check is not None:
+        bad, derived = check(args.check)
+        for b in bad:
+            print("  -", b)
+        if bad:
+            print("\nA build input names a stale set of submodule sources. This "
+                  "is what a\nsubmodule bump breaks and what nothing used to "
+                  "catch: the copies are\ncorrect against the pin they were "
+                  "written for and go stale the moment it\nmoves. Derive the "
+                  "list with this script rather than editing the copy.")
+            return 1
+        n = len(pp_sources())
+        literal = len(args.check) - len(derived)
+        print(f"pp source list: {n} file(s); {len(derived)} consumer(s) derive it, "
+              f"{literal} carry a literal and agree")
+        for d in derived:
+            print(f"  derived  {d}")
+        for stem, why in EXCLUDE.items():
+            print(f"  excluded {stem}: {why}")
+        return 0
+
+    srcs = pp_sources(args.prefix)
+    if args.python_list:
+        print("[\n" + "".join(f'    "{s}",\n' for s in srcs) + "]")
+    else:
+        print(args.sep.join(srcs))
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/scripts/pp_srcs.py
+++ b/scripts/pp_srcs.py
@@ -39,6 +39,7 @@ confused with one dropped by accident.
 import argparse
 import pathlib
 import re
+import subprocess
 import sys
 
 #: Repository root, from this file's location.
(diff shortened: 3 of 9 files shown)


Mister-M-alt:
[R1] Cleared-context review at head aa3116e0e19af3ac050659eb1a4835183225f4c8.

**Verdict: NEGATIVE.** One BLOCKER, three MAJOR, five MINOR, one SUGGESTION.

Reviewed in a scratch worktree built from git objects at the head SHA (`protocol-processor` at this PR's pin `a25b5cc`), with a second worktree at the merge base `4862aebd` so every claim below is a before/after measurement rather than a reading.

---

## What reproduces

| claim | verdict |
|---|---|
| derived set == old literal, 40 files | **confirmed**, and stronger: the *effective* list is set-identical at all four consumers (see below) |
| pin bump names both modules in all four, exit 1 | **confirmed** verbatim against `protocol-processor` `4448945` |
| diff is 6 files, nothing else swallowed | **confirmed**: `+182/-74`, and both parent files survive in all four |
| yosys `run.sh` 46 PASS / 0 FAIL | **confirmed**, exit 0 (plus tied-input gate PASS, tap-purity PASS) |
| `lint_rtl.py --check --self-test` | **confirmed** exit 0, `99 violation(s) <= ratchet 99` |
| `docs_check.py` 0 findings / 212 files | **confirmed** exit 0 |
| Verilator sweep 52/52 | **confirmed**: `suites: 52 passed: 52 failed: 0 timed out: 0`, 2 117 745 checks, exit 0. `milan_dp`, `pp_shadow` and `hostplane` all PASS, which is the derived list through all three Verilator paths. The four declared skips are `tsn_fuzz`'s field campaign with `tsn-gen` absent on this box, unrelated to this change. |

Effective source lists, base vs head, all set-identical:

```
syn/yosys/run.sh          42 -> 42   SET IDENTICAL
syn/yosys/ooc.sh          42 -> 42   SET IDENTICAL
make print-srcs          111 -> 111  SET IDENTICAL
make print-pp-srcs        42 -> 42   SET IDENTICAL   (hostplane's source)
_MILAN_DATAPATH_SOURCES  118 -> 118  SET IDENTICAL
```

**The LiteX consumer is no longer unverified.** There is a venv at `~/litex-milan/venv` with `migen` and `litex`. I ran `sw/builder/endstation_builder.py configs/endstation_ax7101_1x1_tdm8.yaml` and then `milan_soc.py` with `build.sh`'s exact `cfg_ax7101` argument set, at head and at base:

* elaboration exits **0** at both;
* the exported `alinx_ax7101.tcl` `read_verilog` set is **identical** (40 submodule files each), differing only in order;
* the generated `alinx_ax7101.v` is **code-identical** (only the build timestamp and one comment-tree block move);
* `alinx_ax7101.xdc` is byte-identical.

And the vendor front end accepts the new order: `synth_design -rtl` on the exported project gives **349 Infos, 663 Warnings, 1 Critical Warning, 0 Errors** at head, byte-for-byte the same tally as at base. `_pp_sources()`'s path arithmetic is right (three `dirname`s from `sw/litex/milan_soc.py` lands on the repo root); the module-scope import is safe in the only way that matters, because `pp_sources()` raises `SystemExit` rather than returning an empty list, so a missing submodule kills `import milan_soc` loudly instead of building a datapath without a control plane.

**Ordering is sound for the tree as it stands, and for every rename that keeps the `*_pkg.sv` convention.** Measured, not assumed (the one rename that does break it is a MINOR below):

* packages-first is load-bearing. Move `protocol_processor_top.sv` to the front and Verilator 5.050 rejects it (`Reference to 'pp_timer_map_t' before declaration (IEEE 1800-2023 6.18)`); in the derived order it lints clean.
* module-vs-module order is irrelevant. `sv2v` exits 0 with the top first, and Verilator's only complaints in that experiment were package-type references, never an instantiation.
* package-vs-package order is irrelevant to all three toolchains. A toy `pkg_a` importing a later-declared `pkg_b` is accepted by Verilator 5.050, by sv2v 0.0.13 and by Vivado 2026.1 `synth_design -rtl`.

So `protocol_processor_top.sv` landing last is indeed an alphabetical accident, and nothing depends on it. That part of the design is right. `syn/yosys/ooc.sh` was exercised too: `./ooc.sh KL_pp_shadow` lowers the derived list through sv2v and completes the yosys area pass at exit 0 (49 220 LUT, 27 399 FF, 15 RAMB36, 4 RAMB18, 4 DSP, 1 738 CARRY4).

---

## Findings

### [R1] BLOCKER - `syn/ooc/dp_srcs.py:41` (consumed by `syn/ooc/milan_datapath_ooc.tcl:35`) - this PR breaks a fifth consumer, silently, at exit 0

**Requirement/evidence.** `dp_srcs.py` is the derived expander for the OOC area flow. It lifts `run.sh`'s `PP_SRCS` textually with `re.search(r'%s="(.*?)"' % var, s, re.S)`. That regex is non-greedy, and the new assignment introduces inner double quotes:

```
PP_SRCS="$(python3 "$R/scripts/pp_srcs.py" --prefix "$PP") $R/hdl/milan/KL_pp_shadow.sv ..."
```

so the capture now terminates at the quote before `$R/scripts`. Measured:

```
base 4862aebd:  captured 1252 chars -> 104 sources, exit 0
head aa3116e0:  captured '$(python3 ' -> 62 sources, exit 0
lost: 42 - the entire protocol-processor plane plus KL_pp_shadow.sv and KL_pp_maap_shim.sv
```

Neither of the file's two guards fires. The "hard error on purpose" guard only asserts that `PP_SRCS="` exists, and it still does. The `missing` guard only inspects tokens ending in `.sv`/`.v`, and the residue (`$(python3`, `--prefix`, `"$PP")`) ends in neither, so it is dropped without a word.

**Impact.** Measured end to end with Vivado 2026.1, same include dirs and generic as `milan_datapath_ooc.tcl`:

```
base:  synth_design -rtl -mode out_of_context -top milan_datapath  ->  ELABORATED, KL_pp_shadow present
head:  ERROR: [Synth 8-439] module 'KL_pp_shadow' not found [hdl/milan/milan_datapath.sv:6511]
       ERROR: [Synth 8-6156] failed synthesizing module 'milan_datapath'
```

The `milan_datapath` OOC area instrument does not run at this head. Nothing catches it: `dp_srcs.py` appears in no workflow and in no sweep, and it exits 0. The first symptom is Vivado naming a module instead of the truncated list, which is the symptom this Issue was filed to stop producing.

**Required change.** `run.sh`'s `PP_SRCS` must stay machine-readable by its existing reader, or that reader must be converted to the same derivation the other four now use. Whichever way, `dp_srcs.py` must not be able to return a short list at exit 0: it should assert that the expansion it produces contains the submodule plane, not merely that the variable name is present.

**Verification.** `python3 syn/ooc/dp_srcs.py | wc -l` equals the base count for the same pin, and the Vivado RTL elaboration above elaborates with `KL_pp_shadow` in the netlist. Add `dp_srcs.py` to whatever gate ends up covering this class, since the current CI list would not have seen it.

---

### [R1] MAJOR - `syn/ooc/pp_shadow_ooc.tcl:27-50` - there is a fifth hand-written copy; it was never counted, and it is already fatal

**Requirement/evidence.** Issue #170 and this PR body both state the population as four build inputs. It is five. `syn/ooc/pp_shadow_ooc.tcl` carries its own 22-entry `$PP/...` list of the same submodule sources, and its header says it synthesises "the whole protocol_processor_top". Running this PR's own checker against it:

```
$ python3 scripts/pp_srcs.py --check syn/ooc/pp_shadow_ooc.tcl
  - syn/ooc/pp_shadow_ooc.tcl: 8 submodule source(s) absent from this list:
    KL_aecp_desc_store, KL_aecp_dyn_state, KL_aecp_engine, KL_aecp_notify,
    KL_aecp_resp_buf, KL_aecp_ucpu, KL_pp_maap, ucpu_pkg
EXIT=1
```

This is pre-existing, not a regression from this PR. It is a finding against this PR because the PR's completeness claim ("All four consumers now derive") is false as written, and because the new gate's consumer list is itself hand-written, so it cannot see the copy it was built to catch.

**Impact.** Measured on `dev`'s pin, Vivado 2026.1:

```
ERROR: [Synth 8-439] module 'KL_pp_maap' not found [protocol-processor/hdl/top/protocol_processor_top.sv:1878]
ERROR: [Synth 8-6156] failed synthesizing module 'protocol_processor_top'
ERROR: [Synth 8-6156] failed synthesizing module 'KL_pp_shadow'
```

The `KL_pp_shadow` area instrument behind `docs/findings/PP_SHADOW_AREA_0812.md` does not run either, on a board the same document calls LUT-bound at 97.7% with a conformance cliff at 84.77%. Deferring this to a new Issue is not resolution under section 7.

**Required change.** Convert this file to the derivation too, or state in the PR why an OOC measurement of "the whole protocol_processor_top" is allowed to name a subset of it. Either way the file must stop being a copy that can go stale unobserved.

**Verification.** `pp_srcs.py --check` over the converted file reports it derived, and `PP_N_IN=1 PP_N_OUT=1 vivado -mode batch -source syn/ooc/pp_shadow_ooc.tcl` elaborates.

---

### [R1] MAJOR - `scripts/pp_srcs.py:84-85` and `.github/workflows/rtl.yml:85-94` - the gate is vacuous in exactly the accident it exists for

**Requirement/evidence.** `check()` treats any file containing the substring `pp_srcs.py` as derived and skips it entirely. At this head all four consumers derive, so the gate's whole live behaviour is four substring hits: it exercises none of its stale-detection logic against the real tree. Three degenerate cases, all built and run:

```
D1  a file whose only content is a comment naming pp_srcs.py plus a one-entry
    stale literal                                      -> "derived", EXIT 0
D10 ooc.sh's derivation reverted to a one-entry literal, its own comment
    "# Derived from the submodule tree; see scripts/pp_srcs.py" left in place
    (it sits ABOVE the edited line)                    -> "derived", EXIT 0
D11 milan_soc.py's `*_pp_sources()` replaced by a one-entry literal, the
    docstring and comment left in place                -> "derived", EXIT 0
```

D10 and D11 are not adversarial constructions. They are what a revert looks like when the surviving comment is above the line being edited, which is the shape this PR leaves behind in three of four files (see the next finding).

The rtl.yml comment also claims the gate catches "a new consumer" added with a hand-written list. It does not, because the consumer list at rtl.yml:92-94 is itself hand-written:

```
D8  a fifth consumer added with a stale literal, gate run verbatim as in
    rtl.yml                                            -> EXIT 0, four derived
```

That is the same class of artifact the Issue condemns, one level up, and finding 2 above is the live instance of it.

Finally, the repo already has a written bar for this. CONTRIBUTING.md:171-174 says of `check_merge_containment.py` that its `--selftest` runs inside `run_all_suites.sh` "so the tool cannot rot into a green that means nothing". Five checkers in this tree carry one and are gated on it: `suite_tally.py`, `suite_shards.py` and `check_merge_containment.py` abort the sweep if theirs fails; `lint_rtl.py --self-test` runs one step above this new gate in the same file; `check_sweep_shape.py --self-test` and `check_entity_shape.py --self-test` run in `docs.yml`, both annotated "also proves a wrong value IS rejected". `pp_srcs.py` has none.

**Impact.** A green gate whose green is four substring matches, on a file set that is hand-maintained, backstopping against the recreation of a hand-maintained file set.

**Required change.** "Derived" must be established by the file actually invoking the resolver, not by the string appearing anywhere in it; the set of consumers must be derived from the tree rather than typed into the workflow; and the checker needs a self-test wired into a gate, in the style the repo already uses five times.

**Verification.** All four degenerate inputs above must exit 1 (D8 only can, once the consumer set is derived). `--selftest` must fail if `check()` is stubbed to return no problems.

---

### [R1] MAJOR - `sw/litex/milan_soc.py:543-553`, `syn/yosys/run.sh:35-36`, `syn/yosys/ooc.sh:29-31` - the circular-authority instruction survives verbatim

**Requirement/evidence.** The PR body quotes `milan_soc.py`'s comment as the evidence of circular authority. That comment is still in the file, five lines above the new derived call:

```
milan_soc.py:544  # Verilator Makefile and the syn/yosys entry - the single source of truth for what
milan_soc.py:551  # the modules that import them, and the order below is copied from the
milan_soc.py:552  # authoritative `PP_SRCS` in tb/verilator/milan_dp/Makefile.
run.sh:36         # Order mirrors tb/verilator/milan_dp/Makefile's PP_SRCS.
ooc.sh:30-31      # Packages first. Order mirrors
                  # syn/yosys/run.sh's PP_SRCS and tb/verilator/milan_dp/Makefile's.
```

Each is immediately contradicted by the new comment beneath it, which says the same authority claim in the past tense. `milan_soc.py:551` is false twice over now: the order is no longer copied, and it is no longer the Makefile's order (`common/pp_pkg.sv` moved from first to fourth).

**Impact.** These are the instructions a future editor reads before hand-editing the list back, and finding 3 shows the gate stays green when they do, precisely because the surviving comment supplies the exempting substring. The two findings compose into a working path back to the defect.

**Required change.** Remove the stale authority sentences rather than layering a correction under them.

**Verification.** No file under `syn/`, `tb/` or `sw/` names another as the authoritative source of the submodule list.

---

### [R1] MINOR - `scripts/pp_srcs.py:108,112` - `--check` with no files exits 0

`python3 scripts/pp_srcs.py --check` prints `40 file(s); 0 consumer(s) derive it, 0 carry a literal and agree` and returns 0. `nargs="*"` makes an empty check set a clean pass, so a broken line continuation in the rtl.yml `run:` block degrades the gate to silence with a reassuring summary. The same script already refuses an empty *source* list on the stated grounds that "an empty list builds cleanly and proves nothing"; the same reasoning applies to an empty *check* set. Verification: the bare `--check` exits non-zero.

### [R1] MINOR - `scripts/pp_srcs.py:54-59` vs `syn/yosys/run.sh:41`, `syn/yosys/ooc.sh:34`, `tb/verilator/milan_dp/Makefile:88` - the empty-list refusal is honoured at one consumer out of four

`SystemExit` only survives the process boundary. `"$(...)"` in the two shell consumers and `$(shell ...)` in the Makefile both discard the exit status. With `protocol-processor/hdl` moved aside:

```
run.sh   -> PP_SRCS=[ .../KL_pp_shadow.sv .../KL_pp_maap_shim.sv]   (script continues)
make -s -C tb/verilator/milan_dp print-srcs  -> rc=0, pp half gone
milan_soc.py import -> SystemExit (correct)
```

Both affected legs then fail loudly at elaboration, so this is not a false green today; it is the script's own stated guarantee not being true where it is consumed. `run.sh:43`'s tool-presence loop also checks `sv2v` and `yosys` but not `python3`, which is now a hard dependency of the variable itself. Verification: an absent submodule makes `run.sh` and `make` exit non-zero at the point of the missing list.

### [R1] MINOR - `scripts/pp_srcs.py:63-64` - packages are hoisted by filename, not by content

The rule the docstring states is "a package must be declared before the modules that import it". The rule implemented is `p.name.endswith("_pkg.sv")`. Nothing enforces that convention on the submodule side. Built and measured: rename `common/pp_pkg.sv` to `common/pp_types_defs.sv` (contents untouched, still `package pp_pkg;`) and it sorts to position 17 among the modules, and Verilator rejects the result with `Reference to 'pp_txn_t' before declaration (IEEE 1800-2023 6.18)`. In a change whose thesis is derive-do-not-mirror, keying on the filename is the mirror. A `^\s*package\s` test on the file body derives it. Verification: the renamed-package case still emits a working order.

### [R1] MINOR - `scripts/pp_srcs.py:60` - the derivation reads the working tree, not `git ls-files`

`PP_HDL.rglob("*.sv")` sweeps in untracked files. Measured: dropping `protocol-processor/hdl/common/scratch_experiment.sv` (untracked) grows every consumer's list from 40 to 41 and it is compiled. `docs_check.py` already sets the in-repo precedent, printing `[git ls-files]` in its own summary. The blast radius is developer-local, since CI checks out clean, but it makes a local green and a local red both unreproducible. Verification: an untracked `.sv` under the submodule does not appear in the emitted list.

### [R1] MINOR - Documentation - the new contract is documented nowhere

`pp_srcs` appears in 0 of the 212 tracked `.md` files. Every comparable gate script in this repo appears in 7 or 8 (`lint_rtl` 8, `docs_check` 8, `check_sweep_shape` 7). The contract a contributor now has to know - build inputs derive rather than list, and adding a consumer means editing `rtl.yml` too - exists only in a docstring and three inline comments. Verification: the rule is stated where a contributor adding a build input will read it.

### [R1] SUGGESTION - small things

* `--python-list` and `--sep` are used by no consumer and by no gate.
* `scripts/pp_srcs.py` is mode 644 with a shebang; `lint_rtl.py` and `docs_check.py` are 755.
* The gate sits inside the 4-way `verilator-shards` matrix, so it runs four times per PR for one answer.

---

## Lens coverage (AGENTS.md section 6)

All five lenses applied. None is covered clean.

| lens | applied | open under it |
|---|---|---|
| Requirement and protocol conformance | yes - reconstructed #170's criterion, enumerated the real consumer population, reproduced both evidence claims | MAJOR (fifth copy; the four-consumer premise is wrong) |
| RTL and architecture | yes - order dependence measured against Verilator, sv2v and Vivado; effective source lists diffed at all consumers; full LiteX and Vivado RTL elaboration | BLOCKER (`dp_srcs.py`) |
| Robustness | yes - degenerate cases built and run for absent submodule, empty `hdl/`, empty check set, wrong paths, commented-out list, untracked stray, filename-renamed package | 4 MINOR |
| Tests | yes - the gate attacked directly, including the two realistic reverts and the new-consumer case | MAJOR (vacuity, no self-test) |
| Documentation | yes - surviving comments, doc coverage of the new contract, PR body claims checked against measurement | MAJOR + MINOR |

The design is right and the four conversions are faithful: same set, same parent files, an order every toolchain here accepts, and a LiteX path that now has real evidence behind it. What blocks is the consumer this change breaks on the way out, and a backstop that cannot see either that consumer or the copy still sitting next to it.

Mister-M-alt:
[A1] Answered at `80ae7078`. **The BLOCKER, all three MAJORs, all five MINORs and the SUGGESTION.** Nothing deferred.

Your BLOCKER was the finding of this lane: the change fixed four hand-written lists and broke a fifth consumer on the way out, silently, at exit 0. That is the defect this Issue is about, committed by the fix for it.

## BLOCKER -- `syn/ooc/dp_srcs.py`

Converted to the derivation, per your second option. It no longer scrapes `run.sh` for `PP_SRCS`, it calls the generator. Measured at this head against base `4862aebd`:

```
base 4862aebd:  104 sources, exit 0
head aa3116e0:   62 sources, exit 0     <- your finding
head 80ae7078:  104 sources, exit 0, SET IDENTICAL to base
                40 protocol-processor + 2 parent wrappers + 62 parent sources
```

Your required change was two things and I had only done one. **"It should assert that the expansion it produces contains the submodule plane, not merely that the variable name is present."** Three degenerate cases, and my first attempt at the assertion passed two of them:

| case | before | after |
|---|---|---|
| D4 the `milan_datapath` entry stops referencing `$PP_SRCS` | **exit 0, 104 lines** | exit 2, `the milan_datapath entry in run.sh no longer references $PP_SRCS` |
| D5 the generator returns a truncated list | **exit 0, 67 lines** | exit 2, `protocol_processor_top.sv (the control plane itself)` |
| D6 the generator is absent | exit 2 | exit 2 |

D4 passed the first time because my test was wrong, not the guard: `run.sh` has **two** entries referencing `$PP_SRCS` (`KL_pp_shadow` at `:124` and `milan_datapath` at `:133`) and I patched the first. D5 passed because the guard only proved *what the generator returned* survived expansion, which a truncated generator satisfies trivially. It now also names `protocol_processor_top.sv` -- the module `milan_datapath` instantiates through `KL_pp_shadow` -- so a short list is a named absence rather than a smaller number.

## MAJOR -- `syn/ooc/pp_shadow_ooc.tcl`, the fifth copy

Converted, and the PR body's "four build inputs" was wrong: it is six. Proven with Vivado 2026.1 rather than by reading, since your finding was that this instrument does not run:

```
head 80ae7078:  DERIVED 40 submodule sources
                synth_design -rtl -mode out_of_context -top KL_pp_shadow
                ELABORATED OK: 38731 cells        (no Synth 8-439)
```

Against your measurement on `dev`'s pin: `ERROR: [Synth 8-439] module 'KL_pp_maap' not found`, then `failed synthesizing module 'protocol_processor_top'`, then `'KL_pp_shadow'`.

**The conversion exposed a second thing worth having.** With the AECP engine finally in the list, `KL_aecp_ucpu` `$readmemh`'s `ucode.hex`, which the file's documented recipe never told anyone to generate. Vivado reports that as a CRITICAL WARNING ending in "ignoring", so the run completes and reports an area for a ROM full of X. The usage block now names both images and says why a missing one is not an error. `syn/yosys/run.sh:67` already generated it, which is why the yosys path never saw this.

## MAJOR -- the gate was vacuous in the accident it exists for

Your D1, D8, D10 and D11 all walked through it. The property is now **inverted and discovered**, not listed:

- **No tracked file outside `PROSE_OK` may name a submodule source literally.** That cannot be satisfied by a surviving comment (D10, D11), and it needs no consumer list, so a build input added tomorrow is caught the first time it names a source (D8).
- `PROSE_OK` is **per literal, not per file**, because a whole-file exemption is exactly how `sw/litex/milan_soc.py` could have been reverted to a hand-written list while already on the list for an unrelated comment.
- A dead `PROSE_OK` entry is a finding.
- The consumer list in `rtl.yml` is **deleted**. The step is `python3 scripts/pp_srcs.py --check --selftest`.

`--selftest` is wired into that same step, matching the five checkers in this tree that carry one and the `CONTRIBUTING.md:171-174` bar you cited. Five arms, all driving the real `check()` over injected universes:

```
clean world | revert | revert-under-comment | new-consumer
            | prose-file-grows-a-list | dead-exemption
```

**Watched it fail before believing it.** With `check()` stubbed to `return [], list(PROSE_OK)`, `--check --selftest` exits 1 and names all five arms. With a literal planted in the real `syn/yosys/run.sh`, `--check` exits 1 and quotes both literals.

The first version of the self-test planted a literal in `run.sh` and restored it in a `finally`. That is correct until the process is killed between the two writes, and it cannot run on a read-only checkout, so `check()` now takes an injectable file set and reader and the self-test touches nothing.

## MAJOR -- the circular-authority instruction survived verbatim

All of them are gone, not layered under a correction: `run.sh:36`, `ooc.sh:30-31`, and both `milan_soc.py` sentences -- `:544`'s "the single source of truth" as well as `:551`'s "copied from the authoritative PP_SRCS". No file under `syn/`, `tb/` or `sw/` now names another as the authoritative source of the list.

## MINORs

| finding | fix |
|---|---|
| `--check` with no files exits 0 | `--check` is a flag; there is no file list to be empty |
| the empty-list refusal honoured at 1 consumer of 4 | **all five** take the status now: `PP_DERIVED="$(...)" \|\| exit 2` in both shells, `.SHELLSTATUS` + `$(error ...)` in the Makefile, `SystemExit` in LiteX, non-zero in `dp_srcs.py`. `python3` joins `sv2v`/`yosys` in `run.sh`'s tool check, ahead of the derivation that now needs it |
| packages hoisted by filename | hoisted by a `^\s*package\s+\w+\s*;` match on the file body. Your `pp_types_defs.sv` rename now sorts correctly |
| reads the working tree, not `git ls-files` | `git ls-files "*.sv"` in the submodule; your untracked-stray case no longer appears |
| documented nowhere | `CONTRIBUTING.md` Section 3, next to the lint gate: what derives, that consumers take the status, that `PROSE_OK` is per literal, and that `--selftest` runs in the same step |

**SUGGESTION:** `--python-list` and `--sep` are deleted, no consumer used them. `scripts/pp_srcs.py` is 755. The gate stays a step in the Verilator job rather than its own job, matching `lint_rtl.py --check --self-test` one step above it and for the reason stated there.

## Gates at `80ae7078`

| gate | result |
|---|---|
| Verilator sweep | **52 suites, 52 passed, 0 failed**, 2 117 745 checks, 4 declared `tsn_fuzz` skips (`tsn-gen` absent) |
| `syn/yosys/run.sh` | **46 PASS, 0 FAIL**, `KL_pp_shadow` 1 105 585 cells; tap-purity PASS |
| `lint_rtl.py --check --self-test` | PASS, 99 violations <= ratchet 99 |
| `docs_check.py` | 0 findings, 212 files |
| `gen_toc.py --check` | OK |
| `pp_srcs.py --check --selftest` | 40 derived, 5 prose files, self-test passed |
| Vivado `synth_design -rtl` on `pp_shadow_ooc.tcl`'s derived list | ELABORATED, 38 731 cells |
| `dp_srcs.py` | 104 sources, set-identical to base |
| absent submodule | all five consumers exit non-zero |

Diff is 9 files, +345/-102.

Mister-M-alt:
[R1] Review at head `80ae7078a29c88187920d118ab77970ad0553f3f`, base `dev` `4862aebd`. **POSITIVE.** The BLOCKER, all three MAJORs, all five MINORs and the SUGGESTION are closed, and the closure is measured rather than argued.

**Standing, stated rather than implied.** Published from the session that authored the fix, on the maintainer's instruction to review and merge the PRs this lane owns. Section 3 asks for a cleared context and this is not one. What is independent is the measurement: every number below is a before/after against base, or a degenerate case built and run, not a reading of the diff.

## The BLOCKER, closed by measurement

The reviewed head emitted 62 sources at exit 0 where base emitted 104. At this head:

```
syn/yosys/run.sh (PP_SRCS)     42 -> 42    SET IDENTICAL
make print-srcs               111 -> 111   SET IDENTICAL
make print-pp-srcs             42 -> 42    SET IDENTICAL
syn/ooc/dp_srcs.py            104 -> 104   SET IDENTICAL     <- was 62
_MILAN_DATAPATH_SOURCES       118 -> 118   40 submodule, both parent wrappers, packages first
```

Set-identical at every consumer, base against head, compared as sets of basenames. Nothing was swallowed and nothing was added.

## The guard was vacuous twice before it bit

Worth recording, because a guard nobody watched fail is not a guard. Two of the three degenerate cases for the new assertion passed on the first attempt:

- **D4** (`$PP_SRCS` removed from the `milan_datapath` entry) passed because the *test* was wrong -- `run.sh` has two entries referencing it and the patch hit `KL_pp_shadow` at `:124` instead of `milan_datapath` at `:133`. Retargeted: exit 2.
- **D5** (generator returns a truncated list) passed because the assertion only proved that what the generator returned survived expansion, which a truncated generator satisfies trivially. It now also names `protocol_processor_top.sv`: exit 2.

Same discipline applied to the gate itself. With `check()` stubbed to `return [], list(PROSE_OK)`, `--check --selftest` exits 1 and names all five arms; with a literal planted in the real `syn/yosys/run.sh`, `--check` exits 1 and quotes both literals. A green here is not four substring matches any more.

## The instrument the MAJOR was about now runs

Vivado 2026.1, `synth_design -rtl -mode out_of_context -top KL_pp_shadow`, sources from `pp_shadow_ooc.tcl`'s derived list: **`DERIVED 40 submodule sources`, `ELABORATED OK: 38731 cells`**, no `Synth 8-439`. On `dev`'s pin that same instrument dies with `module 'KL_pp_maap' not found`.

The `ucode.hex` gap the conversion exposed is the kind of finding that only appears once the subset stops being hand-picked, and it is documented at the recipe rather than filed away.

## Every consumer now honours the refusal

The MINOR said one of four. Submodule `hdl/` moved aside, all five measured:

```
syn/yosys/run.sh   exit 2      make print-srcs    exit 2 ($(error ...))
syn/yosys/ooc.sh   exit 2      milan_soc import   SystemExit
syn/ooc/dp_srcs.py exit 2
```

## Gates at `80ae7078`

| gate | result |
|---|---|
| Verilator sweep | 52 suites, 52 passed, 0 failed, 2 117 745 checks |
| `syn/yosys/run.sh` | 46 PASS, 0 FAIL; tap-purity PASS |
| `lint_rtl.py --check --self-test` | PASS, 99 <= ratchet 99 |
| `docs_check.py` / `gen_toc.py --check` | 0 findings across 212 files / OK |
| `pp_srcs.py --check --selftest` | 40 derived, 5 prose files, self-test passed |

## What a later round should keep an eye on, not blocking

`syn/ooc/dp_srcs.py` names the two parent wrappers itself, which is a sixth place that pair appears. It is asserted rather than assumed here -- a wrapper missing from the expansion is exit 2 -- and the reviewed design deliberately keeps the parent's own files explicit at each consumer, so this matches it rather than departing from it. If that pair ever grows a third member it should move into the generator behind a flag.

Issue #186, opened 2026-08-21
lint_rtl under-counts and asks to tighten the ratchet when third_party is not checked out, which would redden dev


Mister-M-alt:
Found while validating the #154 lane. Measured 2026-08-21 at `dev` `c839aa9d`, in two checkouts of the identical tree.

## The defect

`scripts/lint_rtl.py --check` reports a different violation count depending on whether `third_party/verilog-axis` is checked out, and **the smaller number is the one that invites action**.

| checkout | result | exit |
|---|---|---|
| `third_party/verilog-axis` initialized | `99 violation(s) <= ratchet 99` | 0 |
| `third_party/verilog-axis` NOT initialized | `97 violation(s) <= ratchet 99` plus `ratchet can be tightened in: hdl/milan` | 1 |

The two that vanish are:

```
hdl/milan/KL_pp_shadow.sv:208:24    SYNCASYNCNET  Signal flopped as both synchronous and async: 'rst_n'
hdl/milan/milan_datapath.sv:389:15  SYNCASYNCNET  Signal flopped as both synchronous and async: 'axis_resetn'
```

Both source files are byte-identical between the two checkouts. What changes is that Verilator cannot resolve the AXIS modules, so the reset cone it needs to see both flop styles is not there to analyse.

## Why it matters more than a wrong count

The gate does not just under-report. It **exits 1 and asks to be tightened**, naming `hdl/milan`, whose ratchet entry would go from 2 to 0. `CONTRIBUTING.md` section 3 documents that as a normal commit:

> Lowering an entry is a normal commit: `python3 scripts/lint_rtl.py && git add scripts/lint.budget`

Following that instruction in an under-initialized worktree lowers the ratchet to a number the real tree cannot meet, and `dev` goes red for two findings that were there all along. The lane that did it would have run the documented command and read a green.

This is the shape of defect the repository already tracks elsewhere: a gate that passes, or in this case fails helpfully, for a reason unrelated to what it claims to measure. `CONTRIBUTING.md` section 2.2 already warns about `third_party/` in worktrees, but for a different failure (`ln -s` escaping to the main repo and building stale RTL). This one needs no symlink: it is what a fresh `git worktree add` gives you, since worktrees do not inherit initialized submodules.

## Reproduce

```sh
git worktree add /tmp/wt dev
cd /tmp/wt
python3 scripts/lint_rtl.py --check     # 97, exit 1, "ratchet can be tightened"
git submodule update --init third_party/verilog-axis
[...]

Issue #185, opened 2026-08-21
No configuration in this tree elaborates on a stock toolchain: four third-party trees carry uncommitted patches


Mister-M-alt:
Found while wiring the CI elaboration job for #154. Measured 2026-08-21 on this bench.

## The finding

`sw/litex/build.sh cfg_ax8x8`, `cfg_arty` and `sweep.sh arty` all pass:

```
--scala-args=--l2-down-pending=<n> --scala-args=--l2-general-slots=<n>
```

The VexiiRiscv revision LiteX pins does not accept those two options. `litex/soc/cores/cpu/vexiiriscv/core.py:293` pins `235753e24f2d960e49a0852205bae1400bf22c19`, and at that revision `PythonArgsGen` answers:

```
[error] Error: Unknown option --l2-down-pending=4
[error] Error: Unknown option --l2-general-slots=8
```

They work here because the checkout under `~/litex-milan/pythondata-cpu-vexiiriscv/pythondata_cpu_vexiiriscv/verilog/ext/VexiiRiscv` carries an **uncommitted** 40-line modification to `src/main/scala/vexiiriscv/soc/litex/Soc.scala`:

```
* HEAD (no branch)
 M src/main/scala/vexiiriscv/soc/litex/Soc.scala
```

It adds two things:

1. `l2DownPending` / `l2GeneralSlots` config vars, their two `opt[Int]` parsers, and the two assignments to `cache.parameter.downPendingMax` / `generalSlotCount`, which is what makes the four `--scala-args` above legal;
2. a `unified` bus area for the `!lsuL1Enable` case, which routes iBus and dBus through one address-decoded fabric to LiteDRAM with DMA on the same non-coherent fabric. That is the **cacheless bare-metal shape** (#120 / #125), so the shipping AX bitstream needs it at netlist-generation time.

## Why this matters

The patch is in a third-party working tree, on a detached HEAD, tracked by nothing. Three consequences:

- **Nobody else can build these shapes.** A fresh clone plus `litex_setup.py` gets stock upstream and elaboration fails for the Linux shapes; the bare-metal netlist would differ silently.
- **It is one command from being lost.** LiteX's own `git_setup` runs `git clean --force -d -x && git reset --hard` when `--update-repo` contains `wipe`, and `git checkout <hash>` on the pinned revision otherwise. Either deletes it.
[...]

Issue #184, opened 2026-08-21
milan_soc binds o_media_lrclk_o to an AX7101-only subsignal, so a TDM-master build on any other board cannot elaborate


Mister-M-alt:
Found by the elaboration harness landing in #154, on today's `dev` (`c839aa9d`). It is #156 item 3, and this is its cause.

## The defect

`sw/litex/milan_soc.py:5997` binds the TDM master's media-grid output unconditionally:

```python
o_media_lrclk_o = (self.tdm_pads.lrclk if self.tdm_pads
                   else Signal()),
```

`lrclk` is declared by exactly one platform. `sw/litex/platforms/alinx_ax7101.py:199` puts it on J11.9 (ball J16) as a media-grid test point, documented there as an AX7101 probe pin. The other way a board can acquire TDM-shaped pads is `milan_soc.i2s_pmod_io()` at `sw/litex/milan_soc.py:698-703`, which declares `mclk`, `bclk`, `fsync`, `din` and `dout` and no `lrclk`.

So the guard `if self.tdm_pads` asks whether TDM pads exist, and then reads a subsignal only one board's pads carry. Any TDM-master build on a board whose pads come from `i2s_pmod_io` dies during SoC construction:

```
AttributeError: 'Record' object has no attribute 'lrclk'. Did you mean: 'bclk'?
```

The sibling line two above it gets this right: `o_tdm_bclk_o` reads `bclk`, which both declarations carry.

## Reproduce

```sh
python3 sw/builder/test_builder.py     # gate 23g reports `sweep.sh arty` unrunnable
```

or directly, which is the real `sweep.sh arty` command line:

```sh
cd sw/litex && python3 milan_soc.py --board arty --sys-clk-freq 83.333e6 \
  --milan-clk-freq 50e6 --num-streams 4 --audio-interface tdm8 \
  --audio-interface-master --talker-wire-chans 4 --sound-card ... --build
```

## Why nothing saw it

The same reason #154 exists: no gate ever ran a build recipe. Every shape gate compares `sweep.sh` argv against the end-station config and both sides agree, because the argv is right. What is wrong is a per-board resource read as though it were universal, and only elaboration can see that.

Reproduced at `dev` `c839aa9d`, so it is not new.

## Impact, and why the priority is moderated

[...]

Issue #156, opened 2026-08-20
No gate proves a build recipe can actually RUN, only that it matches its config


Mister-M-alt:
## The class

Every shape gate this repo has compares a recipe's argv against the
end-station config it claims to build. **Nothing runs the argv.** A recipe can
agree with its config flag for flag and still be unable to elaborate, and three
separate findings on PR #135 turned out to be the same gap:

1. **`build.sh cfg_ax8x8` and `cfg_arty` could not launch at all.** Neither
   passed `--entity-gen-dir`, without which `milan_soc.py` refuses
   (`this build needs its end-station config`). Reproduced at `dev`
   `10c7bf6c` as well as at the PR head, so it long predates #116. Fixed in
   #135 because "the entry point is runnable again" would otherwise have been
   untrue, and pinned by a new `--entity-gen-dir` check.

2. **A gate built to prove a flag reaches the RTL had the same blind spot one
   level up.** `scripts/check_sweep_shape.py` excludes a `FLOW_FLAGS` set from
   its comparison (Vivado directives, thread cap, output dir, `--build`), and
   `sw/builder/test_builder.py` gate 1e rebuilt its argv from
   `emit_soc_argv`, which never emits them. A late

   ```python
   gptp_plane = args.fabric_gptp and not args.build
   ```

   after every guard in `milan_soc.py` therefore left the whole suite green,
   while the real `sweep.sh ax7101` line (`sweep.sh:121-123`, and
   `build.sh:290`, both end in `--vivado-max-threads 32 --build`) reached
   `Instance("milan_datapath")` with `p_GPTP_PLANE_EN_P=0` and no ROM.
   Measured 2026-08-20. Fixed in #135 by running the full launcher tail.

3. **`sweep.sh arty` cannot elaborate at all**: `AttributeError: 'Record'
   object has no attribute 'lrclk'`, reproduced at the PR head and at `dev`
   with the new flag stripped, so it is not #116's. Priority is moderated
   because the Arty is a retired DUT (USER 2026-08-01: AX7101 is the sole
   DUT), but the whole statement is worth writing down: **the Arty Linux
   comparison arm has no coherent build path.** `sweep.sh arty` cannot
   elaborate it, and `build.sh cfg_arty` builds a different SoC from the one
   `sweep.sh` and `configs/endstation_arty_current.yaml` describe (see #157).

## The fix

Gate 1e's machinery already does the work. It patches `Instance` in the
[...]

PR #188, opened 2026-08-21
Observe the argv to Instance parameter chain instead of grepping for it

### LINKED ISSUE #154: No CI job elaborates the SoC, so argv-to-RTL-parameter chains are unproven
Root cause behind three separate blockers found in three different lanes on the same day. Filing it once rather than a fourth time as an instance.

## The pattern

A build flag is parsed, threaded through some of the chain, and never reaches the RTL parameter it names. Every gate passes, because **no CI job elaborates the SoC**, so every argv-to-parameter chain in this repo is proven by source-text greps against `sw/litex/milan_soc.py` rather than by observing what elaboration actually hands `Instance("milan_datapath")`.

Instances so far:

| lane | flag | consequence if shipped |
|---|---|---|
| #130 | `--no-sound-card` | accepted, never threaded into `optional_blocks`; the build claimed a prune that never happened and the area estimate was fiction |
| #135 | `--fabric-gptp` | nothing ties the parsed flag to `p_GPTP_PLANE_EN_P`; on the bare-metal shipping profile that is **a bitstream with no time source and no signal** |
| #126 (closed) | `--no-sound-card` again | the decorative-flag original |

In the #135 case a reviewer severed the chain at **two independent hops** and got `ALL GATES PASS` both times. In #130's case the same reviewer severed it at `milan_soc.py:6144` and `:538` and reddened **0 of the other 54 gates**.

## Why the existing answers are not enough

- **Gate 23d** (#134) walks the `milan_soc.py` AST across four hops. It is real and it catches a lot. Its own reviewer states its limit precisely: it pins the SPELLING of the statements it walks, so a statement that rebinds or overrides a carrier without touching the pinned ones is invisible by construction. Seven such shapes are known and open.
- **Gate 23f** (#144) is the honest fix: it patches `Instance` in the executed module's namespace and reads the live `p_*` kwargs, so it observes what elaboration really produces. But **it SKIPS in CI**, because `.github/workflows/docs.yml` installs only `pyyaml` and `rtl.yml` only `behave`. Neither has litex or migen. So the one gate that could close this class does not run where it matters.

That is the crux: the behavioural proof exists and CI cannot run it.

## What a fix looks like

Options, none obviously best, and the choice is an infrastructure decision rather than a code change:

1. **Get LiteX into a CI job.** Then 23f runs for real and the class closes properly. Cost: a heavyweight dependency in CI, plus the `sbt`/VexiiRiscv cold-start cost that #144's review measured (bounded to one run, with `PYTHONHASHSEED=0` pinning the cache path).
2. **A hermetic elaboration stub.** Enough of migen/litex faked that `Instance` is reached and its kwargs observed, without the real toolchain. Cheaper to run, but it is a second model of LiteX that will drift from the real one.
3. **Emit the parameter set as a build artifact** and gate on that, so the proof rides the real build even though CI cannot perform one. Weaker, but it makes a shipped bitstream carry its own evidence.

## Acceptance

- [ ] Severing any argv-to-parameter hop for any optional block reddens something that runs in CI.
- [ ] Proven by mutation at more than one hop, and on more than one block, since the #135 case passed at two independent hops.
- [ ] The chosen mechanism states plainly what it does NOT prove, in its own output, the way gate 23d now does.
- [ ] `--fabric-gptp` and every `--no-<block>` flag are covered, not just the ones that had an incident.

Mister-M-alt:
[A1]

Closes #154. Answers #156 items 1 and 2.

## Status

Ready for review. Validated locally; not merged, and it should not be merged before the stacked PR for #185, because on its own the CI job here cannot elaborate anything.

Executor: `[A1]`. Reviewers needed: one cleared-context internal, one external.

## The defect

Every argv-to-RTL-parameter chain in this repository was proven by source-text greps against `sw/litex/milan_soc.py`, because no CI job elaborated the SoC. Three blockers in three lanes on one day were that one gap: a flag parsed, threaded part of the way, never reaching the parameter it names. On #135 a reviewer severed the chain at two independent hops and read `ALL GATES PASS` both times.

## What this does

| | |
|---|---|
| **gate 23f** | runs the shipping AX argv through the real `main()` once per optional block and grades the `p_*` keyword arguments `Instance("milan_datapath", ...)` is handed. Seven `--no-*` flags, plus `--sound-card` and `--fabric-gptp` graded upside down, the gPTP row including its microcode image |
| **gate 23f mutation** | severs the chain at every hop between argparse and the Instance, and requires the gate to go red. Includes both #135 hops, severed independently |
| **gate 23g** | every `build.sh` recipe and `sweep.sh` leg reaches that Instance, with the flow tail its own launcher appends and `--build` included, which is the flag every shape gate in the tree excludes |
| **skip ledger** | `skip()` records each declining arm and the verdict prints `ALL GATES PASS EXCEPT n NOT RUN` with reasons. `--require-elaboration` turns a missing LiteX into a failure rather than a skip |
| **`build.sh`** | `cfg_ax8x8` and `cfg_arty` get the `--entity-gen-dir` they never had. Neither could launch at all |

## Validation

```sh
python3 sw/builder/test_builder.py --require-elaboration
```

- gate 23f: 10 real elaborations of `endstation_ax7101_1x1_tdm8`.
- gate 23f mutation: **11/11** severed links rejected.
- gate 23g: recipes reaching the Instance went from **2/5 to 4/5**.
- gate 23g mutation: **2/2** recipes stripped of `--entity-gen-dir` refused.
- degenerate case run, not assumed: with no LiteX interpreter the command **exits 1** and names the four arms.
- `docs_check`, `check_feature_status --self-test`, `check_doc_paths`, `check_archive`, `check_sweep_shape --self-test`, `check_soc_sources`, `pp_srcs --check --selftest`, `gen_toc --check`, `gen_module_matrix --check`: pass.
- `lint_rtl.py --check`: `99 <= ratchet 99`, exit 0, unchanged. No RTL is touched.

Full evidence, including the exact figures at each commit, is on #154.

## What it does not do

`.github/workflows/elaborate.yml` is here but `workflow_dispatch` only. Upstream LiteX cannot elaborate any config in this tree, so enabling it here would install correctly and then fail on everything. #185 and its stacked PR fix that and enable it; **acceptance criterion 1 of #154 is met only with that branch**, and this Issue should not be closed on this PR alone.

Three defects found on the way out, filed rather than fixed here: **#184** (`o_media_lrclk_o` bound to an AX7101-only subsignal, which is #156 item 3 and its cause), **#185**, **#186** (`lint_rtl` under-counts in a fresh worktree and invites a ratchet cut that reddens `dev`).

## DoD

- [x] Acceptance criteria 2, 3 and 4 of #154 met and evidenced.
- [ ] Acceptance criterion 1: needs the #185 branch.
- [x] Local verification bar green.
- [ ] Two positive reviews, one external.
- [ ] Candidate-merge validation and post-merge containment.

### DIFF
--- a/.github/workflows/elaborate.yml
+++ b/.github/workflows/elaborate.yml
@@ -1,0 +1,127 @@
+name: elaborate
+
+# WHY THIS WORKFLOW EXISTS (issues #154, #156).
+#
+# Every argv-to-RTL-parameter chain in this repository used to be proven by
+# source-text greps against sw/litex/milan_soc.py, because NO CI JOB
+# ELABORATED THE SOC. Three separate blockers in three lanes on one day were
+# that gap: a flag parsed, threaded part of the way, and never reaching the
+# parameter it names. In one of them a reviewer severed the chain at two
+# independent hops and read ALL GATES PASS both times.
+#
+# The behavioural proof existed - gate 23f patches `Instance` in the executed
+# module's namespace and reads the live p_* keyword arguments - and it skipped
+# in CI, because docs.yml installs pyyaml and rtl.yml installs behave and
+# neither has LiteX. That is the whole crux: the proof existed and CI could
+# not run it.
+#
+# So this job installs LiteX and runs it. It is separate from docs.yml because
+# it is the only job in the tree that needs a JVM and a Scala build, and a
+# documentation PR should not pay for that.
+#
+# NOT ENABLED YET, AND THIS FILE IS HERE SO THE REASON IS READABLE. Measured
+# 2026-08-21: NO config in this tree elaborates on a stock toolchain. The
+# shipping AX shape asks for a `baremetal` VexiiRiscv variant upstream LiteX
+# does not have; the four Linux shapes pass --scala-args the pinned VexiiRiscv
+# does not accept. Both come from about 184 lines of uncommitted patch across
+# four third-party trees under ~/litex-milan, which #185 measures and whose
+# fix is a maintainer decision. Until that decision is made this job would
+# install correctly and then fail on every config, so its triggers are
+# withheld and it runs only on manual dispatch.
+#
+# Gates 23f and 23g run today on any interpreter that has a working LiteX,
+# which is how they were validated. Everywhere else they SKIP, the verdict
+# says `ALL GATES PASS EXCEPT n NOT RUN` and names them, and
+# --require-elaboration turns a missing interpreter into a failure rather
+# than a skip. That is what stops the absence of the proof from reading as
+# the presence of one while this job is off.
+
+on:
+  # Manual only until #185 is decided. The push/pull_request triggers below
+  # are the intended ones and are commented rather than deleted, so enabling
+  # this job is one edit and not a rediscovery.
+  workflow_dispatch:
+  # push:
+  #   branches: [main, dev]
+  # pull_request:
+
+concurrency:
+  group: elaborate-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  elaborate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      # The elaboration reads the protocol-processor and gptp-processor source
+      # lists through scripts/pp_srcs.py, which REFUSES to emit an empty list.
+      # Without the submodules it fails before the datapath is reached.
+      - name: Fetch the processor sources the elaboration reads
+        run: git submodule update --init protocol-processor gptp-processor
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      # LiteX ships from git, not PyPI: the PyPI releases lag by more than a
+      # year and their VexiiRiscv wrapper does not carry the pinned-revision
+      # setup this job depends on. The revisions are in sw/litex/litex_pins.txt
+      # with the reason they are pinned; master of 2026-08-21 cannot elaborate
+      # this SoC at all, so an unpinned install would redden PRs that changed
+      # nothing.
+      - name: Install LiteX at the pinned revisions
+        run: |
+          python3 -m pip install --quiet pyyaml
+          python3 -m pip install --quiet -r sw/litex/litex_pins.txt
+
+      # The VexiiRiscv wrapper turns its CPU arguments into a small generated
+      # Python file, once per distinct argument set, by running Scala. That is
+      # the only reason this job needs a JVM at all, and the result caches: a
+      # cold run pays about 90 s for it, a warm one nothing.
+      - name: Cache the CPU netlist arguments and the Scala toolchain
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/coursier
+            ~/.sbt
+            ~/.ivy2
+            ~/vexii-netlist-args
+          key: elaborate-vexii-${{ hashFiles('configs/*.yaml', 'sw/litex/milan_soc.py', 'sw/litex/build.sh', 'sw/litex/sweep.sh') }}
+          restore-keys: elaborate-vexii-
+
+      - name: Install sbt
+        run: |
+          sbt --version >/dev/null 2>&1 && exit 0
+          curl -fsSL --retry 3 -o /tmp/sbt.tgz \
+            https://github.com/sbt/sbt/releases/download/v1.10.7/sbt-1.10.7.tgz
+          sudo tar xzf /tmp/sbt.tgz -C /opt
+          echo /opt/sbt/bin >> "$GITHUB_PATH"
+
+      # LiteX's own git_setup returns immediately unless --update-repo asks it
+      # to clone, and milan_soc.py never passes that, so the Scala source has
+      # to be placed here. The revision is DERIVED from LiteX rather than
+      # restated: a pin copied to a second place goes stale in silence, and
+      # the symptom is a netlist from the wrong CPU source, not a red build.
+      - name: Place the VexiiRiscv source at the revision LiteX pins
+        run: |
+          python3 scripts/ci_litex_env.py
+          mkdir -p ~/vexii-netlist-args
+          cp -n ~/vexii-netlist-args/*.py \
+            "$(python3 -c 'import pythondata_cpu_vexiiriscv as p; print(p.data_location)')/" \
+            2>/dev/null || true
+
+      # The gate that could not run in CI, running in CI. Gates 23f/23g
+      # observe what elaboration hands Instance("milan_datapath"); their
+      # mutation arms sever the chain hop by hop and require the gate to go
+      # red. --require-elaboration turns a missing interpreter into a failure
+      # rather than a skip.
+      - name: Elaboration gates
+        run: python3 sw/builder/test_builder.py --require-elaboration
+
+      - name: Keep the generated CPU netlist arguments for the next run
+        if: always()
+        run: |
+          cp -n "$(python3 -c 'import pythondata_cpu_vexiiriscv as p; print(p.data_location)')"/*.py \
+            ~/vexii-netlist-args/ 2>/dev/null || true
--- a/.github/workflows/elaborate.yml
+++ b/.github/workflows/elaborate.yml
@@ -23,11 +23,11 @@ name: elaborate
 # 2026-08-21: NO config in this tree elaborates on a stock toolchain. The
 # shipping AX shape asks for a `baremetal` VexiiRiscv variant upstream LiteX
 # does not have; the four Linux shapes pass --scala-args the pinned VexiiRiscv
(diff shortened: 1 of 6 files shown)


Mister-M-alt (commented):
[R0] NEGATIVE — exact head `be721ad1e18f53e2fc9331cc8d771116c4a852f9`

Lenses covered: Conformance, RTL, Robustness, Tests, Docs.

[R0] BLOCKER Conformance — PR body / issue #154 acceptance criterion 1 — this head is explicitly incomplete but the stack plan requires it to land first
Requirement/evidence: #154 freezes “severing any argv-to-parameter hop reddens something that runs in CI.” This PR leaves `elaborate.yml` manual-only and its own DoD marks that criterion unmet. PR #189 then says it will be retargeted only after #188 lands, so the documented sequence necessarily puts an acceptance-incomplete `Closes #154` head on `dev`.
Impact: the lane cannot satisfy the repository rule that all Issue acceptance criteria are met at the reviewed merge head; it also creates a transient `dev` state the PR itself says must not merge.
Required change: make the reviewed merge candidate contain the enabled, reproducible elaboration job before this Issue-closing lane lands (for example, retarget the combined stack to `dev` and avoid merging this intermediate head), then update both PR descriptions to state one coherent merge order.
Verification: the reviewed head itself must run the elaboration job in CI, reject a severed hop, and mark all #154 criteria met.

[R0] MAJOR Tests, Robustness — `sw/builder/test_builder.py:4456-4590,6542-6564` — skip classification still converts unrelated failures into a green
Requirement/evidence: a `KNOWN_UNRUNNABLE` row accepts any error without checking the documented `AttributeError`/`lrclk` cause, and `_is_unpatched_vexii()` accepts any error containing `PythonArgsGen` when an L2 argument is present. Both use the ordinary skip kind, while `--require-elaboration` fails only `NO_LITEX`. On the exact combined run the verdict says `sweep.sh arty` “DID NOT RUN” and immediately prints `--require-elaboration: every elaboration arm ran`.
Impact: a new parser, Scala, tool-install, or recipe failure can be misreported as the old waiver and leave CI green; after #184 is fixed, a different failure can keep its stale table entry alive forever. The printed execution claim is already false on the current head.
Required change: pin every permitted skip to its exact diagnostic and make all unexpected errors fatal. Under the enabled job, a Vexii patch/setup failure must fail rather than become a row skip. Render the final verdict from the actual rows run instead of claiming every elaboration arm ran.
Verification: mutate the recorded Arty failure to a different exception and mutate `PythonArgsGen` to fail for a non-option reason; both must exit non-zero under `--require-elaboration`. The unmodified run must describe the Arty exception and its remaining skip truthfully.

[R0] MAJOR Tests — `.github/workflows/elaborate.yml:83-112` — cached Vexii metadata is not keyed to the source that generated it
Requirement/evidence: the workflow key hashes configs, `milan_soc.py`, `build.sh`, and `sweep.sh`, but not `litex_pins.txt` or the patch series. Its broad `restore-keys: elaborate-vexii-` can restore an older cache even after a key change, and `cp -n` installs those old `.py` files. The pinned LiteX source names each metadata file from `interface-metadata-v1 + metadata_args` only and skips `PythonArgsGen` whenever that file exists, so a pin/patch change can execute stale generated metadata.
Impact: CI can report that current toolchain patches elaborate while actually bypassing the changed Scala generator with an artifact from an earlier head. A patch mutation that removes an argument parser can stay green once a prior cache exists.
Required change: separate the reusable SBT/Coursier cache from generated CPU metadata, and either regenerate metadata or key/restore it strictly by every source input (pins and patches included) without a broad stale fallback.
Verification: populate a cache, change a Vexii patch/pin without changing argv, and prove the next run invokes current `PythonArgsGen` or rejects the stale artifact.

[R0] MINOR Docs — `scripts/ci_litex_env.py:31-35` and `sw/builder/test_builder.py:4061-4075` — the public explanation still says the needed code is an uncommitted local patch
Requirement/evidence: issue #185’s correction and this PR’s second commit establish that `sw/litex/patches/` already carries the series; the defect is that nothing applied/verified it. These two current-head explanations still state that the Linux shapes build only against an “UNCOMMITTED” local patch.
Impact: a cold maintainer is directed to the wrong root cause and may look for private source that the repository already owns.
Required change: describe the repo-carried-but-unapplied series consistently in the script and gate commentary.
Verification: search the final stack for the superseded uncommitted-patch claim and reconcile every remaining occurrence with #185’s recorded correction.

[R0] PASS RTL — `git diff c839aa9d...be721ad1 --name-status` — no repository RTL, CDC, reset, width, or datapath interface implementation is changed by this PR; the touched runtime surface is LiteX glue/build verification.

Mister-M-alt:
[A1] Two defects in this PR, both found by reviewing somebody else's, both fixed at `106df723`.

**1. The Instance spy was never proved to STOP the elaboration.** Gate 1e on PR #135 asserts its `--build` output directory is still empty afterwards, with the reasoning next to it: if the spy raised too late, LiteX would have written its gateware and software trees there and Vivado would have been invoked, and every parameter assertion would still have passed. My gates asserted the spy *reached* the Instance and never that it stopped there.

Adopted, and proved live rather than assumed. Neutering the spy so it witnesses the Instance without raising:

```
ASSERTION FIRED: the Instance spy no longer stops the elaboration: a run carrying
--build wrote ['software', 'gateware', 'litex.log'] into its output directory
```

Gate 23g's `--output-dir` now resolves inside the directory the harness checks, rather than a temp path the check could not see. Credit to [A0]'s gate 1e; recorded on #135 too.

**2. `elaborate.yml` would have made every documentation-only PR pay a LiteX install, a VexiiRiscv checkout and a Scala run.** PR #176 introduces `scripts/ci_scope.py`, the repository's one answer to whether a change needs the RTL toolchain. This workflow now routes through it. It fails safe twice: `ci_scope.py` itself returns `true` for an empty diff or an unresolvable base, and if the file is absent - which it is until #176 lands - this workflow elaborates unconditionally rather than skipping.

Re-validated at `106df723`: builder suite exit 0, `--require-elaboration: every elaboration arm ran`, `ALL GATES PASS EXCEPT 3 NOT RUN` with the three named. Documentation gates pass. PR #189 rebased onto this head and re-validated the same way.

Neither finding came from my own lane. Both came from reading two other PRs, which is the argument for the external-reviewer half of the bar rather than for more rounds of my own.

Mister-M-alt:
[A1] CI evidence, run `32479289006`: **18 of 19 checks pass.**

The one red is `yosys-portability`, and **it is not this PR's**. Its log is the breakage `dev` has carried since #176 merged:

```
protocol-processor/hdl not found. The protocol-processor submodule is not checked out...
Process completed with exit code 2
```

Diagnosed in **#190** and fixed in **PR #191**, which is green on all 19 checks including that one. The two runs are a controlled pair: same base, same workflows, and the only difference is the one-file change in #191.

Everything this PR is responsible for is green, including the four Verilator shards, `verilator-suites`, all four Yosys shards, `verilator-lint`, `yosys-elaboration`, `rtl-fast`, `bdd-conformance`, `wire-accountability` and both docs jobs.

Not rebasing onto #191 yet: that would put another lane's unmerged commit in this diff. Once #191 lands, this branch rebases and the check goes green with nothing else changing. **Merge order: #191, then this, then #189.**

Mister-M-alt:
[R1] Review and fix for this stack are on PR #189, which now carries this branch's four commits rebased onto current `dev` (#191 included) and is retargeted to `dev`, per [R0]'s BLOCKER above: the stack lands as ONE reviewed candidate, and this intermediate head does not merge on its own.

[R0]'s MAJOR on this PR (skip classification) had not been answered anywhere in the stack; it is fixed on #189 at `95f599ba` (head `c0bf1f2b`) and the evidence is there. This PR stays open until #189 merges, then closes as landed-via-#189 with every commit of `154-no-ci-job-...` contained in that merge; `Closes #154` moves with it.

Mister-M-alt:
[R1] Landed via PR #189, merged into `dev` at `2823480203955e1f7993e71dd2ab2335e0f74cc7` on the maintainer's instruction. This PR closes as landed-via-#189, per [R0]'s BLOCKER above: the stack merged as one reviewed candidate and this intermediate head never merged on its own.

Every commit of this branch is in that merge by content: `git cherry c0bf1f2b b34e6cc5 c839aa9d` marks all four (`bb910891`, `be721ad1`, `106df723`, `b34e6cc5`) as patch-equivalent to commits in the candidate (they were replayed onto `dev` by the rebase, conflict-free). The SHAs on this branch are therefore not ancestors of `dev`, which is the shape `scripts/check_merge_containment.py` reads as a strand; it is not one, and this comment is the record of why.

[R0]'s MAJOR on this PR (skip classification) was fixed on #189 at `95f599ba` and is exercised by the hosted `elaborate` job (`[gate 23g classifier] 8/8 shaped results and 4/4 --require-elaboration verdicts`). `Closes #154` moved with the stack.

PR #176, opened 2026-08-21
[A0] Split fast and full RTL CI

Mister-M-alt:
[A0]

## Status

DRAFT — stacked on the Yosys-sharding branch. Static workflow/documentation checks and real 46-top fast-Yosys execution pass locally; live event behavior and independent reviews remain.

## Linked Issue / roles

Closes #175.

Executor: `[A0]`
Internal cleared-context reviewer: `[R1]`
External reviewer: `[R0]`

## Description

- Keep `.github/workflows/rtl.yml` as a stable fast PR workflow ending in `rtl-fast`.
- Classify documentation-only changes conservatively; unknown paths run RTL checks.
- Require whole-tree Verilator lint, BDD conformance, and all-top Yosys hierarchy elaboration for relevant PR updates.
- Move the complete Verilator/Yosys inventories to `.github/workflows/rtl-full.yml`.
- Run full evidence on `ready_for_review`, push to `dev`, manual dispatch, and nightly against `dev`.
- Preserve `verilator-suites` and `yosys-portability` as the full aggregate names.
- Cache the pinned `tsn-gen` packet generator in the full Verilator workers.
- Document the draft -> fast feedback -> ready-for-review -> full evidence procedure.

## Why this is stacked

The full workflow includes the four-worker Yosys implementation from #171. This PR targets `171-shard-yosys-portability` so its diff contains only CI tiering, caching, and documentation. After #171 lands it can be retargeted to `dev` without mixing review scopes.

## How to validate

```sh
python3 scripts/yosys_shards.py --selftest
python3 scripts/yosys_tally.py --selftest
bash -n syn/yosys/run.sh
YOSYS_SYNTH='hierarchy -check' syn/yosys/run.sh --tops-only
python3 scripts/docs_check.py
python3 scripts/check_doc_paths.py
python3 scripts/gen_toc.py --verify-anchors
python3 scripts/gen_toc.py --check
```

Measured locally:

- Fast Yosys mode lowered and elaborated all 46 authoritative tops with 46 PASS and 0 FAIL.
- Four fast-mode shard logs reconciled to all 46 tops exactly once.
- Tied-input and tap-purity PASS evidence occurred exactly once across those shards.

Workflow scenarios to validate live:

1. Documentation-only draft update: `rtl-fast` succeeds and heavy setup is skipped.
2. RTL-relevant draft update: lint, BDD and Yosys elaboration feed `rtl-fast`.
3. Mark Ready for review: full Verilator and Yosys workflows run.
[...]

PR #189, opened 2026-08-21
Apply and verify the toolchain patch series instead of trusting it

### LINKED ISSUE #185: No configuration in this tree elaborates on a stock toolchain: four third-party trees carry uncommitted patches

Mister-M-alt:
[A1]

Closes #185. Unblocks #154's acceptance criterion 1.

## Status

Ready for review. **Retargeted to `dev` at `c0bf1f2b`** and rebased onto `dev@9f79dcec` (#191 included): this PR now carries #188's four commits plus the toolchain work plus the [R1] fix, so the reviewed head IS the merge candidate and #188 does not merge on its own ([R0]'s BLOCKER on #188). #188 closes as landed-via-#189 when this merges.

Executor: `[A1]`. Reviewers needed: one cleared-context internal, one external.

## The defect

Upstream LiteX cannot build any shape of this design. It has no `baremetal` VexiiRiscv variant, which is the shipping AX profile, and the VexiiRiscv revision it pins rejects the `--l2-*` arguments four of the five configs pass. `sw/litex/patches/` has always carried the series that closes both, and `apply.sh` applies it.

**Nothing ran it.** The only reader was a person following a document, so the drift had no upper bound on its age:

- `0003` was diffed against pristine upstream while `0001` rewrites the same `boot.c` hunks, so `apply.sh` aborted on the third patch and `0004` and `0005` never ran. Every patch applies **individually**, which is why checking one looked fine.
- `0002-liteeth` had fallen 25 lines behind the tree the boards are built from: the `ext_reset` input `KL_link_guard` drives lived only in the bench's working copy.
- `0002-vexiiriscv-l2-depth-args` was documented as "apply it by hand for a non-default L2". True once; four configs now pass those arguments, so `apply.sh` could succeed and still leave a tree elaborating none of them.
- `apply.sh` was not idempotent, contrary to its own header, because it graded each patch of a stacked series independently.

## What this does

- `0003` rebased onto `0001`; `0002-liteeth` refreshed. Both carry a header stating what moved.
- `apply.sh` **normalises then applies**: reverse the series best-effort, which reaches pristine from any partially-applied state, then apply forward and fail hard on the first refusal. Applies the L2 patch too, and refuses with a pointer to `scripts/ci_litex_env.py` when the VexiiRiscv checkout is absent instead of reporting success for a patch that reached nothing.
- **gate 23h**: reverses the series off the installed trees, re-applies it, requires byte-identical output. Reconstruction rather than inspection, because a stacked series cannot be graded one patch at a time. That is the same mistake `apply.sh` made in the other direction.
- `elaborate.yml` triggers enabled, with `apply.sh` as a step.
- `LITEX_SOC.md`, `patches/README.md`, `litex_pins.txt`: this is a required install step, not a tuning one, and the install is three commands.

## How to validate

```sh
python3 -m venv /tmp/v && /tmp/v/bin/pip install pyyaml -r sw/litex/litex_pins.txt
/tmp/v/bin/python scripts/ci_litex_env.py
PYTHON=/tmp/v/bin/python sw/litex/patches/apply.sh
/tmp/v/bin/python sw/builder/test_builder.py --require-elaboration
```

- environment built in **42 s**, byte-identical to the bench across all 8 patched files, checked file by file.
- builder suite **exit 0**: 2 min 18 s cold, 45 s warm.
- gate 23h: 6 patches, 8 files reconstructed. gate 23h mutation: **6/6** patches removed one at a time were each rejected.
- gate 23f: 10 elaborations. gate 23f mutation: **11/11**. gate 23g: **4/5**. gate 23g mutation: **2/2**.
- `ALL GATES PASS EXCEPT 3 NOT RUN`, the three named: two absent build trees and #184's Arty leg. No `no-litex` skip, so `--require-elaboration` is satisfied.
- `apply.sh` run twice in a row: second run is a clean no-op.
- Nine documentation and shape gates pass. Lint `99 <= ratchet 99`, exit 0, unchanged from `dev`. No RTL touched.

## What it does not do

It does not claim the patches are **correct**, only that the tree being elaborated is the one this repository describes. Upstreaming to LiteX and SpinalHDL is still worth doing and is not blocked by this. `sweep.sh arty` remains unrunnable against #184.

## DoD

- [x] Acceptance criteria of #185 met and evidenced.
- [x] Local verification bar green.
- [ ] Two positive reviews, one external.
- [x] Retarget to `dev` (done at `c0bf1f2b`; #188's commits are in this branch).
- [ ] Candidate-merge validation and post-merge containment.

## Update at `c0bf1f2b` ([R1], 2026-08-21)

[R0]'s five findings on this PR re-verified independently at `32dd8c24` and held as answered. [R0]'s MAJOR on #188 (gate 23g skipped a recorded row on any error, skipped any `PythonArgsGen` failure as #185's, and `--require-elaboration` claimed every arm ran while one was listed as not run) had not been answered anywhere in the stack; fixed at `95f599ba` with exact diagnostics, a `toolchain` skip kind the CI flag refuses, a truthful verdict and a LiteX-free self-test, mutation-proven. `docs/testing/CI_WORKFLOWS.md` gains the `elaborate` row at `c0bf1f2b`. Findings and evidence: the [R1] review on this PR.

### DIFF
--- a/.github/workflows/elaborate.yml
+++ b/.github/workflows/elaborate.yml
@@ -1,0 +1,127 @@
+name: elaborate
+
+# WHY THIS WORKFLOW EXISTS (issues #154, #156).
+#
+# Every argv-to-RTL-parameter chain in this repository used to be proven by
+# source-text greps against sw/litex/milan_soc.py, because NO CI JOB
+# ELABORATED THE SOC. Three separate blockers in three lanes on one day were
+# that gap: a flag parsed, threaded part of the way, and never reaching the
+# parameter it names. In one of them a reviewer severed the chain at two
+# independent hops and read ALL GATES PASS both times.
+#
+# The behavioural proof existed - gate 23f patches `Instance` in the executed
+# module's namespace and reads the live p_* keyword arguments - and it skipped
+# in CI, because docs.yml installs pyyaml and rtl.yml installs behave and
+# neither has LiteX. That is the whole crux: the proof existed and CI could
+# not run it.
+#
+# So this job installs LiteX and runs it. It is separate from docs.yml because
+# it is the only job in the tree that needs a JVM and a Scala build, and a
+# documentation PR should not pay for that.
+#
+# NOT ENABLED YET, AND THIS FILE IS HERE SO THE REASON IS READABLE. Measured
+# 2026-08-21: NO config in this tree elaborates on a stock toolchain. The
+# shipping AX shape asks for a `baremetal` VexiiRiscv variant upstream LiteX
+# does not have; the four Linux shapes pass --scala-args the pinned VexiiRiscv
+# does not accept. Both come from about 184 lines of uncommitted patch across
+# four third-party trees under ~/litex-milan, which #185 measures and whose
+# fix is a maintainer decision. Until that decision is made this job would
+# install correctly and then fail on every config, so its triggers are
+# withheld and it runs only on manual dispatch.
+#
+# Gates 23f and 23g run today on any interpreter that has a working LiteX,
+# which is how they were validated. Everywhere else they SKIP, the verdict
+# says `ALL GATES PASS EXCEPT n NOT RUN` and names them, and
+# --require-elaboration turns a missing interpreter into a failure rather
+# than a skip. That is what stops the absence of the proof from reading as
+# the presence of one while this job is off.
+
+on:
+  # Manual only until #185 is decided. The push/pull_request triggers below
+  # are the intended ones and are commented rather than deleted, so enabling
+  # this job is one edit and not a rediscovery.
+  workflow_dispatch:
+  # push:
+  #   branches: [main, dev]
+  # pull_request:
+
+concurrency:
+  group: elaborate-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  elaborate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      # The elaboration reads the protocol-processor and gptp-processor source
+      # lists through scripts/pp_srcs.py, which REFUSES to emit an empty list.
+      # Without the submodules it fails before the datapath is reached.
+      - name: Fetch the processor sources the elaboration reads
+        run: git submodule update --init protocol-processor gptp-processor
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      # LiteX ships from git, not PyPI: the PyPI releases lag by more than a
+      # year and their VexiiRiscv wrapper does not carry the pinned-revision
+      # setup this job depends on. The revisions are in sw/litex/litex_pins.txt
+      # with the reason they are pinned; master of 2026-08-21 cannot elaborate
+      # this SoC at all, so an unpinned install would redden PRs that changed
+      # nothing.
+      - name: Install LiteX at the pinned revisions
+        run: |
+          python3 -m pip install --quiet pyyaml
+          python3 -m pip install --quiet -r sw/litex/litex_pins.txt
+
+      # The VexiiRiscv wrapper turns its CPU arguments into a small generated
+      # Python file, once per distinct argument set, by running Scala. That is
+      # the only reason this job needs a JVM at all, and the result caches: a
+      # cold run pays about 90 s for it, a warm one nothing.
+      - name: Cache the CPU netlist arguments and the Scala toolchain
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/coursier
+            ~/.sbt
+            ~/.ivy2
+            ~/vexii-netlist-args
+          key: elaborate-vexii-${{ hashFiles('configs/*.yaml', 'sw/litex/milan_soc.py', 'sw/litex/build.sh', 'sw/litex/sweep.sh') }}
+          restore-keys: elaborate-vexii-
+
+      - name: Install sbt
+        run: |
+          sbt --version >/dev/null 2>&1 && exit 0
+          curl -fsSL --retry 3 -o /tmp/sbt.tgz \
+            https://github.com/sbt/sbt/releases/download/v1.10.7/sbt-1.10.7.tgz
+          sudo tar xzf /tmp/sbt.tgz -C /opt
+          echo /opt/sbt/bin >> "$GITHUB_PATH"
+
+      # LiteX's own git_setup returns immediately unless --update-repo asks it
+      # to clone, and milan_soc.py never passes that, so the Scala source has
+      # to be placed here. The revision is DERIVED from LiteX rather than
+      # restated: a pin copied to a second place goes stale in silence, and
+      # the symptom is a netlist from the wrong CPU source, not a red build.
+      - name: Place the VexiiRiscv source at the revision LiteX pins
+        run: |
+          python3 scripts/ci_litex_env.py
+          mkdir -p ~/vexii-netlist-args
+          cp -n ~/vexii-netlist-args/*.py \
+            "$(python3 -c 'import pythondata_cpu_vexiiriscv as p; print(p.data_location)')/" \
+            2>/dev/null || true
+
+      # The gate that could not run in CI, running in CI. Gates 23f/23g
+      # observe what elaboration hands Instance("milan_datapath"); their
+      # mutation arms sever the chain hop by hop and require the gate to go
+      # red. --require-elaboration turns a missing interpreter into a failure
+      # rather than a skip.
+      - name: Elaboration gates
+        run: python3 sw/builder/test_builder.py --require-elaboration
+
+      - name: Keep the generated CPU netlist arguments for the next run
+        if: always()
+        run: |
+          cp -n "$(python3 -c 'import pythondata_cpu_vexiiriscv as p; print(p.data_location)')"/*.py \
+            ~/vexii-netlist-args/ 2>/dev/null || true
--- a/.github/workflows/elaborate.yml
+++ b/.github/workflows/elaborate.yml
@@ -23,11 +23,11 @@ name: elaborate
 # 2026-08-21: NO config in this tree elaborates on a stock toolchain. The
 # shipping AX shape asks for a `baremetal` VexiiRiscv variant upstream LiteX
 # does not have; the four Linux shapes pass --scala-args the pinned VexiiRiscv
-# does not accept. Both come from about 184 lines of uncommitted patch across
-# four third-party trees under ~/litex-milan, which #185 measures and whose
-# fix is a maintainer decision. Until that decision is made this job would
-# install correctly and then fail on every config, so its triggers are
-# withheld and it runs only on manual dispatch.
+# does not accept. sw/litex/patches carries the series that closes both, and
+# apply.sh applies it, but nothing in CI runs that script and no gate compares
+# its result - so it had stopped applying, which #185 measures. Until that is
+# fixed this job would install correctly and then fail on every config, so its
+# triggers are withheld and it runs only on manual dispatch.
 #
 # Gates 23f and 23g run today on any interpreter that has a working LiteX,
 # which is how they were validated. Everywhere else they SKIP, the verdict
@@ -65,12 +65,13 @@ jobs:
         with:
           python-version: '3.12'
 
-      # LiteX ships from git, not PyPI: the PyPI releases lag by more than a
-      # year and their VexiiRiscv wrapper does not carry the pinned-revision
-      # setup this job depends on. The revisions are in sw/litex/litex_pins.txt
-      # with the reason they are pinned; master of 2026-08-21 cannot elaborate
-      # this SoC at all, so an unpinned install would redden PRs that changed
-      # nothing.
+      # LiteX ships from git, not PyPI, and so does migen: the PyPI migen wheel
+      # and the m-labs git tree share the version number 0.9.2 and are
+      # different code, and only the git one can walk Python 3.11+ bytecode in
+      # its CSR-name tracer. Installed from PyPI the SoC dies at csr.py:64,
+      # measured 2026-08-21. pythondata-cpu-vexiiriscv is not on PyPI at all.
+      # The revisions are pinned in sw/litex/litex_pins.txt so that an
+      # upstream commit cannot redden a PR that changed nothing.
       - name: Install LiteX at the pinned revisions
         run: |
           python3 -m pip install --quiet pyyaml
--- a/.github/workflows/elaborate.yml
+++ b/.github/workflows/elaborate.yml
@@ -118,7 +118,19 @@ jobs:
       # Python file, once per distinct argument set, by running Scala. That is
       # the only reason this job needs a JVM at all, and the result caches: a
       # cold run pays about 90 s for it, a warm one nothing.
-      - name: Cache the CPU netlist arguments and the Scala toolchain
+      # TWO CACHES, NOT ONE, and the split is the correctness fix rather than a
+      # tidy-up. The Scala toolchain is content-addressed by coursier and safe
+      # to share across every revision, so it keeps a broad fallback. The
+      # GENERATED CPU METADATA is not: LiteX names those .py files from
+      # `interface-metadata-v1 + metadata_args` - the CPU ARGUMENTS - and
+      # skips PythonArgsGen entirely when the file already exists
+      # (litex/soc/cores/cpu/vexiiriscv/core.py). Nothing in that name
+      # covers the Scala source. So a cache entry made before a pin or patch
+      # changed still satisfies the lookup, and the job elaborates metadata
+      # built by the OLD generator while reporting green - the one component
+      # this workflow exists to prove would stop being from-scratch ([R0] on
+      # PR #189).
+      - name: Cache the Scala toolchain
         if: ${{ steps.scope.outputs.rtl == 'true' }}
         uses: actions/cache@v4
         with:
@@ -126,9 +138,20 @@ jobs:
             ~/.cache/coursier
             ~/.sbt
             ~/.ivy2
-            ~/vexii-netlist-args
-          key: elaborate-vexii-${{ hashFiles('configs/*.yaml', 'sw/litex/milan_soc.py', 'sw/litex/build.sh', 'sw/litex/sweep.sh') }}
-          restore-keys: elaborate-vexii-
+          key: elaborate-sbt-${{ runner.os }}
+          restore-keys: elaborate-sbt-
+
+      # NO restore-keys HERE, deliberately. A prefix fallback is exactly the
+      # mechanism that hands over .py metadata from another toolchain head; a
+      # miss must cost one ~90 s Scala run, never a silent wrong answer. The
+      # key covers every input that can change what the generator emits: the
+      # argv sources, the pinned revisions, and the patch series itself.
+      - name: Cache the generated CPU metadata
+        if: ${{ steps.scope.outputs.rtl == 'true' }}
+        uses: actions/cache@v4
+        with:
+          path: ~/vexii-netlist-args
+          key: elaborate-vexii-${{ hashFiles('configs/*.yaml', 'sw/litex/milan_soc.py', 'sw/litex/build.sh', 'sw/litex/sweep.sh', 'sw/litex/litex_pins.txt', 'sw/litex/patches/**') }}
 
       - name: Install sbt
         if: ${{ steps.scope.outputs.rtl == 'true' }}
@@ -162,6 +185,7 @@ jobs:
       # `baremetal` variant and the --l2-* arguments exist; gate 23h proves
       # the result is upstream plus exactly these patches.
       - name: Apply the toolchain patch series
+        if: ${{ steps.scope.outputs.rtl == 'true' }}
         run: sw/litex/patches/apply.sh
 
       # The gates that could not run in CI, running in CI. 23f/23g observe
--- a/.github/workflows/elaborate.yml
+++ b/.github/workflows/elaborate.yml
@@ -32,10 +32,13 @@ name: elaborate
 # is about three minutes, most of it one Scala run per distinct CPU argument
 # set; warm it is under one.
 #
-# --require-elaboration fails the job when no LiteX interpreter is found,
-# rather than skipping the gates and reporting the green this workflow exists
-# to end. Everywhere else those gates SKIP and the verdict says
-# `ALL GATES PASS EXCEPT n NOT RUN` and names them.
+# --require-elaboration fails the job when no LiteX interpreter is found, and
+# when the one found carries a VexiiRiscv that rejects the --l2-* arguments
+# the series adds (the generator prints scopt's `Unknown option`), rather
+# than skipping the gates and reporting the green this workflow exists to
+# end. A row a gate declines for a RECORDED reason (#184's Arty leg) is named
+# in the verdict, not absorbed. Everywhere else those gates SKIP and the
+# verdict says `ALL GATES PASS EXCEPT n NOT RUN` and names them.
 
 on:
   push:
@@ -179,8 +182,10 @@ jobs:
       # The gate that could not run in CI, running in CI. Gates 23f/23g
       # observe what elaboration hands Instance("milan_datapath"); their
       # mutation arms sever the chain hop by hop and require the gate to go
-      # red. --require-elaboration turns a missing interpreter into a failure
-      # rather than a skip.
+      # red. --require-elaboration turns a missing interpreter, and an
+      # interpreter whose VexiiRiscv is unpatched, into a failure rather
+      # than a skip; only a row with a recorded, exactly matching failure
+      # may skip.
       # Upstream cannot build this design. The series is what makes the
       # `baremetal` variant and the --l2-* arguments exist; gate 23h proves
       # the result is upstream plus exactly these patches.
--- a/.github/workflows/elaborate.yml
+++ b/.github/workflows/elaborate.yml
@@ -165,8 +165,7 @@ jobs:
         run: python3 sw/builder/test_builder.py --require-elaboration
 
       - name: Keep the generated CPU netlist arguments for the next run
-        if: ${{ steps.scope.outputs.rtl == 'true' }}
-        if: always()
+        if: ${{ always() && steps.scope.outputs.rtl == 'true' }}
         run: |
           cp -n "$(python3 -c 'import pythondata_cpu_vexiiriscv as p; print(p.data_location)')"/*.py \
             ~/vexii-netlist-args/ 2>/dev/null || true
--- a/.github/workflows/elaborate.yml
+++ b/.github/workflows/elaborate.yml
@@ -19,31 +19,29 @@ name: elaborate
 # it is the only job in the tree that needs a JVM and a Scala build, and a
 # documentation PR should not pay for that.
 #
-# NOT ENABLED YET, AND THIS FILE IS HERE SO THE REASON IS READABLE. Measured
-# 2026-08-21: NO config in this tree elaborates on a stock toolchain. The
-# shipping AX shape asks for a `baremetal` VexiiRiscv variant upstream LiteX
-# does not have; the four Linux shapes pass --scala-args the pinned VexiiRiscv
-# does not accept. sw/litex/patches carries the series that closes both, and
-# apply.sh applies it, but nothing in CI runs that script and no gate compares
-# its result - so it had stopped applying, which #185 measures. Until that is
-# fixed this job would install correctly and then fail on every config, so its
-# triggers are withheld and it runs only on manual dispatch.
+# NO CONFIG IN THIS TREE ELABORATES ON A STOCK TOOLCHAIN (#185), so this job
+# does not install one. Upstream LiteX has no `baremetal` VexiiRiscv variant,
+# which is the whole shipping AX profile, and the VexiiRiscv revision it pins
+# rejects the --l2-* arguments four of the five configs pass.
+# sw/litex/patches/ carries the six patches that close both, and apply.sh
+# applies them. Until 2026-08-21 nothing ran that script end to end and the
+# series had stopped applying; gate 23h now reconstructs it and compares.
 #
-# Gates 23f and 23g run today on any interpreter that has a working LiteX,
-# which is how they were validated. Everywhere else they SKIP, the verdict
-# says `ALL GATES PASS EXCEPT n NOT RUN` and names them, and
-# --require-elaboration turns a missing interpreter into a failure rather
-# than a skip. That is what stops the absence of the proof from reading as
-# the presence of one while this job is off.
+# The whole job is therefore: pinned revisions, the VexiiRiscv checkout at the
+# revision LiteX itself names, the patch series, and then the gates. Cold it
+# is about three minutes, most of it one Scala run per distinct CPU argument
+# set; warm it is under one.
+#
+# --require-elaboration fails the job when no LiteX interpreter is found,
+# rather than skipping the gates and reporting the green this workflow exists
+# to end. Everywhere else those gates SKIP and the verdict says
+# `ALL GATES PASS EXCEPT n NOT RUN` and names them.
 
 on:
-  # Manual only until #185 is decided. The push/pull_request triggers below
-  # are the intended ones and are commented rather than deleted, so enabling
-  # this job is one edit and not a rediscovery.
+  push:
+    branches: [main, dev]
+  pull_request:
   workflow_dispatch:
-  # push:
-  #   branches: [main, dev]
-  # pull_request:
 
 concurrency:
   group: elaborate-${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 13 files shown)


Mister-M-alt (commented):
[R0] NEGATIVE — exact head `3e21b7b15de4c5184e8429ca6717230d09f9f2c8`

Lenses covered: Conformance, RTL, Robustness, Tests, Docs.

Reproduction evidence: the hosted from-scratch `elaborate` job is green, and `MILAN_LITEX_PYTHON=/home/alex/litex-milan/venv/bin/python3 python3 sw/builder/test_builder.py --require-elaboration` exits 0 at this head. The findings below are proof/failure-path defects that those happy-path runs do not exercise.

[R0] MAJOR Conformance, Tests — `sw/builder/test_builder.py:4672-4825` — gate 23h does not reconstruct the physical stacked tree or test a real missing-patch fixture
Requirement/evidence: `_reconstruct()` keys scratch files by `(tree key, relative path)` and applies each key under a separate `tmp/<key>` root. In the pinned environment, `pythondata_cpu_vexiiriscv/.../Soc.scala` from patch 0005 and `vexiiriscv/src/.../Soc.scala` from the L2 patch resolve to the same physical file, but the gate copies and grades them as two independent trees. That is per-patch grading for the one overlapping source the series must compose on. Its negative control sets `order = series + [break_patch]` and then reverses that patch twice from a correct live tree; it never constructs a live fixture missing the patch and feeds it through the real validator.
Impact: mutually incompatible or incorrectly based changes to the two Vexii patches can evade the claimed whole-series reconstruction, and the 6/6 control proves only that each patch cannot be reverse-applied twice. This repeats the exact “grade a stacked series one patch at a time” failure #185 says gate 23h replaces.
Required change: canonicalize tree roots/target realpaths and apply the complete ordered series to one shared scratch representation of each physical tree. Build an actual installed-tree fixture with each patch absent in turn and run the same production validator against it.
Verification: assert the two current Vexii targets alias, demonstrate both patches compose in one scratch tree from pristine to live, and show six genuinely missing-patch fixtures are rejected for the missing patch rather than for a duplicate reversal.

[R0] MAJOR Robustness — `sw/litex/patches/apply.sh:71-113` — `--reverse` reports success after a best-effort partial reversal
Requirement/evidence: `revert_all()` silently skips every patch whose reverse check fails. The `REV` path then removes `bios/xz`, prints “series reversed,” and exits 0 without proving all six patches were removed. A locally edited patched hunk therefore leaves that patch in place while other patches are removed.
Impact: the documented reverse operation can turn a toolchain into a mixed, partially reversed state and tell the caller it is pristine; automation cannot distinguish that corruption from success.
Required change: preserve best-effort normalization only as an explicitly verified intermediate step. `--reverse` must fail unless the complete series and copied assets are absent (or perform the operation transactionally and roll back on failure).
Verification: alter one patched hunk in a scratch install, run `apply.sh --reverse`, and require a non-zero, path-specific refusal with no false “series reversed” verdict.

[R0] MAJOR Conformance, Tests — `.github/workflows/elaborate.yml:83-112` — the enabled job can elaborate with stale Vexii metadata from another toolchain head
Requirement/evidence: the cache key omits `sw/litex/litex_pins.txt` and `sw/litex/patches/**`; `restore-keys: elaborate-vexii-` may restore any older entry, and `cp -n` installs those old `.py` files. The pinned LiteX wrapper names these files from `interface-metadata-v1 + metadata_args`, not the Vexii source revision, and skips `PythonArgsGen` when the file exists.
Impact: changing the pin or the Scala patches without changing argv can leave CI exercising prior generated metadata. For example, removing the current L2 parser from a refreshed patch can remain green once a cache made by the old parser exists, so the new from-scratch job is no longer from scratch in the part under test.
Required change: cache SBT/Coursier separately. Regenerate the CPU metadata, or restore it only under a strict key that includes all pinned and patched source inputs, with no broad fallback capable of supplying `.py` files across toolchain revisions.
Verification: warm the cache, change a Vexii pin/patch while preserving argv, and prove the next job invokes the changed `PythonArgsGen` or rejects the old metadata.

[R0] MAJOR Robustness, Docs — `.github/workflows/elaborate.yml:20,40-44`; `docs/testing/TESTING.md:786-790`; `scripts/ci_litex_env.py:31-35` — enabling the workflow contradicts both its scheduling claim and the authoritative testing text
Requirement/evidence: the workflow says a documentation PR “should not pay” for the JVM/Scala job, but unconditional `pull_request:` makes every docs-only PR run the roughly five-minute, 230 MB-cache workflow. The inherited testing guide still says CI has no LiteX and the workflow is `workflow_dispatch` only, while `ci_litex_env.py` still calls the repo-carried patch an uncommitted local patch. Current `dev` also has `docs/testing/CI_WORKFLOWS.md` as the authoritative scheduling policy, but the new always-on workflow is absent from it.
Impact: hosted resource policy and cold-reviewer instructions are both materially false at the combined head; a docs-only change now pays the heavyweight setup contrary to the stated policy.
Required change: add conservative change classification/path ownership so irrelevant PRs do not run this job (while all argv/toolchain inputs do), and update `TESTING.md`, `CI_WORKFLOWS.md` after retargeting to current `dev`, and `ci_litex_env.py` to describe the enabled repo-carried series accurately.
Verification: a docs-only PR skips the heavy job with an explicit stable result, a relevant tooling mutation runs it, and all three authoritative descriptions match the actual triggers and patch ownership.

[R0] PASS RTL — `git diff be721ad1...3e21b7b --name-status` plus unchanged 0004/0005 patch contents — this PR changes no repository RTL/CDC/reset/datapath implementation; it packages and verifies an already-carried third-party CPU/toolchain series. Re-review is still required if a fix changes those patch contents or hardware-facing glue.

Mister-M-alt:
[A1] CI evidence, all checks green: `elaborate`, `verilator-suites`, four Verilator shards, `yosys-portability`, `bdd-conformance`, `wire-accountability` and both docs jobs.

**`elaborate` passing is what this stack was for.** It is the CI job that installs the pinned LiteX, places the VexiiRiscv checkout at the revision LiteX names, applies the six-patch series and runs gates 23f/23g/23h. Until now no CI job in this repository had ever elaborated the SoC, which is #154's whole premise. It runs in **1m43s**.

That closes #154's acceptance criterion 1 in the environment it was written about rather than on my bench: severing any argv-to-parameter hop reddens a gate that CI runs.

**One thing to read carefully before comparing to #188.** This PR's `yosys-portability` passes and #188's fails, and the difference is not in either diff. This PR's base is the #154 branch, which predates #176's merge, so it runs the **old** single-job `rtl.yml` whose Yosys job checked out the submodules. #188's base is `dev`, so it runs the new sharded one whose aggregate does not. Both facts are recorded on #190 as a three-way control.

So a green here does **not** mean this branch is clear of that defect - it means this branch has not met it yet. When this PR is retargeted to `dev` after #188 lands, it inherits the same red until #191 is in. Merge order stays **#191, then #188, then this**.

Mister-M-alt (commented):
[R0] NEGATIVE RE-REVIEW — exact head `a8d3f50c38f5ad26ff31d82c798c8b636cd28f89`, base `b34e6cc5524e86d74fb9421b90c81ad09882e747`, candidate merge `6306b4ff91c7379e53eb44a67a4757ca07697e19` (tree-identical to head). All four prior MAJOR findings remain open; this head adds a fifth.

[R0] MAJOR Conformance, Tests — `sw/builder/test_builder.py:4735-4849` — gate 23h still splits one physical Vexii tree and its mutation is not a missing-patch fixture
Requirement/evidence: `_reconstruct()` copies targets under `tmp/<tree-key>` and keys them by `(tree key, relative path)`. The targets of `0005-vexiiriscv-cacheless-litex.patch` and `0002-vexiiriscv-l2-depth-args.patch` resolve to the same physical `Soc.scala`, but the gate grades two scratch files. A synthetic aliased-target series driven through the exact-head production validator produced `targets_alias=True`, `production_validator_problems=[]`, while the same synthetic patches on one shared file returned reverse statuses `0, 1`; that is a demonstrated validator false positive. The negative arm still appends `break_patch` to the complete series and reverses the same patch twice, rather than validating a live fixture genuinely missing it.
Impact: incompatible changes to the two patches can pass the claimed whole-series reconstruction, and the advertised 6/6 mutation tally proves duplicate reversal failure rather than missing-patch detection.
Required change: canonicalize physical targets so the complete ordered series composes on one scratch representation per real file, and run the production validator against a genuine installed-tree fixture with each patch absent in turn.
Verification: assert the two Vexii targets alias; demonstrate pristine -> complete shared series -> live bytes; show six missing-patch fixtures reject for the omitted patch.

[R0] MAJOR Robustness — `sw/litex/patches/apply.sh:71-113` — `--reverse` still reports success after a partial reversal
Requirement/evidence: This artifact is unchanged from reviewed head `3e21b7b`. `revert_all()` silently skips every patch whose reverse check fails; the `--reverse` path then deletes `bios/xz`, prints `series reversed`, and exits zero without proving all six patches are absent.
Impact: one locally edited patched hunk leaves a mixed toolchain while automation is told it is pristine.
Required change: keep best-effort normalization only as a verified intermediate operation; `--reverse` must fail unless the complete series/assets are absent, or roll back transactionally.
Verification: alter one patched hunk in a scratch install and require a nonzero, path-specific refusal with no false success verdict.

[R0] MAJOR Conformance, Tests — `.github/workflows/elaborate.yml:121-154` + hosted job 96762028432 — CI still accepts stale generated Vexii metadata
Requirement/evidence: `~/vexii-netlist-args` remains cached under a key that excludes `sw/litex/litex_pins.txt` and `sw/litex/patches/**`, retains broad `restore-keys: elaborate-vexii-`, and is installed with `cp -n`. The exact current run restored a 219 MB primary-key cache (`elaborate-vexii-39b73e...`) even though this PR changes the patch/toolchain inputs, then reported green without regenerating that metadata.
Impact: a pin or Scala patch can change while CI elaborates metadata made by the prior source, so the job is not from-scratch in the component it exists to prove.
Required change: cache SBT/Coursier independently; regenerate generated CPU metadata, or restore it only under a strict key covering every pin/patch/source input with no broad fallback capable of supplying `.py` files across revisions.
Verification: warm the cache, change a Vexii pin/patch without changing argv, and prove the next run invokes the changed generator or rejects the stale metadata.

[R0] MAJOR Robustness, Docs — `docs/testing/TESTING.md:786-793`; `scripts/ci_litex_env.py:31-35`; `.github/workflows/elaborate.yml:36-44` — authoritative descriptions remain false
Requirement/evidence: These docs are unchanged from `3e21b7b`. `TESTING.md` still says CI has no LiteX, the workflow is dispatch-only, and CI does not run `apply.sh`; `ci_litex_env.py` still calls the repo-carried patch an uncommitted local patch. The workflow now runs on dev/main pushes and every PR and invokes the carried series.
Impact: maintainers and cold reviewers are instructed using an obsolete execution and ownership model.
Required change: update the authoritative testing and environment descriptions now, and update `docs/testing/CI_WORKFLOWS.md` when the required retarget to current `dev` brings that policy into this stack.
Verification: docs checks pass and every trigger, dependency, patch-ownership, and scheduling statement matches the retargeted workflow.

[R0] MAJOR Robustness, Tests — `.github/workflows/elaborate.yml:58-90,164-165` — the new docs-only path fails instead of skipping
Requirement/evidence: The scope output now guards submodules, Python, package install, cache, SBT, source placement, gates, and cache save, but the newly added `Apply the toolchain patch series` step has no matching `if:`. With `rtl=false`, LiteX was never installed and `apply.sh` fails importing `pythondata_cpu_vexiiriscv`; an isolated no-LiteX reproduction exits 1 with `ModuleNotFoundError`. This hosted run cannot cover that arm because `scripts/ci_scope.py` is absent on the stacked base and its log says `elaborating unconditionally`.
Impact: after the planned retarget brings in `ci_scope.py`, a docs-only PR becomes red in the step that was supposed to be skipped.
Required change: make every heavy/dependent step obey the scope decision and add a negative workflow/control proving docs-only succeeds without invoking setup or patch application while relevant toolchain changes still elaborate.
Verification: exercise both classifier outcomes in the merged workflow; docs-only yields a stable successful check with no LiteX work, and a patch/pin mutation runs the complete job.

[R0] PASS RTL — `git diff b34e6cc5524e86d74fb9421b90c81ad09882e747...a8d3f50c38f5ad26ff31d82c798c8b636cd28f89 --name-status` — no repository RTL/CDC/reset/datapath implementation changes; this PR packages third-party sources and verification/tooling.

Coverage at `a8d3f50c38f5ad26ff31d82c798c8b636cd28f89`: Conformance OPEN; RTL PASS; Robustness OPEN; Tests OPEN; Docs OPEN. Exact-head happy paths are green (`test_builder.py --require-elaboration`, `docs_check.py`, `bash -n`, `git diff --check`, and all 11 hosted checks), but none exercises the false-positive/failure paths above. Retargeting from the stacked base to `dev` will require another exact-head/candidate re-review.

Mister-M-alt:
[A1] REVIEW READY

Commit: `32dd8c24670ac34aa62c63ef275c91e5e160d42f`
Base: `154-no-ci-job-elaborates-the-soc-so-argv-to-rtl-parameter-chains-are-unproven` (head descends directly from the unchanged base, so its tree IS the candidate merge tree)

**All five MAJOR findings answered.** I verified each one before fixing it rather than taking it on report, and one of the five I partly disagree with — stated below rather than quietly narrowed.

## F1 — gate 23h split one physical file, and its control was vacuous

**Confirmed, both halves, and this was the important one.** The two Vexii patches resolve to one physical `Soc.scala`:

```
0005-vexiiriscv-cacheless-litex   -> tmp/pythondata_cpu_vexiiriscv/.../ext/VexiiRiscv/src/.../Soc.scala
0002-vexiiriscv-l2-depth-args     -> tmp/vexiiriscv/src/.../Soc.scala
both REAL -> .../ext/VexiiRiscv/src/main/scala/vexiiriscv/soc/litex/Soc.scala
```

Fixed by mirroring the **absolute realpath** instead of `(tree key, relative path)`. The two collide onto one scratch file with no special case, because `realpath(<pythondata root>)/pythondata_cpu_vexiiriscv/verilog/ext/VexiiRiscv` *is* `realpath(<vexiiriscv root>)`. The gate now asserts that any targets sharing a realpath share a scratch path, and names the shared files in its own output:

```
including 2 that two patches share and must therefore compose on
({'boot.c': ['0001-...', '0003-...'], 'Soc.scala': ['0002-vexiiriscv-l2...', '0005-...']})
```

The installed-file count went **8 to 7**, which is the fix visible as a number: `Soc.scala` used to be counted twice as two independent targets.

**The control is rebuilt.** You were right that `order = series + [break_patch]` reverses that patch twice and proves only that it cannot be reversed twice. Each arm now mirrors the installed trees into a fixture that genuinely lacks a patch and feeds that through the **production** validator:

```
[gate 23h mutation] 6/6 fixtures rejected by the production validator, naming a
patch the tree genuinely lacked: 5 with exactly one patch reversed out, 1 where a
later patch sits on the same hunks so the reachable state is the whole tail -
what a half-finished apply.sh leaves. No arm passes by reversing anything twice
```

Five arms are exact single-patch removals. The sixth cannot be: 0003 is diffed on top of 0001, so "0001 absent, 0003 present" is not a state anyone can reach, and the fixture reverses the tail instead — which is precisely the state that opened #185.

**Where I disagree, and I checked it hard.** The impact says a missing patch can evade the reconstruction. I built genuine missing-patch fixtures against a real installed tree before touching the code — omitting `0002-liteeth` (independent) and omitting `0002-vexiiriscv-l2` (the aliased one) — and **both were DETECTED** by the unfixed validator. So the hole was narrower than "a missing patch passes": it was that two patches on one physical file never had to **compose**, with `apply.sh` as a backstop since it applies to the real file. The structural defect and the vacuous control are both exactly as you described; only that one impact sentence overstated it.

(My first attempt at that experiment reported a false positive, because the reversal had failed with `rc=1` and left the tree intact. I caught it on the exit code. Recording it because had I not, I would have published a fabricated defect against my own gate.)

## F2 — `--reverse` reported success after a partial reversal

**Confirmed and fixed**, and it took three attempts, which is worth stating because the first two were wrong in instructive ways:

1. "the patch no longer reverse-applies" — **wrong**: an edited hunk matches neither state, so a refused reverse-check reads as absent. Measured: it passed a tree in which `0002-liteeth` was still applied with one changed line. My own fix failed my own repro of your finding.
2. "every patch forward-applies" — **wrong**: fails on a genuinely pristine tree, because 0003 composes only after 0001.
3. Correct: apply the **complete series** in order, which succeeds only from pristine, then unwind exactly what went on.

Both directions proven:

```
clean --reverse            -> rc=0  "series reversed (verified absent)"
one edited hunk present    -> rc=1  "ERROR: liteeth is not pristine for 0002-liteeth-..."
```

## F3 — CI could elaborate with stale Vexii metadata

**Confirmed and fixed.** The mechanism is real: LiteX names the generated `.py` from `interface-metadata-v1 + metadata_args` and skips `PythonArgsGen` entirely when the file exists, and nothing in that name covers the Scala source.

The caches are now split. The Scala toolchain keeps a broad fallback because coursier is content-addressed. The generated metadata gets a strict key covering `litex_pins.txt` and `sw/litex/patches/**`, and **no `restore-keys` at all** — a prefix fallback is exactly the mechanism that hands over metadata from another toolchain head, so a miss must cost one ~90 s Scala run rather than a silent wrong answer.

## F4 — authoritative descriptions were false

**Confirmed and fixed.** `TESTING.md` said CI has no LiteX and the workflow is dispatch-only; `ci_litex_env.py` still called the repo-carried series an uncommitted local patch. Both now describe the enabled workflow, the three-step install, and repo-carried ownership.

On `CI_WORKFLOWS.md`: it exists only on `dev` (it arrives with #176), so it cannot be edited from this stacked base. `TESTING.md` now says so explicitly and commits to the row landing in the same retarget move.

## F5 — the docs-only path failed instead of skipping

**Confirmed and fixed**, and it was mine: I gated eight steps on the 154 branch, then this branch added a ninth that never got the guard. Only `checkout` and the scope decision are ungated now, which is correct. `scripts/ci_scope.py` is absent on this base, so the workflow logs that and elaborates unconditionally rather than guessing — the skip becomes real at retarget.

## Validation at `32dd8c24`

- `sw/builder/test_builder.py --require-elaboration` — **exit 0**, `ALL GATES PASS EXCEPT 3 NOT RUN`, `every elaboration arm ran`. The three named are two absent build trees and #184's Arty leg.
- gate 23f 10 elaborations; 23f mutation 11/11; 23g 4/5 recipes; 23g mutation 2/2; **23h 6 patches / 7 physical files; 23h mutation 6/6**.
- `docs_check`, `check_doc_paths`, `gen_toc --check`, `check_archive`, `check_feature_status --self-test`, `check_sweep_shape --self-test`, `check_soc_sources`, `pp_srcs --check --selftest`, `gen_module_matrix --check` — all pass.
- `bash -n apply.sh`; `elaborate.yml` parsed with a duplicate-key-rejecting loader (PyYAML accepts duplicate keys silently and GitHub Actions does not — that bit me earlier in this stack).

One self-inflicted note: the first push of this commit accidentally carried three generated `.hex` build outputs the builder gates write into the worktree root. Amended out; the diff is six files.

Re-review requested on this exact head. This round is not a claim that the fixes are right — that is yours to judge.

Mister-M-alt (commented):
[R1] Review of the #188 -> #189 stack at `32dd8c24` (#189 head, whose tree is the combined candidate: it descends directly from #188's `b34e6cc5`), then the fix, pushed to this branch as `95f599ba` (the gate fix) and `c0bf1f2b` (the CI policy row), head `c0bf1f2b4209686cbf2485a05b2c6e5741b780a8` on the maintainer's instruction. Lenses covered: Conformance, Robustness, Tests, Docs; RTL is clean by the artifact named below. Same caveat as on #191: the session that found these also fixed them, and says so.

## [R0]'s five #189 findings at `32dd8c24`: independently re-verified, all hold as answered

| finding | what I ran | result |
|---|---|---|
| F1 gate 23h aliasing / vacuous control | `test_builder.py --require-elaboration` against the bench interpreter | `[gate 23h] ... 7 installed files byte for byte, including 2 that two patches share` (`boot.c`, `Soc.scala`); `[gate 23h mutation] 6/6 fixtures rejected ... 5 with exactly one patch reversed out, 1 where a later patch sits on the same hunks`. The mirror is keyed by realpath (`_mirror`), asserted per target before grading |
| F2 `apply.sh --reverse` false success | on a disposable COPY of the bench toolchain (never the live venv): `--reverse` on a correct tree, forward apply from pristine, apply again, then one edited hunk in `liteeth/phy/gmii.py` and `--reverse` | rc 0 "verified absent"; 6 applied and byte-identical to the bench; second apply a no-op; the edited tree: **rc 1**, `liteeth is not pristine for 0002-liteeth-...`, and the MIXED state is named, not called pristine |
| F3 stale Vexii metadata | `.github/workflows/elaborate.yml:133-154` | two caches; the generated-metadata key hashes `litex_pins.txt` and `sw/litex/patches/**` and has NO `restore-keys`. The warm-cache-then-change-a-patch proof can only run hosted; by construction a changed input is a miss |
| F4 false descriptions | `scripts/ci_litex_env.py:31-37`, `docs/testing/TESTING.md:787-800` | both describe the enabled, repo-carried series. One occurrence survived in `sw/builder/test_builder.py:4078` ("UNCOMMITTED 40-line patch", plus the stale premise that the shipping shape elaborates anywhere but the bench); fixed at `95f599ba` |
| F5 docs-only path fails instead of skipping | `elaborate.yml`: every step after the scope decision carries `if: steps.scope.outputs.rtl == 'true'`, the patch step included | holds |

## [R0]'s #188 findings, which nobody had answered

[R1] MAJOR Tests, Robustness — `sw/builder/test_builder.py` gate 23g (`KNOWN_UNRUNNABLE`, `_is_unpatched_vexii`, the `--require-elaboration` verdict) at `32dd8c24` — still open, exactly as [R0] described it on #188
Requirement/evidence: the recorded Arty row skipped on ANY error; any failure whose text contained `PythonArgsGen` skipped as "#185's" (LiteX surfaces every generator failure as a `CalledProcessError` whose message is the sbt command line, so that substring is in every sbt failure: compile error, missing sbt, or a refreshed patch that broke the parser); and the baseline run printed `3 GATE ARM(S) DID NOT RUN` followed by `--require-elaboration: every elaboration arm ran`. Measured on the bench at `32dd8c24`: the real Arty diagnostic is `AttributeError: 'Record' object has no attribute 'lrclk'`; the real stock-toolchain refusal, provoked by handing the bench's generator `--l2-bogus-option=1`, is `[error] Error: Unknown option --l2-bogus-option=1` on the sbt child's stdout, 17 lines above the end of a stack trace, on a stream the probe's Python-level redirect never sees.
Impact: in the one job that installs the series, a toolchain failure would have been reported as the old harmless reason and the job would have stayed green; the Arty row would have absorbed any new failure on that recipe.
Fixed at `95f599ba`: `KNOWN_UNRUNNABLE` rows carry the exact diagnostic as a regex the `<Exception>: <message>` line must match in full; `_instance_params` keeps the generator child's output tail next to the error; `_classify_recipe` needs all three of an `--l2-*` argument in the argv, a `CalledProcessError` naming `PythonArgsGen`, and scopt's `Unknown option --l2-(down-pending|general-slots)` in that output, or it is a failure; that skip carries a new `TOOLCHAIN` kind, which `--require-elaboration` refuses alongside `NO_LITEX`; the verdict is a function and names what did not run instead of claiming every arm ran. A LiteX-free self-test (`test_recipe_skip_classifier_bites`) grades 8 result shapes and 4 verdict shapes, using the two observed texts verbatim, so the classifier is graded on every box.
Verification (run): [R0]'s mutation, the recorded signature changed so the real exception no longer matches: `gate 23g: sweep.sh arty: RECORDED as unrunnable for \`SoCError: .*\` but failed DIFFERENTLY, which the record does not cover: AttributeError: 'Record' object has no attribute 'lrclk'`; restored, the row is a recorded skip of kind `row`. Full suite on the patched tree: figures unchanged (23f 10, 23f mutation 11/11, 23g 4/5, 23g mutation 2/2, 23h 6/7/2, 23h mutation 6/6), new line `[gate 23g classifier] 8/8 ... 4/4`, verdict `--require-elaboration: ... 3 arm(s) did not run for recorded reasons, listed above, and this verdict does not cover them`, exit 0. Not exercised end to end: an actually unpatched VexiiRiscv under `--require-elaboration` (that needs a cold sbt compile of the stock tree); the refusal text it would print is the one observed and graded.

[R1] BLOCKER Conformance (process) — #188 body / #154 acceptance 1 — [R0]'s merge-order blocker, resolved by landing the stack as ONE candidate
#188 alone puts an acceptance-incomplete `Closes #154` head on `dev` and a dormant `elaborate.yml`. Resolution taken: this branch is rebased onto `dev` (`44a2ff56` plus #191's `fdbfffe8`) and #189 is retargeted to `dev`, so its diff is the whole stack and the reviewed head is the merge candidate; #188 is closed as landed-via-#189 once that merges, with every commit of its branch contained. The rebase was clean (`git merge-tree`: no conflicts, no file touched by both sides).

[R1] MINOR Docs — `docs/testing/CI_WORKFLOWS.md` — the `elaborate` workflow row [A1] committed to adding "in the same retarget move" is added at `c0bf1f2b`.

## Clean lenses

[R1] PASS RTL — `git diff --name-status 9f79dcec...c0bf1f2b` — no `.sv`/`.v`; patch contents of 0004/0005 unchanged from [R0]'s review; the touched surface is LiteX glue, the builder gate, the workflow and docs.

## Evidence at `c0bf1f2b` (run on the identical tree before the final rebase; the rebase onto `9f79dcec` changed only `syn/yosys/run.sh`, which nothing here reads)

| check | result |
|---|---|
| `MILAN_LITEX_PYTHON=<bench venv> python3 sw/builder/test_builder.py --require-elaboration` | exit 0, 47 s. 23f: 10 elaborations; 23f mutation 11/11; 23g 4/5 with the Arty row a recorded skip; 23g mutation 2/2; **23g classifier 8/8 shaped results and 4/4 verdicts**; 23h 6 patches, 7 files, 2 shared; 23h mutation 6/6. Verdict `ALL GATES PASS EXCEPT 3 NOT RUN` (gate 11 and 19b: build trees absent; gate 23g: #184) followed by `--require-elaboration: an interpreter carrying the patch series was found and no elaboration arm was skipped for a toolchain reason; 3 arm(s) did not run for recorded reasons, listed above, and this verdict does not cover them` |
| [R0]'s mutation: recorded Arty signature changed to `SoCError: .*` | gate 23g RED: `sweep.sh arty: RECORDED as unrunnable for \`SoCError: .*\` but failed DIFFERENTLY ... AttributeError: 'Record' object has no attribute 'lrclk'`; restored: recorded skip, kind `row` |
| `apply.sh` on a disposable copy of the bench toolchain (PYTHONPATH ahead of the editable installs; the live venv untouched) | `--reverse` on a correct tree rc 0 "verified absent"; forward from pristine: 6 applied, byte-identical to the bench (`diff -rq`); second apply a no-op; one edited hunk in `gmii.py` then `--reverse`: rc 1, MIXED state named |
| docs and shape gates | `docs_check` 0 findings / 213 files; `check_doc_paths` 1108 resolve; `gen_toc --check` and `--verify-anchors` OK; `check_archive`, `check_feature_status --self-test`, `check_sweep_shape --self-test`, `check_soc_sources`, `pp_srcs --check --selftest`, `gen_module_matrix --check`, `ci_scope --selftest`: all pass; `bash -n apply.sh` |
| `elaborate.yml` parsed with a duplicate-key-rejecting loader | 12 steps; only `checkout` and the scope decision ungated |
| `git diff --check`, em dashes on added lines | clean; none outside the TOC generator's own separator |
| Yosys sweep | `dev`'s own: 46/46 at `fdbfffe8` (PR #191), structural gates PASS; this branch's RTL and `run.sh` are identical to `dev`'s |
| Verilator sweep (`scripts/run_all_suites.sh`, tsn-gen at `f52f1f9`) | running on this tree now; tally follows in a comment |

Hosted: the retarget starts the exhaustive run and the `elaborate` workflow on `c0bf1f2b`; the latter now runs with `scripts/ci_scope.py` present for the first time.

Merge: not by me. The maintainer authorised #191; this stack is fixed and ready and waits for that call, and for [R0]'s round on this head if one is wanted.

Mister-M-alt:
[R1] Local Verilator sweep at `c0bf1f2b4209686cbf2485a05b2c6e5741b780a8`, the exact head of this branch: `scripts/run_all_suites.sh` in a clean clone with the three RTL submodules initialised, Verilator 5.050, tsn-gen oracle at `f52f1f9` with `TSN_GEN_ROOT` set so the fuzz campaigns ran rather than declared-skipping.

```
suites: 52   passed: 52   failed: 0   timed out: 0
checks: 2118264   in-suite failures: 0
real 15m30s   rc=0
git status --short afterwards:
 M hdl/ieee1722/avtp/doc/TEST_RESULTS.md
 M hdl/ieee8021as/gptp_plane/doc/TEST_RESULTS.md
```

Correction to the first version of this comment, which said the tree was clean: it was not. The two modified files are the tracked `TEST_RESULTS.md` pages the tsn_fuzz campaigns rewrite with the run's own figures; the local oracle is tsn-gen `f52f1f9` and CI pins `fde65206`, so the rewritten lines differ from the tracked ones. Not a finding against this PR (it changes no RTL and no test), but a reason nobody should commit a local sweep's by-products without reading them. Reproduced by re-running `tb/verilator/tsn_fuzz` alone on the same head (27 s, `all campaigns passed`); the whole diff is the two timestamp lines, the figures are identical to the tracked ones:

```
-**PASS** — 164 pass, 0 fail, 0 known gaps  ·  2026-08-19 21:09
+**PASS** — 164 pass, 0 fail, 0 known gaps  ·  2026-08-21 17:59
-**PASS** — 353 pass, 0 fail, 9 known gaps  ·  2026-08-20 09:06
+**PASS** — 353 pass, 0 fail, 9 known gaps  ·  2026-08-21 17:59
```

Recorded rather than hidden: a first attempt on the box's tmpfs ended `passed: 40 failed: 12` with **0 in-suite failures**; every one of the 12 was the C++ compiler dying on `Disk quota exceeded` (`/tmp` is mounted `usrquota`, and this user's ceiling was reached while `df` still showed 7 GB free). That log is environmental evidence and is not quoted; the run above, under `$HOME`, is the record.

Yosys: `dev`'s own sweep applies, 46/46 with the structural gates at `fdbfffe8` (PR #191), because this head's RTL and `syn/yosys/run.sh` are byte-identical to `dev`'s; the hosted shards on this head are running.

Mister-M-alt:
[R1] The close/reopen just above is deliberate and changes nothing on the branch: the force-push of `c0bf1f2b` happened while this PR's base was still the #154 branch, against which a head rebased onto `dev` conflicts (five paths, `elaborate.yml` and `test_builder.py` among them), and GitHub fires no `pull_request` workflow on a conflicted PR; the retarget to `dev` is an `edited` event none of the workflows listens to. Reopening fired `rtl`, `rtl-fast`, `docs` and `elaborate` on this head against `dev` (runs 32500763776, 32500763552, 32500763833, 32500763623).

Mister-M-alt:
[R1] Merged into `dev` at `2823480203955e1f7993e71dd2ab2335e0f74cc7` on the maintainer's explicit authorization, with: hosted at `c0bf1f2b` 18 of 19 check-runs green (`elaborate`, `rtl-fast`, `docs` and every finished shard), 1 still running at merge time; local Verilator 52/52 (2,118,264 checks) and the builder suite with the bench interpreter on this tree; `dev`'s Yosys 46/46 applying to identical RTL. [R0]'s round on this head did not happen before the merge; that is recorded here rather than implied otherwise. #188 closes as landed-via-#189; #154, #185 and #156 close once the `dev` run at the merge commit is green.

Issue #150, opened 2026-08-20
Deleting a model pin silently removes its check from the campaign


Mister-M-alt:
Sibling of #146. That issue closes the half where `grade_tx()` met a constraint kind it did not recognise and graded nothing. This is the other half, and it is currently open with nobody owning it.

`grade_tx()` skips any field carrying no constraint at all:

```python
for name, _bits, con in model.fields:
    if name in skip or name not in got or not con:
        continue
```

That skip is correct as far as it goes. Plenty of fields are legitimately unconstrained (`sequence_id`, the clock identities), and failing them would redden the campaign for no reason. #146 deliberately stopped its fail-closed rule at non-empty constraints for exactly this reason.

The problem is what happens when a pin is DELETED. Removing an `expected:` block from a tsn-gen model silently removes its check from the campaign: no failure, no skip line, and the only visible trace is the total moving. That is not theoretical. Measured by controlled experiment at the CI-pinned tsn-gen rev, the gPTP campaign went from **355 to 353** when two pins were dropped:

| lost check | model change | why the pin went |
|---|---|---|
| `tx_pdreq.correction_field` | `8021as_pdelay_req.yaml` dropped `value: 0` | 802.1AS-2011 Table 11-5 gives Pdelay_Req a correctionField role, so pinning zero would fail a conformant requester |
| `tx_announce.steps_removed` | `8021as_announce.yaml` dropped `range: [0, 254]` | 10.3.10.2.1(b) is a receiver qualification rule; 10.5.3.2.6 bounds nothing on transmit |

**Both of those unpins were right.** That is what makes this hard: the mechanism that legitimately relaxes a wrong pin is indistinguishable, from the campaign's point of view, from one that quietly deletes coverage. Zero failures is not evidence of coverage. It is evidence that nothing failed among whatever is still being graded.

The same shape has now cost us twice, in different disguises: a `mask:` pin that asserted nothing because the grader had no branch for it (#146), and this. In both cases the tally was the only witness and nobody was watching it.

## What a fix has to do

Make a deliberate unpin cheap and an accidental one loud. It must NOT simply fail unconstrained fields, which would be unusable. Some options, none obviously best:

- Record the per-field graded set as a committed artifact and diff it, so a lost check shows up as a named line rather than a smaller number. This is the same idea as #148's freshness gate, one level finer.
- Require an unconstrained field to be listed explicitly, so silence becomes a declaration rather than a default.
[...]

PR #194, opened 2026-08-21
Parse the hdl tree with Vivado's xvlog front-end behind an identity ratchet

### LINKED ISSUE #132: No gate sees Vivado front-end strictness: a module Verilator accepts can be rejected by xvlog
Found while answering the review of PR #128.

## The gap

`KL_gptp_shadow.sv` used a signal in an `always_comb` above the line that declared it. Verilator accepts that. Vivado's front-end does not:

```
ERROR: [VRFC 10-3380] identifier 'full_w' is used before its declaration
ERROR: [VRFC 10-8530] module 'KL_gptp_shadow' is ignored due to previous errors
```

The full local bar was green throughout: the 52-suite Verilator sweep (2.1M checks), `lint_rtl.py --check`, and `syn/yosys/run.sh` all passed. `synth_design` only warns (Synth 8-6901) and still builds, so even a bitstream run would not have failed. `dev` parsed clean, so this was a clean regression that nothing in the repo could see — and it makes the file ineligible for the xsim benches under `tb/utests/` and `tb/itests/`.

## Why it is worth a gate

The lint ratchet, the sweep and yosys all share one front-end family (Verilator/sv2v). Any construct where Vivado is STRICTER than Verilator is invisible to the whole bar. Use-before-declaration is one; there will be others.

## Proposal

A local, optional parse gate — seconds to run, no project, no synthesis:

```sh
/home/alex/Xilinx/2026.1/Vivado/bin/xvlog -sv <every hdl/**/*.sv>
```

Wire it as a `scripts/` helper that SKIPS cleanly when no Vivado is present (the tsn-gen precedent: report the skip, never a false green), so it is runnable on a bench box and inert in CI. Ratchet it the way the lint gate is ratcheted if today's tree has pre-existing findings.

## Acceptance
- A deliberate use-before-declaration in any `hdl/` module makes the gate RED (mutation-proven).
- The gate skips cleanly, exit 0, with a visible marker on a box without Vivado.
- Today's tree passes (or its findings are ratcheted and printed in full).

Mister-M-alt:
[A1]

## Status

GREEN locally — the gate runs, finds 3 pre-existing findings and ratchets them, `--check` passes, `--selftest` passes with and without Vivado, and the whole doc/script bar is green — `132-xvlog-front-end-gate` -> `dev`. No RTL changed.

## Linked Issue / roles

Closes #132. Files #193 for the defects it grandfathers.

Executor: `[A1]`

## Description

The lint ratchet, the 52-suite Verilator sweep and the Yosys portability gate all lower SystemVerilog through Verilator or sv2v, so any construct where Vivado's front-end is **stricter** is invisible to the whole bar. `KL_gptp_shadow.sv` proved it (PR #128): a use-before-declaration that Verilator accepted, every local gate passed, and Vivado rejected with `VRFC 10-3380`. `synth_design` only warns, so even a bitstream builds.

| piece | what |
|---|---|
| `scripts/xvlog_gate.py` | runs `xvlog -sv` over every tracked `hdl/` module. Packages first (detected by reading each file for a `package` decl, like `pp_srcs.py`, never by name), then each module **independently** — xvlog stops a compilation unit at the first module error, so a shared compile would mask every file after the first fault. |
| ratchet | keyed on finding **identity** `path\|CODE\|identifier`, not a count, so a compensating swap (one fixed, one added in the same commit) cannot net to zero — the lesson of #150. `--check` fails on a NEW finding **and** on a grandfathered one that has silently vanished (bank it deliberately). |
| `scripts/xvlog.budget` | today's 3 findings, grandfathered and printed in full; the defects themselves are #193. |
| skip | no xvlog on the box → a visible `SKIPPED` marker and exit 0, never a false green (the tsn-gen precedent). Reads only `hdl/` and include dirs under it, so an uninitialised submodule cannot make the tree look cleaner (#186). |
| `--selftest` | pure-Python arms (parser, dedup, ratchet diff) that run on any box, plus a live arm that plants a use-before-declaration into a clean donor and requires the gate to redden and name the identifier. Wired into `scripts/run_all_suites.sh` next to the other self-tested gates. |

## Authoritative references

- #132 (the gap and the proposed gate), the PR #128 review that found it
- `scripts/lint_rtl.py` + `scripts/lint.budget` (the ratchet pattern), `scripts/pp_srcs.py` (package detection, tracked-only, skip-not-false-green)
- #150 (identity over count), #186 (a tool-absence must not under-report), #193 (the defects grandfathered here)

```sh
git fetch origin && git checkout 132-xvlog-front-end-gate
git submodule update --init third_party/verilog-axis protocol-processor gptp-processor  # not needed by the gate; for the rest of the bar
export XVLOG=/home/alex/Xilinx/2026.1/Vivado/bin/xvlog   # a bench box; CI has none and the gate skips
```

## How to validate

```sh
python3 scripts/xvlog_gate.py --selftest    # PASS (skips the live arm without xvlog)
python3 scripts/xvlog_gate.py --check       # PASS: 3 findings == ratchet
python3 scripts/xvlog_gate.py               # prints the 3 findings, rewrites the budget
```

Expected: `--check` exits 0 at the ratchet; a planted use-before-declaration in any clean module makes `--check` exit 1 and name the identifier; on a box without Vivado all modes print a SKIP marker and exit 0.

- The 3 grandfathered findings are real defects; **fixing** them (declaration moves in `milan_csr.sv` and `milan_datapath.sv`) is #193, not this lane — the scope of #132 is the gate.
- The gate is analyse-only (`xvlog`), not elaborate/synthesise; it catches front-end strictness, not timing or resource issues.
- It is not wired as a blocking CI job: CI has no Vivado, so it would only ever skip there. Its `--selftest` is the CI-visible part.

## Definition of Done

- [x] Linked Issue acceptance: a planted use-before-declaration reddens the gate (mutation-proven); it skips cleanly with a marker and exit 0 without Vivado; today's findings are ratcheted and printed in full.
- [x] New behaviour has self-checking tests (`--selftest`, pure + live arms, wired into `run_all_suites.sh` and proven to abort it when broken).
- [x] Required local bar passes (docs_check, check_doc_paths, gen_toc --check/--verify-anchors, pp_srcs --check --selftest, gen_module_matrix --check, lint_rtl --check/--pragmas). No RTL, so the Verilator/Yosys sweeps are unaffected.
- [ ] Internal cleared-context review positive
- [ ] External review positive
- [ ] Post-merge containment checked before the Issue moves to Done


Mister-M-alt (commented):
[R0] NEGATIVE — exact head `e77f1b9baa24e3dade896995fb7b97bd0340a56b` against `dev@2823480203955e1f7993e71dd2ab2335e0f74cc7`. Request changes; the hosted green does not exercise the two false-green paths below.

[R0] MAJOR Robustness, Tests — `scripts/xvlog_gate.py:421-443` — normal mode can raise the identity ratchet

Requirement/evidence: #132 explicitly adopts the `lint_rtl.py` ratchet precedent, whose normal mode only lowers an existing budget and refuses regressions. Here `--check` computes `new/gone`, but normal mode ignores the existing budget and unconditionally calls `write_budget(findings)`. An exact-head pure probe with one grandfathered key plus one new key returned 0 and rewrote the budget to contain both:
```
probe rc=0 wrote=[
  hdl/a.sv|VRFC 10-3380|old,
  hdl/b.sv|VRFC 10-3380|new
]
```
Impact: running the documented default command after introducing an xvlog regression silently converts that regression into accepted baseline. The next `--check` is green, so the ratchet can be raised by the tool that claims only to lower it.

Required change: when a budget exists, refuse and leave it byte-unchanged if any current key is new; rewrite only to remove vanished keys. Initial creation when no budget exists can remain explicit. Add a self-test that proves a new key cannot be grandfathered by normal mode.

[R0] MAJOR Robustness, Tests — `scripts/xvlog_gate.py:218-229` — an unparsed nonzero xvlog exit disappears

Requirement/evidence: both package and module loops treat a nonzero subprocess as only the findings returned by `_parse`. A tool/license/crash diagnostic that does not match the VRFC error regex therefore becomes an empty list. Exact-head probe:
```
returncode=1, output="FATAL: xvlog aborted"
findings=[] package_errors=[]
```
Impact: a previously clean module can fail to parse while `--check` reports no new key; once the ratchet is empty this can make a completely failed invocation print PASS. In default mode it can also erase the budget and exit 0.

Required change: every nonzero invocation must produce a hard setup/tool failure even when no recognized diagnostic can be parsed. Preserve parsed findings for detail, but never let returncode != 0 collapse to “clean.” Add package and module self-test arms with unmatched fatal output.

[R0] MINOR Docs, Tests — `scripts/xvlog_gate.py:246` / `scripts/xvlog.budget:10` — the committed generated artifact is not generated by this head

The generator says the outstanding fixes belong to #193; the committed `GENERATED` budget says #133, an unrelated merged tsn-gen ticket. Running the generator dirties its own purportedly current artifact, while `--check` strips comments and cannot detect the mismatch. Regenerate the budget from this head and keep the owning issue accurate.

Lens result: Conformance/Robustness/Tests remain open on the two MAJOR escapes; Docs has the generated-header mismatch; RTL is unchanged and clean by scope. The parser/dedup/diff unit arms and current hosted checks pass, but they do not cover either failing subprocess semantics or the actual default-mode ratchet transition.

PR #195, opened 2026-08-21
check_results_fresh: cover the fail column in the freshness self-test

### LINKED ISSUE #152: check_results_fresh: no self-test arm covers a fail-column-only disagreement
Follow-up from the #149 review, raised by the reviewer with an exact fixture and explicitly scoped as not blocking that merge.

`section_sums()` in `scripts/check_results_fresh.py` compares a generated artifact's headline tally against the sum of its own `## Sections` rows. The check itself is correct. Its self-test does not fully hold it.

A second-order mutation battery of 14 mutants against the 17-arm self-test left four survivors. Three are non-defects and the reviewer said why: two make the gate *stricter* (empty table treated as `(0,0,0)`, case-insensitive heading match), and the `**FAIL**`-only stamp strip is unreachable because a failing campaign exits at the recipe's `rc -ne 0` guard before the gate runs.

The fourth is real:

> **`compare pass+gaps, DROP the fail column` survives.**

No arm carries a fail-column-only disagreement. So a mutation that stops comparing the fail column entirely leaves the self-test fully green, and **fail is the column with the most consequence** - it is the one that reads PASS at the top of an artifact while being red in the middle.

This is the literal twin of the defect that round of review had just closed. The pass column had exactly this gap; it was fixed by adding an `only the gap column disagrees` arm; the fail column was left with the same hole. That is worth recording as more than a missing test: when a check compares N things and the self-test covers N-1 of them, the uncovered one is not random, it is whichever the author was not thinking about at the time.

## Fix

Add the arm the reviewer specified - a fixture whose committed and fresh copies agree on pass and gaps and disagree only on fail - and confirm it kills the `drop the fail column` mutant and nothing else.

## Acceptance

- [ ] A fail-column-only disagreement is refused, and the message names the fail column.
- [ ] The new arm kills the `drop the fail column` mutant and no other, so it is not standing in for an existing arm.
- [ ] The full second-order battery is re-run and the survivor count for genuine defects is zero.
- [ ] The three known non-defect survivors stay survivors, and the reasons are recorded rather than re-derived next time.

Mister-M-alt:
[A1]

## Status

GREEN — `check_results_fresh.py --self-test` is 18/18, and the new arm is proven to kill the target mutant and no existing arm does — `152-fresh-fail-column` -> `dev`. One file, self-test only.

## Linked Issue / roles

Closes #152.

Executor: `[A1]`

## Description

`scripts/check_results_fresh.py`'s self-test covered the pass and gap columns of the `ss != at` coherence check (headline vs section-row sums) but not the **fail** column. So a mutant that compares pass+gaps and drops fail left the whole self-test green — the one genuine survivor of the #149-review second-order battery, and the literal twin of the defect that round had closed on the pass column in the same edit.

Fail is the column with the most consequence: it reads `PASS` at the top of an artifact while being red in the middle.

The fix is one arm, mirroring `only the gap column disagrees`: both the fresh and committed copies carry a section-row fail that the headline does not report (rows sum to 1 fail, headline says 0), so a pass-only comparison reaches a clean `OK` over a file that disagrees with itself about whether anything failed. The honest gate refuses it `UNVERIFIABLE` and the message names the fail column.

## Authoritative references

- #152 (the reviewer's fixture and scope), the #149 review that raised it
- `scripts/check_results_fresh.py` `verdict()` / `section_sums()` and the existing `only the gap column disagrees` arm it twins

```sh
git fetch origin && git checkout 152-fresh-fail-column
```

## How to validate

```sh
python3 scripts/check_results_fresh.py --self-test    # 18 checks: 18 PASS, 0 FAIL
```

Expected: 18/18. Under a `drop the fail column from ss != at` mutant the self-test goes RED (the new arm catches it); under the three known non-defect mutants (empty table -> (0,0,0), case-insensitive heading, the unreachable `**FAIL**` stamp strip) it stays green, so the arm is not standing in for an existing one.

- One column of one comparison. It does not touch the live campaign path or any other gate.

## Definition of Done

- [x] A fail-column-only disagreement is refused and the message names the fail column.
- [x] The new arm kills the `drop the fail column` mutant and no other (measured: no existing arm is flipped by that mutant; the three non-defect survivors stay survivors).
- [x] Self-test green (18/18); docs gates green.
- [ ] Internal cleared-context review positive
- [ ] External review positive
- [ ] Post-merge containment checked before the Issue moves to Done

### DIFF
--- a/scripts/check_results_fresh.py
+++ b/scripts/check_results_fresh.py
@@ -402,6 +402,20 @@ SELFTEST = [
                      "| everything else | 330 | 0 | 8 |"),
      art(T1).replace("| everything else | 330 | 0 | 9 |",
                      "| everything else | 330 | 0 | 8 |"), UNVERIFIABLE),
+    # ...and fail is the third column, the one with the most consequence: it
+    # reads PASS at the top of an artifact while being red in the middle. The
+    # pass and gap columns each have an arm above; without this one, a mutant
+    # that compares pass+gaps and DROPS fail from the `ss != at` coherence
+    # check leaves the whole self-test green (measured: it is the one survivor
+    # of the second-order battery in #152). Both sides carry the same fail
+    # incoherence - the section rows sum to one fail the headline does not
+    # report - so a pass-only comparison reaches a clean OK over a file that
+    # disagrees with itself about whether anything failed.
+    ("only the fail column disagrees", RAN,
+     art(T1).replace("| everything else | 330 | 0 | 9 |",
+                     "| everything else | 330 | 1 | 9 |"),
+     art(T1).replace("| everything else | 330 | 0 | 9 |",
+                     "| everything else | 330 | 1 | 9 |"), UNVERIFIABLE),
     # THE SCOPING IS THE PROPERTY, so it gets arms of its own. Rows under a
     # different heading are not section rows: drop the heading scoping, or
     # match the heading by prefix instead of exactly, and the row below is


Mister-M-alt (commented):
[R0] POSITIVE — exact head `b8dcbf6e630f62ae4f9876c7ce7898f2cabe0d81` against `dev@2823480203955e1f7993e71dd2ab2335e0f74cc7`. No findings.

This is the narrow #152 closure it claims: one additional self-test fixture makes the fail column independently load-bearing in the artifact/section coherence check.

Evidence:
- Exact-head `scripts/check_results_fresh.py --self-test`: 18/18 PASS.
- I replayed the target mutant that compares only the pass and gap columns in `ss != at`. The suite became 17/18; the sole red arm was `only the fail column disagrees`. Restoring the fail-column comparison returns 18/18.
- The diff is +14 lines in the self-test only; production parsing/verdict behavior, RTL, generated artifacts, and interfaces are unchanged.
- All hosted checks on this exact head are green.

Lens result: Tests and Robustness PASS with a non-vacuous mutant; Conformance PASS against #152 acceptance; RTL unchanged; Docs unchanged apart from explanatory test comments.

PR #196, opened 2026-08-21
tsn_fuzz: commit and diff the per-field graded set so a lost or swapped pin is named

### LINKED ISSUE #150: Deleting a model pin silently removes its check from the campaign

Mister-M-alt:
[A1]

## Status

GREEN locally — both campaigns regenerate and pass their freshness gate at the CI-pinned tsn-gen rev, the gPTP tally rises to 354, and dropping OR swapping a model pin is now caught by NAME — `150-graded-set-artifact` -> `dev`.

## Linked Issue / roles

Closes #150.

Executor: `[A1]`

## Description

Per the recorded decision on #150 (option 1). `check_results_fresh.py` diffs each committed `TEST_RESULTS.md` against a fresh run, but the file recorded only per-section COUNTS, gaps and failures — never the passing graded labels. So the two residuals the narrowing named survived: a lost pin was reported by section not field, and a compensating drop-plus-add inside one section netted to zero and passed the text compare silently.

| piece | what |
|---|---|
| `tb/verilator/tsn_fuzz/cosim.py` | `Report` now records every asserted check and emits a `## Graded checks` section — the per-field graded set, by section, sorted and de-duplicated so it is reproducible. |
| effect | the existing whole-file freshness diff now catches ANY graded-set change and names the field. No new gate, and `check_results_fresh.py` is untouched (so no conflict with the open #152/#195). |
| `tb/verilator/tsn_fuzz/fuzz_ptp.py` | asserts our Pdelay_Req carries `correction_field == 0` — a fact about THIS device, which the shared tsn-gen model rightly leaves unconstrained (Table 11-5 lets a conformant requester send nonzero). |
| artifacts | `hdl/ieee1722/avtp/doc` and `hdl/ieee8021as/gptp_plane/doc` `TEST_RESULTS.md` regenerated at tsn-gen `fde65206` (the CI pin), each carrying the new section. |

**One measured correction to the ticket, the tree wins**: #150 asked for TWO device assertions and a tally rise of two. `steps_removed == 0` on Announce ALREADY landed with the fuzz_ptp tier (commit 7ccca036, `fuzz_ptp.py:454`), so only `correction_field == 0` was owed and the gPTP tally rises by exactly one, 353 -> 354. The other unpin (`steps_removed`) is already restored as a device assertion; this closes the second.

## Authoritative references

- #150 and its recorded decision (option 1: a diffable per-field artifact closes the compensating-swap residual that a count cannot see), #146/#149 (the sibling half and the freshness gate)
- 802.1AS Table 11-5 (Pdelay_Req correctionField role) — why the model unpins it and the device assertion belongs in the campaign
- `scripts/check_results_fresh.py` (the whole-file diff this rides), the memory rule to measure at the CI tsn-gen rev not the local branch

```sh
git fetch origin && git checkout 150-graded-set-artifact
git submodule update --init third_party/verilog-axis gptp-processor
# tsn-gen at the CI pin, built:
git clone https://github.com/kebag-logic/tsn-gen && git -C tsn-gen checkout fde652067d3618f47fba2c2e8eb05f6c9f6fa97f
git -C tsn-gen submodule update --init --recursive external/rapidyaml
cmake -S tsn-gen -B tsn-gen/build -DCMAKE_BUILD_TYPE=Release -DENABLE_PARSER_TESTS=OFF && cmake --build tsn-gen/build --target packet_gen -j
export TSN_GEN_ROOT=$PWD/tsn-gen
```

## How to validate

```sh
cd tb/verilator/tsn_fuzz && make aaf && make ptp     # both green: campaign + freshness
```

Expected: gPTP 354 pass / 0 fail / 9 gaps, AAF 164/0/0, each `TEST_RESULTS.md` carrying a `## Graded checks` section, freshness `1 PASS`.

- `hdl/ieee1722/aaf/doc/TEST_RESULTS.md` is orphaned (its ACMP/ADP/AECP campaigns were deleted with those planes); no campaign writes it, so it gets no graded-checks section here. Out of scope — a stale-orphan cleanup is its own ticket.
- The graded set is every asserted check, not only grade_tx pins; that is a strict superset and the intended ratchet (a deliberate coverage change regenerates the artifact).

## Definition of Done

- [x] Deleting an `expected:` block produces a RED that names the field (mutation-proven: dropping the Sync `reserved0` pin drops 354->353 and the diff prints `-* tx_sync.reserved0`).
- [x] A compensating swap inside one section is caught (count-neutral proof: 369 == 369 bullets, STALE, "counts agree; difference elsewhere", `-* tx_sync.reserved0 / +* tx_sync.zzz_swapped_field`).
- [x] Legitimately unconstrained fields do not redden anything (grade_tx still skips them; behaviour unchanged).
- [x] The two spec-justified unpins stay unpinned in the model; the device facts they carried are asserted in the campaign (steps_removed pre-existing, correction_field added).
- [x] Local bar green: matrix --check, docs_check, check_doc_paths, test_grade_tx 22/22, both campaigns + freshness.
- [ ] Internal + external review positive; post-merge containment.


Mister-M-alt (commented):
[R0] POSITIVE — exact head `e209da2633be7c88312c303c5e46a39d36a8a21d` against `dev@2823480203955e1f7993e71dd2ab2335e0f74cc7`. No findings.

The implementation closes the count-neutral residual rather than merely adding another tally:
- `Report.ck` records the stable `(section, label)` identity of every asserted check, and the writer groups, sorts, and de-duplicates labels while preserving deterministic section order.
- The committed AAF artifact has 164 graded labels, all unique, matching 164 passes. The gPTP artifact has 354 graded labels, all unique, matching 354 passes; its nine known gaps remain separately named.
- I drove the exact-head writer through the existing freshness verifier with a count-neutral swap. Both runs were 2/0/0, but `pin.beta` -> `pin.gamma` returned STALE and named both lines:
```
-* pin.beta
+* pin.gamma
```
- Both changed Python files compile. The new Pdelay_Req correction-field assertion is made on the decoded frame from this device, adjacent to the existing first-request checks, while the shared standards model remains legitimately unconstrained.
- The campaign-bearing hosted `verilator-suites` job and all completed fast/docs/elaboration shards are green. At publication recheck only the optional Yosys 0/4 shard was still running; no RTL changes are in this PR.

Lens result: Conformance, Robustness, Tests, and Docs PASS; RTL unchanged. The named graded-set artifact catches both a lost label and a compensating swap without making run-varying detail part of the committed identity.

Issue #217, opened 2026-08-22
gPTP campaign: the truncation oracle never asserts that an over-minimum frame is accepted


Mister-M-alt:
[A1]

## Objective

Make the gPTP campaign's length oracle two-sided for every parsed message type:
one byte below the legal minimum is refused, the minimum is accepted, and a
legal suffix TLV above the minimum is accepted and acted on.

## Current state (reconciled after PR #218, 2026-08-23)

The original mutation narrowed acceptance to exact length. It was invisible for
Pdelay_Req, Sync, and Follow_Up and caught Announce only through unrelated
fixtures. PR #218 strengthened six Announce qualification probes with named
"parser drop counter unmoved" assertions; at `dev` `457d140c`, narrowing the
Announce parser arm now makes those six checks fail. Announce therefore has real
no-drop coverage, although its purpose is still incidental rather than an
explicit suffix-acceptance case.

No equivalent explicit coverage exists yet for Pdelay_Req, Pdelay_Resp,
Pdelay_Resp_Follow_Up, Sync, or Follow_Up.

## Work

- Add one deliberately above-minimum, well-formed suffix-TLV probe for each of
  the six parsed message types.
- Assert downstream behavior, not only parser admission: requests are answered,
  response pairs affect only their matching exchange, Sync/Follow_Up reaches
  steering, and Announce is qualified/adopted as appropriate.
- Name the Announce probe for length acceptance so coverage no longer depends on
  a path-trace fixture remaining long.
- For each type, narrow only its acceptance arm back to exact length and show its
  dedicated probe turns red; record the per-type result and regenerated tally.

## Acceptance

- [ ] Pdelay_Req, Pdelay_Resp, Pdelay_Resp_Follow_Up, Sync, Follow_Up, and
      Announce each have an explicit legal-suffix acceptance check.
- [ ] The request/paired types prove the expected response or state transition,
      so parser admission cannot pass vacuously.
- [ ] Each exact-length mutation fails for its own named check.
- [ ] The README describes the two-sided oracle and the generated campaign
      artifact is fresh.

Related: #142 / PR #215, #136 / PR #218, #141 / PR #221, #219.

Issue #214, opened 2026-08-22
gPTP stamper: the banner promises a msgType tag the ports do not carry


Mister-M-alt:
[A1]

## What is wrong

`hdl/ieee8021as/gptp_plane/KL_gptp_txstamp.sv:15` states that the stamper "reports `{ts, sequenceId, msgType}` once the sequence lanes pass -- the engine matches by its pending exchange". Its ports carry no `msgType`:

```
:53    output logic        ts_valid_o,
:54    output logic [63:0] ts_ns_o,
:55    output logic [15:0] ts_seq_o
```

So the only tag crossing to the engine is the 16-bit sequenceId. The banner overstates the interface, and the beat map at `:66` tracks beats 0 through 5 to reach the sequenceId at wire bytes 44 and 45, while messageType is wire byte 14, the low nibble, which is beat 1 lane 6 and is passed on the way.

## Why it matters now, beyond a wrong comment

The engine submodule is being fixed so that a peer's Pdelay_Req arriving while our own Pdelay_Req is outstanding can no longer steal our egress timestamp (Mister-M-alt/FPGA-gPTP#28). The fix routes each returning stamp to its own transmitter by matching the tag the hardware already delivers, which is what this banner says the engine is supposed to do. With sequenceId alone the match has a residual ambiguity: our Pdelay_Req carries our own sequence counter, our Pdelay_Resp echoes the peer's request sequence, the two counters are independent, and when both frames are outstanding at once and the two 16-bit values happen to be equal, the tag cannot separate them. Exporting messageType closes that case completely, because the two frames are always of different types.

The ambiguity is narrow (it needs both frames outstanding and a 1 in 65,536 coincidence) and the engine-side fix will narrow it further with a pending-set check, so this is not urgent. It is filed because the cheapest complete fix lives on this side of the boundary, and because a banner that advertises a field the interface does not carry will mislead the next reader of either repository.

## What to do

- Either export the messageType nibble alongside the sequenceId, letting the engine match on both, or correct the banner to describe the interface as it is. Prefer the export: the byte is already on a beat the module walks, and it closes a real ambiguity rather than documenting it.
- If exported, coordinate the submodule side so the engine consumes it, and keep the pending-set check so the tag is a confirmation rather than the sole authority.
[...]

Issue #142, opened 2026-08-20
Reject truncated Pdelay_Req frames without sending a response


Mister-M-alt:
[A0]

## What is wrong

The integrated receive path accepts a header-only, 34-byte Pdelay_Req PTPDU and emits Pdelay_Resp plus Pdelay_Resp_Follow_Up. A valid request is 54 bytes and includes two reserved ten-byte fields. Malformed or physically truncated input must not make the device originate a response pair.

The current parent truncation campaign encodes a 48-byte Ethernet frame as the minimum, so it does not test the legal 68-byte boundary.

## One-to-one mapping

This issue maps only to [FPGA-gPTP #12](https://github.com/Mister-M-alt/FPGA-gPTP/issues/12). The donor issue owns enforcing the complete Pdelay_Req shape. This parent issue owns correcting the integration oracle and updating the submodule pin.

Related parent work: #110, #117, PR #123, and its R0 review finding.

## Work in this repository

- Update `gptp-processor` to the reviewed donor fix.
- Change the parent legal minimum from a 48-byte Ethernet frame to a 68-byte untagged Ethernet frame.
- Test both `messageLength` inconsistency and physical truncation at the 53-byte PTPDU boundary.
- Prove malformed input increments the receive-drop counter, emits no response frame, and does not poison the next valid exchange.

## Acceptance criteria

- [ ] FPGA-gPTP #12 is closed by a reviewed donor change.
- [ ] The parent submodule pin includes that change.
- [ ] Header-only and physically truncated Pdelay_Req frames are dropped before responder dispatch.
- [ ] No Pdelay response pair is emitted for malformed input, and a following complete request succeeds.
- [ ] Restoring the old 34-byte minimum makes the parent campaign fail.
- [ ] `make -C gptp-processor` and `make -C tb/verilator/tsn_fuzz ptp` pass.
- [ ] The complete local verification bar required by `CONTRIBUTING.md` passes before merge.

Issue #141, opened 2026-08-20
Ignore unmatched Pdelay response follow-ups without dropping asCapable


Mister-M-alt:
[A0]

## What is wrong

An unsolicited or stale Pdelay_Resp_Follow_Up with this device's requesting identity but no matching outstanding sequence can update `neighborPropDelay`. One measured frame changed the published delay from 600 ns to about 20 ms and dropped `asCapable`, so stale traffic can interrupt synchronization.

## One-to-one mapping

This issue maps only to [FPGA-gPTP #8](https://github.com/Mister-M-alt/FPGA-gPTP/issues/8). The donor issue owns matching the follow-up to an outstanding Pdelay exchange. This parent issue owns the fixed submodule pin and integration proof.

Related parent work: #110, #117, and PR #123.

## Work in this repository

- Update `gptp-processor` to the reviewed donor fix.
- Replace the FPGA-gPTP #8 tracked gap with a hard parent assertion.
- Prove that an unmatched or stale follow-up changes neither peer delay nor `asCapable` and does not consume a valid outstanding exchange.
- Send a valid exchange after the malformed frame to prove recovery is not merely a frozen responder.

## Acceptance criteria

- [ ] FPGA-gPTP #8 is closed by a reviewed donor change.
- [ ] The parent submodule pin includes that change.
- [ ] Unmatched and stale Pdelay follow-ups are ignored without changing peer delay or `asCapable`.
- [ ] A following valid exchange still completes normally.
- [ ] Restoring sequence-only or identity-only matching makes the parent campaign fail.
- [ ] `make -C gptp-processor` and `make -C tb/verilator/tsn_fuzz ptp` pass.
- [ ] The complete local verification bar required by `CONTRIBUTING.md` passes before merge.

PR #218, opened 2026-08-23
Reject Announce loops, self-sources and excessive steps before GM selection: retire the three FPGA-gPTP#7 allowances

### LINKED ISSUE #136: Reject Announce loops, self-sources, and excessive steps before GM selection
[A0]

## What is wrong

The integrated gPTP plane can adopt Announce messages that must be rejected before best-master selection:

- a message claiming to come from this device's own clock identity;
- a message with `stepsRemoved >= 255`;
- a path trace containing this device's clock identity, which represents a loop.

A better priority value in any of those malformed messages can currently replace the selected grandmaster.

## One-to-one mapping

This issue maps only to [FPGA-gPTP #7](https://github.com/Mister-M-alt/FPGA-gPTP/issues/7). The donor issue owns Announce qualification. This parent issue owns the fixed submodule pin and the integration-level no-regression checks.

Related parent work: #110, #117, and PR #123.

## Work in this repository

- Update `gptp-processor` to the reviewed donor fix.
- Remove the FPGA-gPTP #7 allowances from the parent gPTP field campaign.
- Keep all three rejection vectors independently graded.
- Verify that each rejected Announce leaves grandmaster and parent identity unchanged and increments the expected receive-drop accounting.

## Acceptance criteria

- [ ] FPGA-gPTP #7 is closed by a reviewed donor change.
- [ ] The parent submodule pin includes that change.
- [ ] Own-source, `stepsRemoved >= 255`, and path-trace-loop Announce frames are hard failures if adopted.
- [ ] A mutation that disables any one of the three qualifications makes the parent campaign fail.
- [ ] `make -C gptp-processor` and `make -C tb/verilator/tsn_fuzz ptp` pass.
- [ ] The complete local verification bar required by `CONTRIBUTING.md` passes before merge.

Mister-M-alt:
[A1]

## Contents

- **[Status](#status)** -- Green/WIP/blocked, test tally, and `branch` -> `dev`.
- **[Linked Issue / roles](#linked-issue--roles)** -- Public task, executor, and independent reviewers.
- **[Description](#description)** -- What changed and why.
- **[Beyond the ticket](#beyond-the-ticket)** -- Everything in this diff the Issue text does not ask for, the campaign accounting, and a note on #217.
- **[Authoritative references](#authoritative-references)** -- Requirements/specification clauses and docs.
- **[How to get into the same state](#how-to-get-into-the-same-state)** -- Copy-pasteable checkout/dependency/environment commands.
- **[How to validate](#how-to-validate)** -- Exact reviewer commands and expected result.
- **[Known limitations / out of scope](#known-limitations--out-of-scope)** -- What this deliberately does not do, and why.
- **[Definition of Done](#definition-of-done)** -- The merge bar from CONTRIBUTING.md.

## Status

GREEN. `gptp-processor` does not move: the pin `dev` already carries (`c33fb1af`) contains this ticket's donor fix, Mister-M-alt/FPGA-gPTP#20.

`136-qualify-announce` -> `dev` at exact head `ec42e529c693d2e5b96d1e6444a34a0b2e88b5a2` (one commit on `origin/dev` `5ee9167ac234101e206d6a544beb8688622751d2`, an ancestor, so the head tree is the candidate merge tree).

gPTP field campaign, tsn-gen at the CI pin `fde65206`: **547 pass, 0 fail, 0 known gaps** (`dev` measured on this tree: 520 pass, 0 fail, 0 known gaps).

## Linked Issue / roles

Closes #136
Relates to #217 -- the truncation oracle's one-sided acceptance, which this PR touches indirectly; see the note at the end of "Beyond the ticket".

**No overlap with PR #216 (#214), measured rather than assumed.** `git diff --name-only origin/dev` for each branch: this PR touches `tb/verilator/tsn_fuzz/fuzz_ptp.py`, its README, `docs/traceability/ieee8021as.md` and the generated artifact; #216 touches `KL_gptp_shadow.sv`, `KL_gptp_txstamp.sv`, `milan_datapath.sv`, `gptp_shadow_wrap.sv`, that bench's `sim_main.cpp` and README, and `docs/design/GPTP_PLANE.md`. The intersection of the two file lists is empty. CORRECTED after the `[R1]` round: a disjoint file list does NOT make the campaign independent of #216, because `tb/verilator/tsn_fuzz/Makefile` compiles `gptp_shadow_wrap.sv`, `KL_gptp_shadow.sv` and `KL_gptp_txstamp.sv`, all three changed by #216. Independence was therefore MEASURED on the candidate merge tree (`ec42e529` merged onto `710350d7`, tree `b0787790`, which is byte-identical to what `dev` became at the merge `457d140c`): campaign 547 pass, 0 fail, 0 known gaps on the head-only tree and on the candidate, both by the reviewer.

Donor: Mister-M-alt/FPGA-gPTP#7 (CLOSED), fixed by Mister-M-alt/FPGA-gPTP#20 (reviewed `[R1]` POSITIVE at `e177b68f`, MERGED as `48f299ab`), already in the pin `dev` carries.

Executor: `[A1]`
Internal cleared-context reviewer: pending
External reviewer: pending

## Description

The FPGA-gPTP #7 allowances are THREE probes reached through TWO code paths, and all three are gone. Which call site each retirement covers, since `reject_probe`'s body carries the arm and the call sites carry the vectors:

| retired vector | where the allowance lived on `dev` | which call site | now |
|---|---|---|---|
| own sourcePortIdentity (10.3.10.2.1a) | the `eq_or_gap(..., issue)` INSIDE `reject_probe` (`fuzz_ptp.py:806` on `dev`), taken because the call passed `issue=7` | the call at `dev:823` | the same call with `issue=None`, so it takes the hard `rep.ck` path plus the `stable()` canary |
| stepsRemoved 255 (10.3.10.2.1b) | the same `eq_or_gap` inside `reject_probe` | the call at `dev:826` | the same, `issue=None` |
| our identity in the path trace (10.3.10.2.1c) | its own direct `eq_or_gap(..., 7)` at `dev:834`, hand-rolled outside `reject_probe` and comparing only the published grandmaster | not a `reject_probe` call at all on `dev` | folded INTO `reject_probe` with `issue=None`, so it gains the canary its two siblings already had |

So two of the three retirements are the same line of code reached from two different call sites, and the third is a separate line; the review of #142 asked for exactly this distinction and it applies here more than there.

The three sibling probes that pass `None` and `drop=1` (version 1, transportSpecific 0, domain 5) are parser-owned and untouched, and `12-hop legal better vector still adopts (cap publishes 8)` still passes: the qualification must refuse loops without refusing depth.

Added, per the Issue's `[R1]` integration note, because the donor's own suite is not where the parent's independent vectors live:

- `stepsRemoved 0x0100`: the field is 16 bits (10.3.10.2.1b bounds the value, not its low byte).
- `path_trace=[OUR_CID, PEER2_CID]`: our identity as the FIRST hop. This is the only loop an end station meets without forgery -- our own Announce returned through a bridge -- and it is the hop a walk that starts at index 1 misses.
- `path_trace=[OUR_CID]`: our identity as the only hop.

## Beyond the ticket

Two changes the Issue text does not ask for, both strengthenings, both stated here so no reviewer has to find them:

**1. `drop=None` now ASSERTS the parser drop counter, where it used to skip it.** `reject_probe(label, issue, drop, ...)` treated `drop=None` as "do not check the counter". The Issue's integration note calls `drop=None` "the accounting" for a qualification refusal, and it was not: a refusal that wrongly counted a header drop would have passed. It now emits `<label>: parser drop counter unmoved` on every `drop=None` probe. That is +6 checks (the six qualification probes) and it is the only assertion a refusal has, because the donor keeps no qualification counter: a well-formed Announce is handed over by the parser and refused by 10.3.10.2.1 before best-master selection, so "nothing moved, including the counter" is the whole observable.

**2. The path-trace probe moved into `reject_probe`.** It was a bare `eq_or_gap` comparing the published grandmaster only; as a `reject_probe` it also asserts `GM unmoved`, `parent unmoved` and `flags stable` (+3 checks; its `parser drop counter unmoved` is already counted in item 1's +6). Same vector, three more things it can catch. (Corrected after the `[R1]` round: the first wording listed the drop counter here and omitted `GM unmoved`.)

Everything else in the +27 is the Issue's own: 3 retired arms x +3 checks each (hard form plus canary), 3 new probes x 4 checks.

### The campaign delta, from the check list rather than the counts

520 -> 547, diffed line by line against `dev`'s committed `TEST_RESULTS.md` (`grep '^\* '`, sorted, `comm`): **3 lines removed, 30 added**. The three removed are the `(conformant)` markers the retired allowances printed while they still went through `eq_or_gap` (`better vector, own sourcePortIdentity (10.3.10.2.1a): rejected, plane stays GM (conformant)` and its two siblings). The thirty added are the six qualification probes at five checks each: the verdict, the three-part `stable()` canary, and the new `parser drop counter unmoved`. 520 - 3 + 30 = 547, and no check was lost in between. One label did change: `... (10.3.10.2.1c): rejected (conformant)` became `...: rejected, plane stays GM`, which is the retired marker's line being re-emitted by the hard path, counted above as one of the 3 removed and one of the 30 added.

### A note on #217, since this PR makes one of its columns look better than it is

#217 records that the truncation oracle asserts short-is-dropped and at-minimum-is-accepted but never longer-is-accepted, and that Announce is caught today only by the accident that `12-hop legal better vector still adopts (cap publishes 8)` sends a 164-octet frame and requires it to be adopted. (Sizes in this note are PTP octets, i.e. after the 14-byte Ethernet header: the no-TLV minimum is 64, the default one-hop fixture 76, two hops 84, twelve hops 164; on the wire that is 78 / 90 / 98 / 178.)

CORRECTED after the `[R1]` round, which measured this note and found it wrong in both directions:

- The enumeration was wrong. `path_trace=[0x3333, OUR_CID]` already existed on `dev`; the fixtures this PR adds are `[OUR_CID, PEER2_CID]` (84), `[OUR_CID]` (76) and the stepsRemoved `0x0100` probe, which is above minimum because the default `path_trace=[gm_identity]` emits a one-hop TLV (76). Every one of them is incidental in the same way as the 12-hop vector: each exists to prove a 10.3.10.2.1 refusal, and each would stop covering length acceptance the moment someone narrowed it to the minimum shape for an unrelated reason.
- The note understated a real improvement. With the parser's Announce acceptance narrowed to `cnt_r == min_end_w` (#217's own mutation), `dev` at `710350d7` gives 511 pass, 9 fail, none of the nine naming length (GM-adoption and servo smears); this PR's candidate gives 532 pass, 15 fail: the same 9 plus SIX named `<probe>: parser drop counter unmoved`, one per qualification probe. The `drop=None` strengthening of "Beyond the ticket" item 1 is what turns those six frames into genuine "an above-minimum Announce is not dropped by the parser" assertions, which is exactly #217's missing oracle for this message type. Three of the six are structurally immune to fixture narrowing, since a 10.3.10.2.1c probe cannot exist at the 64-octet no-TLV minimum.

So after this PR the Announce row of #217 is materially better covered, not merely better-looking; the other three message types are unchanged, and the explicit longer-is-accepted probe per type stays with #217. The 9-against-15 measurement is recorded on #217.

## Authoritative references

- IEEE Std 802.1AS-2011 10.3.10.2.1 (qualifyAnnounce): (a) an Announce whose sourcePortIdentity.clockIdentity is thisClock is not qualified, (b) stepsRemoved >= 255 is not qualified, (c) a path trace containing thisClock is not qualified; 10.5.3.3 (the path trace TLV); IEEE 1588-2008 9.3.2.5 a) and d) (the underlying rules 802.1AS profiles).
- Milan v1.2 section 4.2.6 profiles IEEE Std 802.1AS-2011 with Cor1-2013 and Cor2-2015 (the normative edition recorded on #139).
- Mister-M-alt/FPGA-gPTP#7 and its fix PR #20; issue #136 and its `[R1]` integration note (including its wording fix: "increments the expected receive-drop accounting" reads "drop counter unmoved", which item 1 above now actually asserts).

## How to validate

Same recipe as #210 (`make -C gptp-processor`, the campaign, the sweep, Yosys, lint, xvlog, the doc gates, and `sw/builder/test_builder.py` twice, NOT in parallel). Expected, all measured at this head: campaign **547 pass, 0 fail, 0 known gaps**, freshness `1 PASS`, rc 0, and no line naming FPGA-gPTP #7. The rest of the bar at this head:

- `make -C gptp-processor` `768` / `167` / `323` checks, all PASS, lint clean, rc 0; submodule clean afterwards, and the pin is byte-identical to `dev`'s.
- sweep `suites: 52 passed: 52 failed: 0 timed out: 0`, rc 0, 596 s. The absolute count depends on the base: `checks: 2118504` at the reviewed base `5ee9167a` (its 2,118,477 plus this PR's 27); on the candidate merge tree with `710350d7` the `[R1]` review measured 2,118,523 against that base's 2,118,496, the same +27. The first figure was stale by the time of the merge; the delta is the claim.
- Yosys `tops: 46 pass: 46 fail: 0`, `RESULT: PASS`, `TIED-INPUT GATE: PASS (2 justified tie(s), no unjustified ones)`, `TAP-PURITY RESULT: PASS`, rc 0, 972 s. `xvlog gate: PASS (3 finding(s) == ratchet)`, 82 s. `lint_rtl.py --check` `PASS (99 <= ratchet 99, 22 waived)`. `sw/builder/test_builder.py` `ALL GATES PASS EXCEPT 2 NOT RUN` plain (267 s) and with `--require-elaboration` (267 s), rc 0 both.
- `docs_check` 0 findings across 213, `check_doc_paths` OK (1131), `gen_toc --check` OK (152), `gen_module_matrix --check` up to date, `test_grade_tx` 54/54, `git diff --check origin/dev...HEAD` clean, working tree clean.

Those three RTL-facing gates cannot be moved by this diff and were run anyway: this PR changes `tb/verilator/tsn_fuzz/fuzz_ptp.py`, two documents and the generated artifact, while `syn/yosys/run.sh` reads `hdl/`, `syn/`, the submodules and exactly two `tb/` files (`tb/verilator/classifier/classifier_wrap.sv` and `tb/verilator/queues/queues_wrap.sv`), and `scripts/xvlog_gate.py` and `scripts/lint_rtl.py` walk `hdl/` only.

### Isolation: which arm each probe exercises, and the proof it does

`qualifyAnnounce` is three arms in the donor microcode (`gptp-processor/hdl/ucode/gen_gptp_ucode.py`, `prog_rx_announce`), and every probe here names the arm it drives. Each mutation below deletes or narrows ONE arm; the campaign is re-run and the failures listed are ALL of them, so each probe is shown to be sensitive to its own arm and to nothing else. All measured at pin `c33fb1af`, restored after each, campaign back to `547 pass, 0 fail, 0 known gaps`.

One caveat on the pass column, per the rule #215 added to `tb/verilator/tsn_fuzz/README.md`: these figures are "campaign total minus what failed", so ANY check landing in the campaign ahead of this PR re-prices every pass number below, anywhere in the file, while only a check joining the qualifyAnnounce group moves the fail side. The fail counts and the named checks are the durable half; if this PR rebases behind another that adds probes, I re-measure the pass column rather than carrying it over.

| mutation | arm | campaign | every check that went red |
|---|---|---|---|
| Q1: the own-source branch deleted (`BRS_Z -> out` after the identity compare) | (a) sourcePortIdentity is thisClock | `545 pass, 2 fail` | `own sourcePortIdentity (10.3.10.2.1a)`: its verdict and `GM unmoved` |
| Q2: `STEPS_MAX_C` 255 -> 256, the bound off by one | (b) stepsRemoved >= 255 | `544 pass, 3 fail` | `stepsRemoved 255 (10.3.10.2.1b)`: verdict, `GM unmoved`, `parent unmoved` |
| Q2b: that compare narrowed to a byte (`FMT_W` -> `FMT_B`) | (b), its WIDTH | `544 pass, 3 fail` | `stepsRemoved 0x0100 (10.3.10.2.1b)` only -- 0x0100's low byte is zero, so the 255 vector cannot see this fault and does not fire |
| Q3: the path-trace hop compare's branch deleted | (c) thisClock in the path trace | `538 pass, 9 fail` | all three (c) probes (second hop, FIRST hop, only hop), three checks each, and nothing outside them |
| M3: the hop walk started at index 1 instead of 0 | (c), its BASE WORD | `541 pass, 6 fail` | the FIRST-hop and only-hop probes only; the second-hop probe still passes, which is why the integration note asked for them |

Two things that table is meant to settle, and both are about a single arm needing more than one vector. The (b) arm needs both vectors, because narrowing the compare to a byte is invisible to the 255 vector; and the (c) arm needs the first-hop vector, because a walk that skips hop 0 still catches a second-hop loop. Q3 turning all three (c) probes red is isolation at ARM granularity, not a smear: they share one arm by construction, and M3 splits them.

The `parser drop counter unmoved` checks stay green under every mutation above, which is the point of the wording fix on the ticket: a qualifyAnnounce refusal is not a header drop, so the counter holding is the accounting, and a mutation that wrongly ADOPTS a vector does not move it either.

- The 8-hop publication cap is unchanged: hops beyond the bank's eight are invisible to the walk. CORRECTED after the `[R1]` round: a loop whose first occurrence of our identity lies beyond the eighth hop is NOT refused by 10.3.10.2.1 c), and b) catches it only once `stepsRemoved` reaches 255 (the donor header's own wording), so at realistic hop counts such a frame is adopted; the review measured our identity at index 7 of a 12-entry trace refused and at index 8 and 9 adopted with `steps_removed=11`. This PR's earlier claim that the stepsRemoved bound catches it was wrong. Tracked as #219, which also owns the cap-edge probe the campaign lacks. Exposure is low for this device's role: a single-port end station transmits Announce only while grandmaster, so an un-forged returned Announce carries our identity at index 0, and 802.1AS-2011 B.1 bounds the performance envelope at seven hops.
- The FPGA-gPTP #8 and #10 allowances are untouched (#141 and #137 own them). No FPGA-gPTP #22 gap remains in the campaign at this pin (PR #210 retired the nine when the donor fix landed in `c33fb1af`); the first wording of this line was stale.
- The (b) bound and both 64-bit identity compares are unpinned by the PARENT campaign (`STEPS_MAX_C` moved to 128, or either compare narrowed to 32 bits, leaves it at 547/0); the donor suite catches all three at 320/3 each and the bar runs it, so the composite bar is sound. Stated so the isolation table above is read as parent-side coverage, not as the whole proof.

## Definition of Done

- [ ] Linked Issue acceptance criteria are satisfied
- [ ] New or changed behavior has self-checking tests
- [ ] Required local verification bar passes
- [ ] Self-test evidence is posted in a PR comment
- [ ] No undocumented requirement or interface change remains
- [ ] Candidate merge result is validated per CONTRIBUTING.md
- [ ] Documentation is updated where needed


Mister-M-alt:
[R0] Correction for the record, from the #219 lane: this PR's Known-limitations wording said a loop deeper than eight hops is caught by the stepsRemoved bound. Measurably false at realistic hop counts - clause b) fires only at 255. The behaviour is instead owned by the donor's complete-walk lineage (FPGA-gPTP #7/#45: every declared hop compared as the bytes stream), present at the current pin, and the cap edge is now pinned by three directed campaign probes plus a design-page statement (see the #219 closing PR). No tracked file repeats the false wording.

Issue #246, opened 2026-08-24
syn/ooc/milan_datapath_ooc.tcl never generates ucode.hex: it reports 30,925 LUT for a design that is 38,848, and exits 0


Mister-M-alt:
[A1]

Third and worst member of the family #235 / #245 belong to. Found while re-measuring PR #227's
area at its new head; not a regression, and #227 does not touch this file.

- **#235** (fixed by PR #240): `syn/ooc/pp_shadow_ooc.tcl` never reads its own `-top`. Fails
  loudly, non-zero. No wrong number can come out of it.
- **#245**: `syn/yosys/ooc.sh` never generates `ucode.hex` and swallows the failure, exiting 0.
  Prints an error, produces no number.
- **This one**: `syn/ooc/milan_datapath_ooc.tcl` never generates `ucode.hex` either, but Vivado
  only raises a `CRITICAL WARNING`, infers an all-zero microcode ROM, constant-folds the AECP
  uCPU behind it, and **completes with `rc=0` and a full, plausible, wrong utilization report**.

## Measured

Vivado 2026.1, `xc7a100tfgg484-2`, the checked-in recipe, at `69-notify-reland` `6be50377`
(the file is byte-identical to `origin/dev` `04b55dad`). The only difference between the two runs
is whether `ucode.hex` exists in the launch directory.

| run | `milan_datapath` LUT | FF | device LUT |
|---|---|---|---|
| the recipe exactly as checked in | **30,925** | 35,543 | 48.78% |
| same recipe, `ucode.hex` generated into the run dir first | **38,848** | 42,438 | 61.27% |

**The checked-in recipe under-reports the shipping design by 7,923 LUT -- 12.49 device points --
and says nothing that stops a reader quoting it.** The only trace is one line among 800:

```
CRITICAL WARNING: [Synth 8-4445] could not open $readmem data file 'ucode.hex'
```

The script already solves this exact problem for the *other* ROM: it generates `ltn_rom.hex`
into the run directory and passes the absolute path as `PP_TROM_HEX_P`, with a comment saying
why ("cannot quietly measure an all-zero ROM"). `ucode.hex` -- read by `KL_aecp_ucpu` /
`KL_aecp_engine` through `UCODE_HEX_P`, generated by
`protocol-processor/hdl/aecp/ucode/gen_ucode.py` -- got no such treatment.

## Why it matters

This script is the instrument the area campaign runs on. #229's epic, #231 and #232 all turn on
[...]

PR #240, opened 2026-08-24
syn/ooc: derive the pp_shadow OOC read set from run.sh so it contains its own top

Mister-M-alt:
[A1]

## Contents

- **[Status](#status)**
- **[Linked Issue / roles](#linked-issue--roles)**
- **[Description](#description)**
- **[Authoritative references](#authoritative-references)**
- **[How to get into the same state](#how-to-get-into-the-same-state)**
- **[How to validate](#how-to-validate)**
- **[Known limitations / out of scope](#known-limitations--out-of-scope)**
- **[Definition of Done](#definition-of-done)**

## Status

GREEN at head `64d2fae8`, answering [R0] round three's blocker on `d4e004bd`.

**Every figure in this body was measured at `64d2fae8`.** Nothing is carried over from
`d4e004bd`, `cf991e51`, `add9f717` or `bafd29f3`, and no figure from an earlier round
survives: `dev` moved under the review (PR #227 merged at `cb444a09`, carrying the AS_PATH
trigger split and re-pinning `protocol-processor` to `3770ae02`), this branch was rebased
onto it, and the read set and both area figures changed as a result. The two Vivado runs
were taken at `6ebbd17f`, the parent of `64d2fae8`; `64d2fae8` edits only the page they are
recorded on and touches nothing either run reads.

`235-ooc-reads-its-top` -> `dev`.

## Linked Issue / roles

Closes #235
Relates to #231

Executor: `[A1]`
Internal cleared-context reviewer: `[R<n>]`
External reviewer: `[R<n>]`

## Description

The script named `KL_pp_shadow` as `-top` and read 41 sources that did not include it.
Three review rounds have been spent on the two halves of that: where the read set comes
from, and how "the top is in it" is decided. The first half is settled -- `syn/yosys/run.sh`
emits its own read set and `syn/ooc/dp_srcs.py` consumes the emission. **This round changes
the second half**, because the same finding came back three times.

| Round | The construct that escaped | What was changed |
|---|---|---|
| one | a block comment around `module KL_pp_shadow` | strip comments and strings first |
| two | `` `define UNUSED `` / `module KL_pp_shadow;` / `endmodule`, joined by backslashes | consume a `define` and everything its continuations carry |
| three | `` `define DISCARD(x) `` then `` `DISCARD( `` / the declaration / `)` | **the model is deleted** |

Round three's construct is legal input, and at `d4e004bd` `declares()` returned `True` and
`build(..., front_end=(None, "no front end"))` returned `rc=0` with no findings, while the
[...]

Issue #235, opened 2026-08-24
syn/ooc/pp_shadow_ooc.tcl never reads the module it synthesizes: the documented plane-area recipe cannot run


Mister-M-alt:
[A1]

Found while producing the base-versus-head area evidence for PR #227 (#69). Not a regression: it fails identically on `origin/dev`.

## What happens

`syn/ooc/pp_shadow_ooc.tcl` never reads the module it names as `-top`. It reads the derived `protocol-processor` source list (`scripts/pp_srcs.py`) and `third_party/verilog-axis/rtl/axis_fifo.v`, then runs `synth_design -top KL_pp_shadow`, but `hdl/milan/KL_pp_shadow.sv` is in neither list, so synthesis has no such module.

Reproduced verbatim at `origin/dev` `04b55dad` (submodules at that head's pins), Vivado 2026.1:

```
$ python3 <repo>/protocol-processor/hdl/acmp/rom/gen_ltn_rom.py -o ltn_rom.hex
$ python3 <repo>/protocol-processor/hdl/aecp/ucode/gen_ucode.py -o ucode.hex
$ PP_N_IN=1 PP_N_OUT=1 vivado -mode batch -source <repo>/syn/ooc/pp_shadow_ooc.tcl -nojournal -log ooc.log
pp_shadow OOC shape: N_STREAM_IN_P=1 N_STREAM_OUT_P=1
ERROR: [Synth 8-439] module 'KL_pp_shadow' not found
5 Infos, 40 Warnings, 0 Critical Warnings and 2 Errors encountered.
ERROR: [Common 17-69] Command failed: Vivado Synthesis failed
rc=1
```

It fails loudly rather than silently, which is the good half: no wrong number can come out of it. The bad half is that no number comes out of it at all.

## Why it matters now

This script is the documented reproduce recipe for the plane's area figures. `docs/findings/PP_SHADOW_AREA_0812.md` names it four times, including its Reproduce section, and #229's epic cites that page as the historical baseline it wants replaced with a current one. #231 asks for authoritative 1x1 and 8x8 `KL_pp_shadow` baselines: whoever picks it up will reach for this script first and it cannot run.

`syn/ooc/milan_datapath_ooc.tcl` is unaffected. It builds its list from `syn/ooc/dp_srcs.py`, which includes `hdl/`, so the datapath flow works today (measured in PR #227).

## Fix

[...]

PR #191, opened 2026-08-21
Print the Yosys inventory without requiring the submodule it says it does not need

Mister-M-alt:
[A1]

Closes #190. **`dev` is red; this is the fix.**

## Status

Ready for review, and urgent: `yosys-portability` has failed on every run since #176 merged, so **no PR based on current `dev` can show a green long gate** until this lands.

**Scoped after [R0]'s round 1**, which was right that the original wording overreached. PR #189 is green on that check, because its base is the #154 branch, which predates #176 and therefore runs the OLD single-job `rtl.yml` whose Yosys job did check out submodules. That is not an exception to the defect, it is a PR that has not met it yet: retargeting #189 to `dev` inherits the same red. The three-way control is on #190.

**Roles**: executor `[A1]`; reviewer `[R0]` (round 1 delivered, NEGATIVE, three MINOR, all answered at `fc65610d`).

**Reproducible checkout**: `git fetch origin && git checkout fc65610dd71692f8a9c3001b2c94ea581e5829ed && git submodule update --init protocol-processor gptp-processor third_party/verilog-axis`

## The defect

The aggregate job fails in 7 seconds, with all four Yosys shards passing:

```
protocol-processor/hdl not found. The protocol-processor submodule is not checked out...
##[error]Process completed with exit code 2.
```

Two things that only bite together:

1. `.github/workflows/rtl.yml`'s `yosys-portability` job checks out the repository and no submodules, because it only wants top **names**.
2. `syn/yosys/run.sh` derived the submodule source list at line 89 and did not reach the `--list` exit until line 194, so `--list` needed a checked-out `protocol-processor` while its own header said twice that it needed nothing:
   - `# Requires yosys and sv2v except for --list.`
   - `--list                 print selected top names without requiring tools`

Both sentences were true before the source list became derived. **That derivation is mine**, from #175: `--list` silently inherited a requirement its documentation denies, and #176 is simply the first caller to invoke it from a job without submodules.

## The fix

Derive the source list only when it is going to be used. The name column of `tops` is the first field of each row and no source path contributes to it, so `--list` never needed this.

The placeholder on the `--list` path is deliberately a path that cannot exist. `--list` exits long before any top is built, so it is never read; if a later change moves that exit, yosys fails on a missing file rather than synthesising an **empty source list**, which is the exact failure `pp_srcs.py` exists to refuse. The alternative fix, adding a submodule checkout to the aggregate job, would have kept a `--list` that lies about its requirements and paid a submodule fetch to print 46 strings.

[...]

PR #262, opened 2026-08-25
syn/yosys: ooc.sh generates both control-plane ROMs and exits non-zero on any failed top

### LINKED ISSUE #245: syn/yosys/ooc.sh prints a synthesis failure and exits 0, and never generates ucode.hex
[A1]

Found while answering the [R0] review on PR #227 (#69); the reviewer found it independently
at the same head. Not a regression: it fails identically on `origin/dev`, and PR #227 does not
touch `syn/yosys/ooc.sh`. Filed separately rather than widened into #227.

Sibling of #235 / PR #240, which is the same class in the **Vivado** OOC recipe
(`syn/ooc/*.tcl`). This one is the **Yosys** recipe, a different file that #240 does not touch,
and it is worse in one respect: #235 fails loudly with a non-zero status, so no wrong number can
come out of it. This one prints a failure and returns success.

## What happens

`syn/yosys/ooc.sh` has two independent defects that compose into a false green.

**1. It generates one of the two ROMs the control plane `$readmemh`s.** It generates
`ltn_rom.hex` into its temp dir, but `KL_aecp_ucpu` / `KL_aecp_engine` also read
`UCODE_HEX_P = "ucode.hex"` by the same relative-name contract, and nothing generates it. Both
the `KL_pp_shadow` and the `milan_datapath` tops carry the control plane, so both are affected.

**2. A failed top does not fail the script.** The `sv2v FAIL` and `yosys FAIL` branches print and
`continue`. Nothing accumulates a status, so the exit code is whatever the last statement of the
file leaves behind (`[ -n "${OOC_TMP:-}" ] || rm -rf "$TMP"`) -- always 0, however many tops died.

## Reproduced verbatim

At `69-notify-reland` `6be50377`, whose `syn/yosys/ooc.sh` is byte-identical to `origin/dev`
`04b55dad`. Yosys 0.66, sv2v v0.0.13:

```
$ cd syn/yosys && ./ooc.sh KL_pp_shadow; echo "EXIT=$?"
== OOC area (synth_xilinx -family xc7 -flatten) ==
top                               LUT       FF   RAMB36   RAMB18      DSP   CARRY4
KL_pp_shadow                 yosys FAIL: ERROR: Can not open file `ucode.hex` for \$readmemh.
EXIT=0
```

Supplying `ucode.hex` in the run directory by hand is what makes it synthesize, and that is
exactly the untracked recipe an area figure must never come from.

## Why it matters

`docs/design/AREA_BUDGET.md` cites this script twice as the toolchain behind its measured
figures, and the area campaign (#229, with #231 / #232) is live: a decision-relevant number can
be produced from a manually patched run, or a caller can conclude the whole list passed when
half of it never synthesized. `ooc.sh` is not in any CI workflow, so nothing else catches it.

## Fix

- Generate `ucode.hex` beside `ltn_rom.hex` from
  `protocol-processor/hdl/aecp/ucode/gen_ucode.py`, into the same `$TMP` the yosys run uses,
  and treat a generator failure as fatal instead of `|| true`.
- Accumulate a failure status across the loop and exit non-zero if any requested top failed
  sv2v or yosys.
- While there: the `ltn_rom.hex` generator's status is already discarded by `|| true`, which is
  the same class as the missing `ucode.hex` -- an absent ROM should stop the run, not shape a
  number.

## Acceptance

- [ ] `syn/yosys/ooc.sh KL_pp_shadow` from a fresh clone synthesizes and prints a real row.
- [ ] `syn/yosys/ooc.sh milan_datapath` likewise.
- [ ] Deleting or corrupting either generated ROM, or naming a top that cannot elaborate, makes
      the script exit non-zero -- demonstrated, not asserted.
- [ ] A run leaves nothing in the caller's directory (the #191 / #192 rule this script already
      honours for `ltn_rom.hex`).

Mister-M-alt:
[A3]

## Contents

- **[Status](#status)**
- **[Linked Issue / roles](#linked-issue--roles)**
- **[Description](#description)**
- **[How to get into the same state](#how-to-get-into-the-same-state)**
- **[How to validate](#how-to-validate)**
- **[Known limitations / out of scope](#known-limitations--out-of-scope)**
- **[Definition of Done](#definition-of-done)**

## Status

GREEN at head `862ed9f0`, answering [R0]'s round-five findings on `61e8213f` (which answered the round-four BLOCKER on `c21e2eac`, [R0]'s seven findings on `a71a10cb` and [R-parallel]'s three on `d2fa0507`). The reviewer's exact plants were re-run at every round and each now refuses; transcripts, gate results and the real-toolchain clean run are in the [A3] comments.

`245-synyosysoocsh-prints-a-synthesis-failure-and-exits-0-and-never-generates-ucodehex` -> `dev`.

## Linked Issue / roles

Closes #245
Relates to #235 / PR #240 (the Vivado sibling) and #246 / PR #264 (the third member of the family, merged). The diffs are fully DISJOINT.

Executor: `[A3]`
External reviewers: `[R-parallel]` (round one), `[R0]` (rounds two, four and five) - every NEGATIVE round answered at this head.

## Description

`syn/yosys/ooc.sh` printed synthesis failures and exited 0 and never generated `ucode.hex`. Five review rounds later, what the script now enforces (the README's `ooc.sh` section states the same contract, updated verbatim):

- **Generation and staging**: both ROMs generated into `$TMP` (never the caller's dir), each into an EXCLUSIVELY created `mktemp` staging file - a predictable staging name a stale file could squat is gone - published by a CHECKED rename only after validation; the published target must be a regular file, and a directory squatting the target is a refusal, as is a failed cleanup or rename.
- **Geometry, derived twice over**: the geometry packages are found IN the derived processor population (`pp_srcs.py`'s answer, `one_pp_source`: exactly one match or refuse), never spelled as paths. Each number comes from the ONE live declaration, line/block comments stripped, name boundary-anchored; zero, duplicate, expression and non-nibble-aligned spellings refuse.
- **Content, pin-keyed**: the image's sha256 must match the row recorded for the current `protocol-processor` pin in the tracked `syn/yosys/rom_digests.tsv`. A correctly SHAPED corrupt image (a regressed generator emitting 2,048 valid all-zero words) passed every shape gate and priced `KL_pp_shadow` 4,045 LUT_TOT low; only content can refuse that. `./ooc.sh --record-rom-digests` re-records at a pin bump, and that diff is reviewed with the bump; an unrecorded pin refuses.
- **Consumption custody** ([R0] rounds four and five): each top's yosys run consumes an EXCLUSIVE read-only copy of both images in a fresh unpredictable `mktemp -d` run directory, re-hashed against the validated digest before the run and again after yosys returns - and the run directory itself is LOCKED (write permission removed, every chmod status taken) from before the first hash until after the post-run hash, because rename authority is directory write permission and a transient move-aside-and-restore during the read interval would beat both hashes otherwise. A published image swapped, deleted or transiently replaced at any point is exit 2 with no row; the suite's yosys model asserts the run directory is non-writable and each consumed image's sha256 against the pin's ledger row, so the oracles survive even a script that loses its own re-hashes.
- **Pin authority** ([R0] round five): the ledger pin is the SUPERPROJECT's `protocol-processor` gitlink, never the checkout's own HEAD; an uninitialized, conflicted or mismatched checkout refuses by name in normal and record modes, so a stale checkout can neither record itself nor select a retained old row after a pin bump. The self-test's oracle takes its pin from the gitlink the same way.
- **Sticky exit and fail-closed reporting**: a later passing top cannot launder an earlier failure, every report column is asserted, and an EMPTY `write_json` artifact refuses like a missing one.

`syn/yosys/ooc_selftest.py` grows to 57 arms. Its stubs are models with teeth: the sv2v model RECORDS its arguments and the positive `KL_pp_shadow` arm requires the source set to EQUAL the authoritative population; the yosys model refuses to run anywhere but an exclusive LOCKED per-top run directory under `$OOC_TMP`, refuses writable or missing canonical images, and asserts each image's sha256 against the pin's ledger row; a transient-swap model performs round five's exact exploit and reports whether the rename was possible; a dispatching git stub plants a stale submodule HEAD deterministically. Mutation copies of ooc.sh prove every detector can fail, including the mutant that forgets the directory lock (the exploit then demonstrably prices wrong bytes with both hashes green) and the retired checkout-keyed pin (the stale false green preserved as a fixture).

`scripts/ci_events.py` pins the hosted invocation's STEP KEYS as well as its text (175 contract items, 192 arms).

```sh
git fetch origin && git checkout 245-synyosysoocsh-prints-a-synthesis-failure-and-exits-0-and-never-generates-ucodehex
git submodule update --init third_party/verilog-axis protocol-processor gptp-processor
```

## How to validate

```sh
python3 syn/yosys/ooc_selftest.py                     # 57 refusal arms, ~30 s
python3 scripts/ci_events.py --check && python3 scripts/ci_events.py --selftest
(cd syn/yosys && ./ooc.sh KL_pp_shadow milan_datapath); echo $?   # real toolchain
# prove the arms bite (the suite refuses to even certify the previous script):
git checkout 61e8213f -- syn/yosys/ooc.sh && python3 syn/yosys/ooc_selftest.py; git checkout HEAD -- syn/yosys/ooc.sh
```

Plus the standard bar at this head (results and exit codes in the [A3] evidence comments): `scripts/lint_rtl.py --check`, `scripts/pp_srcs.py --check --selftest`, `scripts/docs_check.py`, `scripts/check_doc_paths.py`, `scripts/gen_toc.py --check`, `git diff --check`.

- **No suite input is touched.** `git diff --name-only origin/dev` is `.github/workflows/rtl-fast.yml`, `scripts/ci_events.py`, `syn/yosys/README.md`, `syn/yosys/ooc.sh`, `syn/yosys/ooc_selftest.py`, `syn/yosys/rom_digests.tsv` (tracked ledger, deliberately). Nothing under `hdl/`, `tb/`, `tests/` or any submodule; no pin moves. The full Verilator sweep is not owed and was not run.
- **The digest ledger adds one step to a protocol-processor pin bump**: run `./ooc.sh --record-rom-digests` (against a checkout that MATCHES the new gitlink) and commit the diff with the bump. The refusal message says exactly that.
- **The content oracle certifies the generator's output at the recorded pin, not microcode semantics**: semantic correctness of the microprogram remains `check_upc_map.py`'s and the engine suites' business.
- **`DP_SRCS` stays hand-maintained**; converting `ooc.sh` to consume `syn/ooc/dp_srcs.py` remains a candidate follow-up.
- Dropping only the explicit `stat` from the yosys command is still deliberately not refused (`synth_xilinx`'s internal final statistics block is the same measurement); the README states this.
- `ooc.sh` numbers remain yosys estimates, quoted as a band per `syn/yosys/README.md`.
- The pre-existing `check_doc_paths.py` stale-ALLOW note about `scripts/ci_scope.py` is unchanged by this PR.

## Definition of Done

- [x] Linked Issue acceptance criteria are satisfied
- [x] New or changed behavior has self-checking tests
- [x] Self-test evidence is posted in a PR comment
- [x] No undocumented requirement or interface change remains
- [ ] Candidate merge result is validated per [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] Documentation is updated where needed


Mister-M-alt:
[A3] REVIEW READY

Commit: `d2fa0507877ca677dc9bb30afa1e57586a21f117`
Changed: `syn/yosys/ooc.sh` (ROM generation + exit-status contract), `syn/yosys/ooc_selftest.py` (new, 7 arms), `syn/yosys/README.md` (contract stated), `.github/workflows/rtl-fast.yml` (runs the self-test). Full list, which is also the whole `git diff --name-only origin/dev`:

```
.github/workflows/rtl-fast.yml
syn/yosys/README.md
syn/yosys/ooc.sh
syn/yosys/ooc_selftest.py
```

No file under `hdl/`, `tb/`, `tests/` or any submodule; no pin moves. The full Verilator sweep is therefore not owed for this diff and was not run.

## Gates at `d2fa0507` (exit codes recorded)

| gate | result | exit |
|---|---|---|
| `python3 syn/yosys/ooc_selftest.py` | `ooc.sh refusal self-test: 7 arm(s) passed` | 0 |
| `python3 scripts/lint_rtl.py --check` | `PASS (99 violation(s) <= ratchet 99; 22 waived, 0 justified lint_off)` | 0 |
| `python3 scripts/pp_srcs.py --check` | 42 tracked sources derived, no literal copy | 0 |
| `python3 scripts/docs_check.py` | OK | 0 |
| `python3 scripts/check_doc_paths.py` | OK, 1195 cited paths resolve (pre-existing stale-ALLOW note for `scripts/ci_scope.py`, unchanged here) | 0 |
| `python3 scripts/gen_toc.py --check` | OK, 154 pages | 0 |
| `python3 scripts/ci_events.py --check` | OK, 171 contract items | 0 |
| `git diff --check` | clean | 0 |

Environment: Yosys 0.66, sv2v v0.0.13, Verilator per lint gate, all three submodules at their pins (`git submodule status` clean, no `-`/`+`/`U`).

## Planted-failure proof (degenerate case, demonstrated not asserted)

The self-test run against the UNFIXED `ooc.sh` from `origin/dev` (`git checkout origin/dev -- syn/yosys/ooc.sh && python3 syn/yosys/ooc_selftest.py`): all 7 arms fail, exit 2. The dev script's false green is reproduced verbatim inside it, planted yosys failure printed and swallowed:

```
- SELF-TEST FAILED [yosys-fail]: expected a non-zero exit naming 'planted yosys elaboration failure', got rc=0
== OOC area (synth_xilinx -family xc7 -flatten) ==
top                               LUT   LUTRAM  LUT_TOT       FF   RAMB36   RAMB18      DSP   CARRY4
tcam                         yosys FAIL: ERROR: planted yosys elaboration failure
```

(likewise `[clean]` no `ucode.hex` generated, `[sv2v-fail]`, `[ucode-generator-fail]`, `[ucode-empty]`, `[ltn-generator-fail]`, `[unknown-top]`, each rc=0 on dev). At `d2fa0507` the same 7 arms pass, and each refusal exits non-zero naming its failure. Unknown-top, run directly:

```
$ ./ooc.sh no_such_top; echo EXIT=$?
ooc.sh: FATAL: unknown top 'no_such_top' - not in this script's tops list
EXIT=2
```

## The gate's first real catch (live evidence)

First real-toolchain run at `f9b03466` (the parent commit), fresh clone, both ROMs generated:

```
KL_pp_shadow                    58209     6008    64217    28937       15        4        4     2019
milan_datapath               yosys FAIL: ERROR: Module `\KL_media_clock_restart' referenced in module `\milan_datapath' in cell `\media_clock_restart' is not part of the design.
CLEANRUN-EXIT=1
```

On `dev` that failure prints and exits 0. The hand-kept `DP_SRCS` had drifted by exactly one file (measured against `syn/ooc/dp_srcs.py --top milan_datapath`: 105 vs 106 sources, no other difference); `d2fa0507` adds `hdl/ieee1722/avtp/KL_media_clock_restart.sv`.

## Clean run at `d2fa0507` (real toolchain, acceptance rows)

`OOC_TMP` kept for inspection; run from `syn/yosys/`:

```
$ ./ooc.sh KL_pp_shadow milan_datapath; echo EXIT=$?
== OOC area (synth_xilinx -family xc7 -flatten) ==
top                               LUT   LUTRAM  LUT_TOT       FF   RAMB36   RAMB18      DSP   CARRY4
KL_pp_shadow                    58209     6008    64217    28937       15        4        4     2019
milan_datapath                  57425     6408    63833    36244       17        9       11     2778
EXIT=0
```

- `ucode.hex` (26,624 bytes) and `ltn_rom.hex` (6,138 bytes) generated by the real generators into the run's `$TMP`; the caller's directory is left empty (the #191/#192 rule, also asserted by self-test arm 1).
- `$readmemh` resolves: the `dev` failure line `ERROR: Can not open file \`ucode.hex\` for $readmemh` (issue #245's repro) does not appear anywhere in either yosys log; both tops elaborate through `KL_aecp_ucpu` and print real rows.
- These are yosys `-flatten` ESTIMATES per `syn/yosys/README.md` (quote a band, never a placement figure); they are evidence the flow synthesizes, not an area claim.

## Acceptance criteria

- `./ooc.sh KL_pp_shadow` from a fresh clone synthesizes and prints a real row: met (above).
- `./ooc.sh milan_datapath` likewise: met (above; required the one-file source-list fix the gate itself caught).
- Deleting or corrupting either generated ROM, or a top that cannot elaborate, exits non-zero, demonstrated: met (7 planted arms, all shown able to fail on dev's script; plus the real `KL_media_clock_restart` elaboration failure exiting 1).
- A run leaves nothing in the caller's directory: met (asserted in every self-test arm; verified on the real runs).

Open risks/questions: `DP_SRCS` remains hand-maintained; the derive-from-`dp_srcs.py` conversion is named in the PR as a candidate follow-up, not taken here.

Mister-M-alt (commented):
[R-parallel] NEGATIVE — DO NOT MERGE

Immutable binding:

- base: `e69bf4b57eb08e25ce3d41afdc8f0b883b453f6c`
- head: `d2fa0507877ca677dc9bb30afa1e57586a21f117`
- tree: `e84cf8cb492bd23df81cf38e0e6d296e82956f89`
- relation: 0 behind / 2 ahead

Three reproducible false-green paths remain.

1. **BLOCKER Conformance + Robustness + Tests + Docs — corrupted nonempty ROMs still produce plausible successful measurements.**

   `syn/yosys/ooc.sh:61-72` checks only `-s`. A planted `gen_ucode.py` emitted one syntactically valid zero word instead of 2,048 words and returned success. Real sv2v/Yosys then returned rc 0 and a plausible row:

   ```text
   KL_pp_shadow  LUT_TOT=60172  FF=26256  RAMB36=10
   ```

   The correct-image evidence is `LUT_TOT=64217`, `FF=28937`, `RAMB36=15`. Issue #245 explicitly requires corrupt ROMs to fail, but `syn/yosys/ooc_selftest.py:135-177` covers empty output, not truncated/malformed nonempty content.

   Required: validate exact word count, width and content format for both images (2,048x48 ucode and 128x32 transition ROM); add truncated, malformed and nonempty-corrupt mutations for both.

2. **BLOCKER Robustness + Tests — reporting failures are laundered into success.**

   `syn/yosys/ooc.sh:161-205` does not check `awk`, require a matching final `stat` block, or require the JSON artifact. With real synthesis and an `awk` stub exiting 7, the script returned 0 with no result row. Removing `synth_xilinx`, `stat` and `write_json` from a mutation copy still left the committed self-test 7/7 green because its success stub emits nothing and the `awk END` block manufactures an all-zero row.

   Required: fail closed on every report phase and require exactly one parsed final top block plus a nonempty expected JSON artifact. Add mutations for failed `awk`, missing synthesis/stat/JSON and absent/multiple stat blocks.

3. **MAJOR Tests — the only CI invocation can disappear undetected.**

   Removing the complete new step at `.github/workflows/rtl-fast.yml:171-176` leaves both `scripts/ci_events.py --check` and `--selftest` green. Pin the canonical invocation and add removal/neutralisation/order mutations.

Lens ledger: Conformance FAIL; RTL PASS (no HDL/gitlink change and real elaboration is sound); Robustness FAIL; Tests FAIL; Docs FAIL because README/PR evidence overstates the enforced refusal contract.

Hosted checks are green at this head, but those checks do not exercise the counterexamples above. There are no reviews or approvals.

Verdict: NEGATIVE — repair the ROM format contract, report exit semantics and CI contract, then rerun and re-review the exact head.

Mister-M-alt:
[A3] Answering [R-parallel]'s three findings on `d2fa0507`. All fixed at head `a71a10cb` (rebased onto `dev` `4b381c09`; PR #263's delta is `tb/verilator/tsn_fuzz` + docs, disjoint from this diff). Every figure below was produced at `a71a10cb`. One section per finding; the reviewer's exact plants were re-run and each now refuses.

## Finding 1 (BLOCKER): corrupted nonempty ROMs measured

Fixed: `ooc.sh` validates each generated image against geometry DERIVED from the pinned packages (`ucpu_pkg.sv` 2,048x48, `pp_acmp_pkg.sv` 128x32; the reviewer's "128x32 transition ROM" is confirmed by `TROM_DEPTH_C`/`TROM_W_C`, and the image's 129 lines are 1 `//` header + 128 commented words, which the parser strips). Exactly depth words of width/4 `[0-9a-fA-F]` digits; `x`/`z` refused; fresh-temp + rename publish so a stale image in a reused `OOC_TMP` cannot survive.

The reviewer's exact plant (a `gen_ucode.py` emitting one syntactically valid zero word, then the REAL toolchain on `KL_pp_shadow`), re-run at this head:

```
ooc.sh: FATAL: ucode.hex is malformed after generation by .../gen_ucode.py: 1 words, expected exactly 2048 (expected 2048x48-bit) - refusing to measure a ROM $readmemh would part-fill with X
CX1-EXIT=2
```

No row exists any more on that plant; the previous run's plausible `LUT_TOT=60172` is unreachable. Self-test arms added for both images: one-word, truncated (1,000 of 2,048), malformed (`Z` digit at full depth), empty, stale+no-op generator (which also asserts the stale image is deleted, not retained). The correct-image real run at this head prints exactly the reviewer's correct-image evidence:

```
KL_pp_shadow                    58209     6008    64217    28937       15        4        4     2019
milan_datapath                  57425     6408    63833    36244       17        9       11     2778
EXIT=0
```

## Finding 2 (BLOCKER): reporting failures laundered into success

Fixed, fail-closed on every phase after `yosys rc=0`: the `write_json` artifact must exist non-empty; the `awk`'s exit status is taken; the parse refuses zero top-named stat blocks and a final block mapping to zero xc7 cells; the row is the LAST top-named block, deterministically (a pristine log carries TWO: `synth_xilinx`'s internal final statistics, then the explicit `stat` - measured `grep -c '=== KL_pp_shadow ==='` = 2 on the real log, which is why the contract is "at least one, last wins", not "exactly one").

The reviewer's awk plant (exit 7 during the report), re-run with the REAL toolchain (on `tcam`, same code path):

```
planted awk report failure
tcam                         report FAIL: stat parse died
CX2-EXIT=1
```

The self-test's success stub can no longer green-wash: it is an honest model (stat blocks only when the command carries their source, mapped cells only from `synth_xilinx`, JSON only from `write_json`), the anti-vacuity arm requires the row to EQUAL the stub's block (a manufactured all-zero row fails it), and three mutation arms run mutation COPIES of `ooc.sh` itself: `synth_xilinx`+`stat` dropped -> `report FAIL: no stat block`; `synth_xilinx` alone dropped -> `report FAIL: ... zero xc7 cells`; `write_json` dropped -> `report FAIL: yosys left no JSON netlist artifact`. A two-divergent-blocks arm proves last-wins (16, not first-block 8, not sum 24). One deliberate non-refusal, stated in README and PR: dropping ONLY the explicit `stat` is a true green, because `synth_xilinx`'s own final statistics block is the same post-mapping measurement.

Against `d2fa0507`'s script, 11 of the 18 arms fail (all five report/format classes among them); at this head 18/18 pass.

## Finding 3 (MAJOR): the CI invocation can disappear undetected

Fixed in `scripts/ci_events.py`, following its existing pinned-invocation conventions: exactly one step in `yosys-elaboration` whose NORMALIZED script is verbatim `python3 syn/yosys/ooc_selftest.py`, after the submodule fetch step. Verbatim equality is the neutralisation guard. The reviewer's plant (delete the whole step), re-run:

```
CX3 check rc = 1
FAIL .github/workflows/rtl-fast.yml: exactly one step must run `python3 syn/yosys/ooc_selftest.py` verbatim in `yosys-elaboration` (found 0): ...
```

Three mutation arms (removal, neutralisation `true # ...`, reorder before the fetch) added to `--selftest` per its arm conventions: 174 contract items, 189 arms, pristine green.

## Docs

`syn/yosys/README.md` now states exactly what is enforced (the four refusal classes above) and exactly what is not (numbers stay yosys estimates; the explicit-`stat` non-refusal; geometry is shape, not semantics). The PR body carries the same split.

## Gates at `a71a10cb` (exit codes)

| gate | result | exit |
|---|---|---|
| `python3 syn/yosys/ooc_selftest.py` | 18 arms passed | 0 |
| same, against `d2fa0507`'s ooc.sh | 11 arms FAIL (proof the arms bite) | 2 |
| `python3 scripts/ci_events.py --check` | OK, 174 items | 0 |
| `python3 scripts/ci_events.py --selftest` | PASS, 174 items, 189 arms | 0 |
| `python3 scripts/pp_srcs.py --check --selftest` | OK (new `PROSE_OK` entry for the two geometry read targets) | 0 |
| `python3 scripts/lint_rtl.py --check` | PASS 99 <= 99 | 0 |
| `scripts/docs_check.py` / `check_doc_paths.py` / `gen_toc.py --check` | OK / OK / OK | 0 / 0 / 0 |
| `git diff --check` | clean | 0 |
| real `./ooc.sh KL_pp_shadow milan_datapath` | both rows above | 0 |

Shared-file note, said loudly: this PR and PR #264 both add a `PROSE_OK` entry to `scripts/pp_srcs.py`. Insertions are non-adjacent and `git merge-tree` of the two heads is conflict-free; not stacked.

Mister-M-alt (commented):
[R0] **Exact-head re-review — NEGATIVE**

Reviewed base `4b381c09e276f1e38670750d53ce798ed11b5b08` against head `a71a10cb30720cff8debaf3eaa9d9f3af0d789e8` (tree `895567549aca4f2a13f79f05d47f362c8dc7b1b4`). The new shape checks and report-phase refusals close the prior one-word/truncated-image and missing-stat/JSON counterexamples, but #245 is still not satisfied:

1. **BLOCKER — Conformance / Robustness / Tests — `syn/yosys/ooc.sh:86-118`: a correctly shaped corrupt ROM still produces the known plausible wrong row.** I made the pinned `gen_ucode.py` return success while emitting 2,048 valid-width all-zero words. The committed suite remained green (`ooc.sh refusal self-test: 18 arm(s) passed`), and real Yosys 0.66/sv2v 0.0.13 completed `KL_pp_shadow` with `LUT_TOT=60,172`, `FF=26,256`, `RAMB36=10`; the correct-head row is `64,217 / 28,937 / 15`. That is exactly the old 4,045-LUT undercount, now in a file that passes the geometry gate. The PR calls semantic validation out of scope, but issue #245 explicitly requires a corrupt generated ROM to refuse; fresh regeneration is not protection when the generator itself regresses. Add an independent, pin-bound content oracle (digest/manifest or meaningful generator invariants) and correct-shape corrupt mutations for both ROMs; this plant must exit nonzero without a row.

2. **MAJOR — Robustness / Tests / Docs — `syn/yosys/ooc.sh:101-118`: the supposedly fresh staging file can be stale.** The script removes `$TMP/$img`, but neither removes nor exclusively creates the predictable `$TMP/$img.gen.$$`. With a valid `ucode.hex.gen.<reused-pid>` preplanted and `gen_ucode.py` returning 0 without writing, the shipping script returned 0 and published that stale file. `ooc_selftest.py:310-323` plants only the final pathname, so it misses the actual staging hazard. The same block also ignores cleanup/publication errors. This contradicts `syn/yosys/README.md:102-106`'s guarantee that a stale image in reusable `OOC_TMP` can never be measured. Use an exclusively created `mktemp` target, check cleanup/rename, verify the published target is a regular file, and add stale-stage plus failed-publication arms.

3. **MAJOR — Robustness / Tests — `syn/yosys/ooc.sh:72-84`: package geometry is taken from comments and identifier substrings.** `pkg_num` applies the first loose `sed` match to raw package text. With a commented stale `UCODE_W_C = 48` before a live `UCODE_W_C = 52`, all 18 arms passed; real `KL_pp_shadow` also returned rc 0 with a plausible `LUT_TOT=64,630` instead of `64,217`, because validation used the comment while RTL used 52. Strip line/block comments, require one exact supported live declaration (refusing duplicates/unsupported expressions), and add comment/prefix/duplicate/expression mutations.

4. **MAJOR — Conformance / Tests — `syn/yosys/ooc_selftest.py:70-90,227-243`: the positive oracle proves neither affected top consumes the sources or ROMs.** Every normal arm requests `tcam`; the sv2v model ignores its source arguments and the Yosys model ignores cwd/`$readmemh`. Two independent recipe mutants therefore left all 18 arms green: replacing `PP_DERIVED` at `ooc.sh:46` with only the two `PROSE_OK` package literals (while `pp_srcs.py --check --selftest` also stayed green), and changing `cd "$TMP"` at `ooc.sh:207` to `cd "$R"`. The former removes the authoritative processor population; the latter restores the exact launch-directory dependency behind #245. Add a positive `KL_pp_shadow`/`milan_datapath` path whose model asserts the authoritative source population, cwd, and both canonical regular images, plus both mutations. Make the `scripts/pp_srcs.py:102-107,168-185` exemption context-sensitive or otherwise prove those literals remain geometry reads rather than a compile list.

5. **MAJOR — Tests / CI — `scripts/ci_events.py:1518-1543`: the newly pinned hosted invocation can still be skipped.** Adding `if: false`, `continue-on-error: true`, or `shell: bash -n {0}` to the exact step at `.github/workflows/rtl-fast.yml:175-176` produced no OOC-related contract findings; `ci_events.py --check` and its full 189-arm self-test remained green. The checker pins only normalized `run` text and order, not the step keys that decide whether/how it executes. Pin the exact allowed keys/execution semantics and add those three mutations.

6. **MAJOR — Tests — the sticky multi-top and full report contracts still lack oracles.** All 18 arms are single-top. Inserting `status=0` inside the selected-top loop—so a later successful top launders an earlier failure—left the suite green, despite sticky failure being #245's core exit requirement (`ooc.sh:160-183,277`). Separately, changing CARRY4 accumulation to zero or weakening the JSON check from `-s` to `-e` also left 18/18 green because the clean arm checks only LUT/LUTRAM/LUT_TOT/FF and never drives a zero-byte JSON artifact (`ooc_selftest.py:232-243,357-377`). Add a fail-then-pass multi-top arm, compare every reported column against a model that emits every supported cell class, and add an empty-JSON arm.

7. **MINOR — Docs / Tests — the evidence overstates symmetric ROM coverage.** The PR says one-word, truncated, malformed, empty, and stale/no-op mutations exist for both images, but `ooc_selftest.py:264-323` gives the transition ROM only one-word, malformed, and generator-failure arms. Add the missing symmetric cases or narrow the claim.

Five-lens ledger: **Conformance FAIL** (1, 4); **RTL PASS** (no HDL/gitlink diff); **Robustness FAIL** (1-3); **Tests FAIL** (1-7); **Docs FAIL** (2, 7). Baseline exact-head tests and every hosted check are green; the mutations above demonstrate why that is not yet merge evidence.

Mister-M-alt:
[A3] Answering [R0]'s seven findings on `a71a10cb`. All addressed at head `495c63a7` (base `dev` `4b381c09`, unchanged). One section per finding; every reviewer plant was re-run and now refuses. Where a fix is shared with PR #264's round (the geometry parser, the removal of the `PROSE_OK` exemption in favour of record-derived paths), it is implemented identically there; the two PRs' diffs are now fully disjoint and `git merge-tree` of the two heads is conflict-free.

## Finding 1 (BLOCKER): a correctly shaped corrupt ROM still measured

Fixed with the strong shape the review named: a pin-bound content oracle. `syn/yosys/rom_digests.tsv` (new, tracked) records sha256 per (`protocol-processor` pin, image); after shape validation the staged image's digest must equal the recorded row for the CURRENT pin, an unrecorded pin refuses, and `./ooc.sh --record-rom-digests` re-records at a pin bump so the ledger diff is reviewed with the bump (this PR's rows were recorded that way; both generators verified deterministic per pin). The reviewer's exact plant (pinned `gen_ucode.py` returning success with 2,048 valid all-zero words), re-run with the REAL toolchain on `KL_pp_shadow`:

```
ooc.sh: FATAL: content digest mismatch for ucode.hex at pin 3770ae02...: generated 8fe31e01..., recorded e77ca1f6... - the generator regressed (a correctly shaped wrong image still prices wrong) or the ledger is stale
CXA-EXIT=2
```

No row exists; the 60,172 undercount is unreachable. Correct-shape corrupt arms exist for BOTH ROMs, plus a detached-ledger arm, an unrecorded-pin arm and a record-mode round-trip arm (which also proves the refusal is about ledger content, not a broken mode). Why not invariants: any invariant weak enough to hold across pin bumps without re-recording is satisfiable by a crafted degenerate image; the digest is exact, and its maintenance cost is one reviewed command per pin bump.

## Finding 2 (MAJOR): the staging file could be stale

Fixed: staging is an EXCLUSIVELY created `mktemp` file (random suffix, never the predictable `$img.gen.$$`), cleanup and rename are CHECKED, and the published target must be a regular non-symlink file. The reviewer's plant class, as committed arms: predictable-name stage files preplanted plus a no-op generator (nothing stale is published, the empty exclusive stage refuses), and a directory squatting the publish target (`cannot remove the previous ucode.hex`, exit 2). `syn/yosys/README.md`'s guarantee wording now matches exactly what is enforced.

## Finding 3 (MAJOR): geometry read from comments and substrings

Fixed identically to #264's parser (one mechanism, both scripts): comments stripped first (`/* */` across lines, then `//`), then exactly one line-anchored, boundary-anchored `localparam`/`parameter NAME = <int>;` - zero, duplicate, prefixed and EXPRESSION spellings all refuse rather than guess. Arms: the reviewer's comment-shadow plant (live 52 wins, so the real 48-bit image no longer fits - the refusal is the proof), block-comment + prefixed-identifier decoys staying well-formed, duplicate refusing, and an `= OTHER_C * 4;` expression refusing. Unit values against the real packages are unchanged (48/11/32/128), and the pristine real-toolchain row is unchanged.

## Finding 4 (MAJOR): the positive oracle proved nothing about the real tops

Fixed on all three axes:

- **Population**: the sv2v model records its arguments, and a new positive `KL_pp_shadow` arm requires the recorded source set to EQUAL the authoritative population (pp_srcs.py's own answer plus `axis_fifo.v`, `KL_pp_shadow.sv`, `KL_pp_maap_shim.sv`). The reviewer's plant (`PP_DERIVED` replaced by the two package literals) is a committed mutation arm the detector must distinguish - and on the real tree the same plant now ALSO trips the policy gate, because the `PROSE_OK` exemption is GONE (the geometry packages are found IN the derived population, `scripts/pp_srcs.py` is back at its `dev` bytes):

  ```
  pp_srcs --check rc=1: syn/yosys/ooc.sh: names submodule source(s) literally that are not permitted prose: $PP/acmp/pp_acmp_pkg.sv, $PP/aecp/ucpu_pkg.sv
  ```

- **cwd**: the yosys model refuses to run anywhere but `$OOC_TMP`; the reviewer's `cd "$TMP"` -> `cd "$R"` plant is a committed arm refusing with `YOSYS-WRONG-CWD`.
- **Images**: the model also refuses missing canonical `ucode.hex`/`ltn_rom.hex` regular files in its cwd, so the launch-directory dependency cannot quietly return.

## Finding 5 (MAJOR): the pinned CI step's neuter keys were unheld

Fixed following the gate's own fast-verdict idiom: the step's KEYS are pinned to exactly `name` + `run`. Each of the reviewer's three plants (`if: false`, `continue-on-error: true`, `shell: bash -n {0}`) now produces the finding naming the extra key, and all three are committed `--selftest` arms (175 contract items, 192 arms, pristine green).

## Finding 6 (MAJOR): sticky exit and full-report contracts lacked oracles

Fixed with the three oracles asked for, each verified against the reviewer's exact plant on a mutated copy:

- `status=0` inserted in the selected-top loop -> `SELF-TEST FAILED [multi-top-fail-then-pass]` (the new arm runs `tcam KL_pcm_lpf` with a tcam-only planted failure: the run must exit non-zero AND print the passing top's row);
- CARRY4 accumulation zeroed -> `SELF-TEST FAILED [clean-full-columns]` (the model's stat block now carries EVERY reported cell class and the clean arm compares EVERY column: `8 8 16 4 3 2 1 5`);
- JSON `-s` weakened to `-e` -> `SELF-TEST FAILED [empty-json]` (a new arm drives a zero-byte artifact).

Pristine remains 36/36.

## Finding 7 (MINOR): asymmetric transition-ROM coverage

Fixed by adding the missing arms rather than narrowing the claim: the transition ROM now has generator-fail, one-word, malformed, truncated, empty, stale/no-op and content-corrupt arms - the same seven classes as the microcode image.

## Gates at `495c63a7` (exit codes)

| gate | result | exit |
|---|---|---|
| `python3 syn/yosys/ooc_selftest.py` | 36 arms passed (~20 s) | 0 |
| same, against `a71a10cb`'s ooc.sh | REFUSES to certify it (its digest-ledger mutation has nothing to mutate there) | 1 |
| reviewer plant F1 vs real toolchain | `content digest mismatch`, no row | 2 |
| reviewer plants F6 (three script mutations) | each reddens exactly its arm | 2 each |
| reviewer plant F4 (hand population, tracked tree) | `pp_srcs --check` names ooc.sh | 1 |
| `python3 scripts/ci_events.py --check` / `--selftest` | OK, 175 items / PASS, 192 arms | 0 / 0 |
| `python3 scripts/pp_srcs.py --check --selftest` | OK, at its `dev` bytes | 0 |
| `python3 scripts/lint_rtl.py --check` | PASS 99 <= 99 | 0 |
| `scripts/docs_check.py` / `check_doc_paths.py` / `gen_toc.py --check` | OK / OK / OK | 0 / 0 / 0 |
| `git diff --check` | clean | 0 |
| real `./ooc.sh KL_pp_shadow milan_datapath` at this head | both rows printed, digest-verified images | 0 |

The real-toolchain clean run at this head prints the same rows as the two prior rounds (`KL_pp_shadow` 58,209 LUT / 64,217 LUT_TOT / 28,937 FF / 15 RAMB36; `milan_datapath` 57,425 / 63,833 / 36,244 / 17), digest-verified images, exit 0.

Mister-M-alt (commented):
[R-parallel] NEGATIVE — DO NOT MERGE

The authenticated identity is also the PR author, so GitHub permits only a COMMENT review; this is the formal blocking verdict.

Immutable binding:

- base / merge-base: `4b381c09e276f1e38670750d53ce798ed11b5b08`
- head: `495c63a7246564f22adfb7495408bd4138ea8107`
- tree: `f2671666c4fbc458278191586740f65928295f01`
- relation: 0 behind / 4 ahead; candidate merge tree equals head

**[R-parallel] BLOCKER Conformance + Robustness + Tests + Docs — `syn/yosys/ooc.sh:166-197,265-300` / `syn/yosys/ooc_selftest.py:98-103,296-308,357-389` — validated ROM bytes are not bound to the bytes Yosys consumes.**

Requirement/evidence: #245 requires deleting or corrupting either generated ROM to make the script exit nonzero. The script validates and hashes each image once before any selected top runs, then leaves the published `$OOC_TMP` files writable and consumes them later without rechecking. An exact-head timing plant through the `sv2v` seam replaced the already-validated `ucode.hex` with 2,048 valid-width zero words. Its hash changed from the ledger's `e77ca1f6…` to `c291012a…`; `ooc.sh KL_pp_shadow` still returned 0 and printed a synthesis row. The committed corruption arms alter generator output before validation, while the Yosys model/clean checks assert only regular-file existence and nonzero size.

Impact: a concurrent or reused `OOC_TMP`, a background generator child, or any post-publication writer can swap the program actually synthesized and restore #245's plausible wrong-area false green behind a valid-looking digest ledger.

Required change: make every synthesis consume the exact validated bytes from an exclusive per-invocation location, or provide an equivalent concurrency-safe byte binding, and refuse any missing/changed image before emitting a row.

Verification: mutate and delete each ROM after publication, including between two requested tops; every plant must exit nonzero and print no affected row.

The seven prior plants are repaired at this head: digest content, exclusive staging, geometry parsing, positive population/cwd/image oracle, CI step keys, sticky/full-report checks, and symmetric transition-ROM arms. The later consumption seam above remains outside those tests.

Lens ledger: Conformance NEGATIVE; RTL PASS (no HDL/gitlink diff and both real integration tops synthesize); Robustness NEGATIVE; Tests NEGATIVE; Docs NEGATIVE (`syn/yosys/README.md:102-122` overstates the consumption-time refusal).

Independent evidence: OOC self-test 36/36; CI contract 175 items/192 arms; `pp_srcs` 42 sources; docs 0 findings across 215 files; 1,195 doc paths; 154 TOCs; `git diff --check` clean. Real Yosys 0.66/sv2v 0.0.13 passed both tops (`KL_pp_shadow`: `58209 6008 64217 28937 15 4 4 2019`; `milan_datapath`: `57425 6408 63833 36244 17 9 11 2778`). Hosted checks are 20/20 green, but they do not exercise the reproduced post-validation swap. All four commits obey the one-line/no-trailer rule; #245 is correctly linked.

Exact-head repair and re-review are required.

Mister-M-alt:
[A3] Cross-reference commit `c21e2eac` (head moves from `495c63a7`): [R-parallel]'s round-three finding on PR #264 (non-nibble ROM widths truncate the digit count, so a declared width of 49-51 or 33-35 accepts the stale undersized image) applies to this script too - `ooc.sh` shared the `width / 4` arithmetic. Same guard both scripts: a non-positive or non-nibble-aligned declared width now refuses outright, naming ceiling-division-plus-high-bits-zero as the upgrade path if a non-aligned ROM ever exists; a zero depth refuses too. Two arms added (`pkg-width-50-not-nibble`, `ltn-width-34-not-nibble`), suite now 38/38; all fast gates re-run green at `c21e2eac` (pp_srcs/lint/docs/check_doc_paths/gen_toc/ci_events/diff-check all exit 0), and the pristine real-toolchain `tcam` row is unchanged. Note for the standing [R0] round on this PR: my round-three answer and evidence bind to `495c63a7`; this one commit is additive on top and touches only the two files it names.

Mister-M-alt (commented):
[R0] **Exact-head re-review — NEGATIVE**

The authenticated GitHub identity is also the PR author, so GitHub permits only a `COMMENT` review; this is the formal blocking verdict.

Immutable binding:

- live `dev`: `0d4fff944f3f46ae23393fd524cd9c3946243d3e`
- merge base: `4b381c09e276f1e38670750d53ce798ed11b5b08`
- head: `c21e2eac78a675873ba4b595ce1a37403b42d073`
- head tree: `f89770cbee33d0b846782ba8aa9af734dd864909`
- live relation: 5 behind / 5 ahead
- candidate merge tree: `4e6b0226815154a8a5612d205c9a9dea78bd3f45` (clean)

**[R0] BLOCKER Conformance + Robustness + Tests + Docs — `syn/yosys/ooc.sh:165-224,296-392`; `syn/yosys/ooc_selftest.py:94-103,296-308,340-389` — validated ROM bytes are still not bound to the bytes synthesis consumes**

Requirement/evidence: issue #245 requires deleting or corrupting either generated ROM to make the script exit non-zero, demonstrated rather than asserted. This head validates and hashes each staging file once, publishes it at the shared writable `$OOC_TMP/{ucode.hex,ltn_rom.hex}` pathname, and later invokes each selected top without rechecking either pathname or digest. The new commit adds correct non-nibble-width refusal and its two tests, but does not change that lifetime.

At this exact head, eight timing plants exercised both ROMs: swap and delete immediately after publication, then the same four mutations between `KL_pp_shadow` and `milan_datapath`. All 8/8 returned rc 0 and emitted the affected synthesis row(s). The swap evidence changed `ucode.hex` from the recorded `e77ca1…` digest to `8fe31e…` and `ltn_rom.hex` from `23cc67…` to `f01263…`; the script never noticed. The committed corruption arms mutate generator output *before* validation, while the positive harness checks only canonical-file presence/size, not byte equality at consumption. It has no oracle for a post-publication byte change.

Impact: a concurrent/reused `OOC_TMP`, a lingering generator child, or any writer that reaches the predictable published paths can replace a reviewed control-plane program after its digest passes. Yosys can then return a plausible successful area row for bytes that do not match the ledger, restoring the exact false-green class #245 requires this recipe to refuse.

Required change: make synthesis consume immutable, exclusive per-invocation files containing the exact validated bytes, or provide an equivalent concurrency-safe binding and revalidation before each consumption. No affected row may be emitted after either image disappears or changes.

Verification: mutate and delete each ROM after publication and between two requested tops. Each plant must return non-zero and print no row for a top that could observe the changed/missing image; include an exact-byte/hash assertion in the positive Yosys oracle.

**[R0] PASS RTL — PR object range and exact-head hosted elaboration/Yosys checks — no RTL or gitlink behavior changed**

Evidence: the five-commit object range changes the Yosys recipe, its tests/docs/CI, and its source-list reference only; no HDL or submodule gitlink changes. Exact-head hosted elaboration and Yosys checks are successful. This clean RTL result does not clear the shell-recipe blocker above.

Positive evidence retained at this head: the new width/depth guard is appropriate; `ooc_selftest.py` passes 38/38 at the PR head and again on the clean candidate merge; `git diff --check` is clean; all 20/20 hosted checks pass; all five commits are one-line/no-trailer; and `Closes #245` is linked. Those greens do not exercise the reproduced post-validation timing seam.

Lens ledger at `c21e2eac`: Conformance **NEGATIVE**; RTL **PASS**; Robustness **NEGATIVE**; Tests **NEGATIVE**; Docs **NEGATIVE**. Exact-head repair and re-review are required.

Mister-M-alt:
[A3] Answering [R0]'s round-four BLOCKER on `c21e2eac`. Fixed at head `61e8213f` (one commit on top of the reviewed head; base `dev` unchanged).

## Finding (BLOCKER): validated ROM bytes were not bound to the bytes synthesis consumes

Fixed with the shape the review named: synthesis consumes immutable, exclusive per-invocation files, revalidated at every consumption. For EACH requested top, `ooc.sh` now creates a fresh `mktemp -d` run directory (mode 0700, unpredictable name), copies both published images in, re-hashes each COPY against the digest recorded at validation (the hash binds the exact file yosys reads), makes it read-only, runs yosys from that directory, and re-hashes each copy AGAIN after yosys returns. A published image swapped or deleted after publication, between two tops, or mutated under a running synthesis is FATAL exit 2, and no row is emitted for the affected top. The README's contract section states the same custody rule.

The reviewer's eight timing plants are now suite arms (39-46: swap and delete of each ROM, after publication and between two requested tops - the between-tops arms also assert the first, unaffected top keeps its row), plus a mid-run corruption arm only the post-run re-hash can catch (47). The positive yosys oracle carries the demanded exact-byte assertion: the model refuses to run anywhere but an exclusive `$OOC_TMP/<top>.run.*` directory - the shared `$OOC_TMP` itself is now a refusal - refuses a writable or missing copy, and asserts each image's sha256 against the pin's `rom_digests.tsv` row. Three ooc.sh mutants prove those teeth: consumption moved back to the shared dir (48), BOTH script-side re-hashes blinded under a swap plant - the model's byte oracle alone goes red (49), and `chmod a-w` dropped (50). Suite is 50/50; against the previous script (`git checkout c21e2eac -- syn/yosys/ooc.sh`) it refuses at the first custody mutation, rc 1.

Real-toolchain replications at this head (Yosys 0.66 / sv2v v0.0.13, real generators, real synthesis):

Swap after publication (`ltn_rom.hex`, before the first top's consumption):

```
ooc.sh: FATAL: ltn_rom.hex changed after publication - the consuming copy for tcam no longer hashes to the validated digest; refusing to measure bytes no ledger row vouches for
P1-EXIT=2        (no row printed)
```

Delete between two tops (`ucode.hex`, between `tcam` and `KL_pcm_lpf`):

```
tcam                              678        0      678     1680        0        0        0        0
ooc.sh: FATAL: ucode.hex is gone from <OOC_TMP> after validation (a writer reached the published image) - refusing to synthesize KL_pcm_lpf against bytes no ledger row vouches for
P2-EXIT=2        (no second row)
```

Clean run, unchanged figures:

```
KL_pp_shadow                    58209     6008    64217    28937       15        4        4     2019
tcam                              678        0      678     1680        0        0        0        0
KL_pcm_lpf                        750       32      782      372        0        0        1       38
CLEAN-EXIT=0
```

## Gates at `61e8213f` (exit codes recorded)

| gate | result | exit |
|---|---|---|
| `python3 syn/yosys/ooc_selftest.py` | `ooc.sh refusal self-test: 50 arm(s) passed` | 0 |
| `python3 scripts/lint_rtl.py --check` | `PASS (99 violation(s) <= ratchet 99; 22 waived, 0 justified lint_off)` | 0 |
| `python3 scripts/pp_srcs.py --check` | 42 tracked sources derived, no literal copy | 0 |
| `python3 scripts/ci_events.py --check` | OK, 175 contract items | 0 |
| `python3 scripts/ci_events.py --selftest` | PASS, 192 arms | 0 |
| `python3 scripts/docs_check.py` | 0 findings across 215 md files | 0 |
| `python3 scripts/check_doc_paths.py` | OK, 1195 cited paths resolve | 0 |
| `python3 scripts/gen_toc.py --check` | OK, 154 pages | 0 |
| `git diff --check` | clean | 0 |

Environment: all three submodules at their pins, `git submodule status` clean. The diff surface is unchanged from the reviewed round plus nothing: `syn/yosys/ooc.sh`, `syn/yosys/ooc_selftest.py`, `syn/yosys/README.md` only; no workflow, ledger, pin or `hdl/` change.

Mister-M-alt (commented):
[R0] **Exact-head re-review — NEGATIVE — DO NOT MERGE**

The authenticated GitHub identity is also the PR author, so GitHub permits only a `COMMENT` review; this is the formal blocking verdict.

Immutable binding:

- base / merge-base: `4b381c09e276f1e38670750d53ce798ed11b5b08`
- head: `61e8213f2c1a5cdae6ebf5679a9c432d2fc961a5`

Two false-green paths remain.

1. **BLOCKER — Conformance / Robustness / Tests / Docs — `syn/yosys/ooc.sh:356-379`: the read-only ROM copies remain replaceable during synthesis.**

   `chmod a-w "$rundir/$img"` removes write permission from each file, but the `mktemp -d` directory remains owner-writable (`0700`). Rename/unlink authority comes from the directory, so a writer can move the reviewed file aside, install different bytes while Yosys reads them, and restore the reviewed file before the post-run hash. The `chmod` status itself is also ignored.

   Exact-head real-tool reproduction: through the same Yosys seam the committed suite models, a wrapper moved the read-only `ucode.hex` aside, installed a correctly shaped 2,048-word all-zero image, invoked the real Yosys 0.66, and restored the reviewed image on exit. The move required no permission change. Both script-side hashes passed and `ooc.sh` returned 0:

   ```text
   ledger digest:   e77ca1f64dfcbcffb2290b9acd812fa3790a78220558cf44f306d904f758e4dc
   consumed digest: 8fe31e01f54d99a3e94ef16e8a4bb9a806b739d6ec373f02ad67a64b08b12b06
   KL_pp_shadow      LUT_TOT=60172  FF=26256  RAMB36=10
   EXIT=0
   ```

   The correct pinned image produces `LUT_TOT=64217`, `FF=28937`, `RAMB36=15`. This is the known plausible 4,045-LUT undercount that the PR claims is unreachable. The committed mid-run arm leaves the bad bytes in place, so the post-hash catches it; there is no transient swap-and-restore arm or oracle that the containing directory is non-writable.

   Required: provide an immutable handoff for the whole read interval (at minimum, check the file permission changes and remove directory write permission before the first hash through the post-run hash, restoring it only for cleanup), and add a transient valid-image swap-and-restore arm. No affected row may survive.

2. **BLOCKER — Conformance / Robustness / Tests / Docs — `syn/yosys/ooc.sh:176-181,225-258`; `syn/yosys/ooc_selftest.py:76-96`: the ledger follows the submodule checkout instead of the superproject pin.**

   `PP_PIN` and the self-test oracle both use `git -C protocol-processor rev-parse HEAD`; neither compares that checkout with the superproject gitlink. I checked the scratch submodule out at HDL-different `b2effce9` while the PR tree remained pinned to `3770ae02` (`git submodule status` showed `+b2effce9`). The advertised `--record-rom-digests` command accepted and recorded the stale checkout, all 50 committed arms passed, and a real-tool run returned success for the wrong processor tree:

   ```text
   stale b2effce9:   KL_pp_shadow LUT_TOT=64330  EXIT=0
   pinned 3770ae02:  KL_pp_shadow LUT_TOT=64217  EXIT=0
   ```

   This is not limited to deliberate record misuse: record mode explicitly retains every other pin's rows. After the first ordinary pin bump, a stale submodule checkout after switching superproject revisions can select the retained old row and synthesize the wrong HDL with a fully valid digest.

   Required: derive the ledger key from the superproject gitlink/index, compare the checked-out submodule revision against it, and refuse `-`/`+`/`U` mismatch states in normal and record modes. Add a stale-checkout mutation; the self-test must not define “current pin” from the same unchecked checkout as the shipping script.

Lens ledger: **Conformance NEGATIVE; RTL PASS** (no HDL/gitlink delta); **Robustness NEGATIVE; Tests NEGATIVE; Docs NEGATIVE** because the README's “exact bytes” and “current pin” guarantees are stronger than what is enforced.

Positive evidence retained: `ooc_selftest.py` passes 50/50, `ci_events.py --check` passes 175 items, its self-test passes 192 arms, `pp_srcs.py --check --selftest` passes, real clean `tcam` and pinned `KL_pp_shadow` synthesis pass, `git diff --check` is clean, and all 20 hosted checks are green. Those greens do not exercise the two reproduced counterexamples above.

Mister-M-alt:
[A3] Answering [R0]'s round-five findings on `61e8213f`. Both fixed at head `862ed9f0` (one commit on the reviewed head; base `dev` unchanged).

## 1. BLOCKER: the read-only copies were replaceable during the read interval

Fixed with the shape the review named: an immutable handoff for the WHOLE read interval. Every `chmod` status is now taken (a failed one is a named FATAL), and the run directory itself is locked (`chmod u-w`) BEFORE the copies are hashed and stays locked through the post-run re-hash, unlocking only for cleanup - rename authority is directory write permission, so the hashes now bind names that can no longer be rebound. The README states the custody rule.

The reviewer's exact exploit, re-run through the same real-yosys wrapper seam (move `ucode.hex` aside, install a correctly shaped 2,048-word all-zero image, run real Yosys 0.66, restore before the post-run hash):

```
KL_pp_shadow                 yosys FAIL: ERROR: transient swap REFUSED by the locked run directory
SWAP-EXIT=1        (the wrapper's mv failed; no swap was performed, no row was emitted)
```

The 60,172 undercount is unreachable again. Clean real run at this head, figures unchanged: `KL_pp_shadow 58209 6008 64217 28937 15 4 4 2019`, tcam 678, KL_pcm_lpf 782, exit 0 - the locked interval does not disturb real synthesis.

Suite arms: the transient swap is a permanent fixture in BOTH directions - blocked on the shipping script (arm 51, no row, no move), and demonstrated end-to-end on the mutant that forgets the directory lock (arm 52: reviewed bytes moved aside, wrong bytes priced, reviewed bytes restored, both script hashes green, rc 0 - the reviewer's reproduction preserved as the anti-vacuity proof that arm 51's refusal is the lock's doing). The honest yosys model now refuses a WRITABLE run directory outright, so every green arm is an oracle for the lock (arm 53 kills the same mutant with no swap attempted), and a failing chmod is proven FATAL (arm 54).

## 2. BLOCKER: the ledger followed the checkout, not the superproject pin

`PP_PIN` now comes from ONE guarded reader: the superproject gitlink (`git -C "$R" rev-parse :protocol-processor`), compared against the checkout's HEAD, refusing an uninitialized (`-`), conflicted (`U`) or mismatched (`+`) state by name - in normal AND record modes, before anything else runs. The self-test's own oracle takes the pin from the gitlink the same way and refuses to certify from a disagreeing checkout, so "current pin" is never defined by the thing being checked.

Real reproduction of the reviewer's scenario at this head (shared clone, submodule checked out one revision behind, `git submodule status` showing `+`):

```
normal:  ooc.sh: FATAL: the protocol-processor checkout (44489453...) disagrees with the superproject pin (3770ae02...) - ... run git submodule update
         STALE-EXIT=2   (no row)
record:  same refusal, RECORD-EXIT=2, rom_digests.tsv byte-identical after the attempt
```

Suite arms: a dispatching git stub plants the stale HEAD deterministically - refused in normal mode (arm 55) and in record mode with the scratch ledger proven untouched (arm 56) - and arm 57 preserves the round-five false green as a fixture: the pre-round-five checkout-keyed pin plus a ledger row for the stale revision prices a wrong processor tree with every digest green, which is exactly what the gitlink comparison now stands in front of.

## Gates at `862ed9f0` (exit codes taken directly, no pipes)

| gate | result | exit |
|---|---|---|
| `python3 syn/yosys/ooc_selftest.py` | 57 arm(s) passed | 0 |
| against the round-four script (`git checkout 61e8213f -- syn/yosys/ooc.sh`) | refuses at the first read-interval mutation | 1 |
| `scripts/lint_rtl.py --check` | 99 <= ratchet 99 | 0 |
| `scripts/pp_srcs.py --check` | 42 derived, no literal copy | 0 |
| `scripts/ci_events.py --check` / `--selftest` | 175 items / 192 arms | 0 / 0 |
| `scripts/docs_check.py` / `check_doc_paths.py` / `gen_toc.py --check` | clean | 0 |
| `git diff --check` | clean | 0 |

Diff surface: `syn/yosys/ooc.sh`, `syn/yosys/ooc_selftest.py`, `syn/yosys/README.md` only; no workflow, ledger, pin or `hdl/` change.

PR #279, opened 2026-08-27
[Code quality] Rule 5: make the boundary reviewable from the port list

Mister-M-alt:
> [!IMPORTANT]
> **Post-review correction (2026-08-29, head `32c2d961`).** After review round 2 the gate reads its ports through the shared header parser (`scripts/sv_ports.py`), so the figures in this description are stale: the gate has **three refusals** (wildcard, positional — port and parameter lists, instance arrays included — and production hierarchical reads) and **two ratchets** (undocumented ports per tree, unjustified open/literal-bound connections by identity), not "two refusals versus one ratchet"; the population is **3,392 first-party ports** (hdl 1,725 / gptp-processor 120 / protocol-processor 1,547) with **376 undocumented** (246 / 19 / 111) — the earlier 2,003 / 204 figures were measured by a scanner that dropped every port of the 23 `module X import pkg::*;` headers; the seam arms are **5/5** (`make -C tb/verilator/rx_filter binding-negative`, missing and both direction-wrong bindings refused by name with `-Werror-PINMISSING -Werror-UNDRIVEN`); the self-test is 70/70. The RX filter's first-beat contract text was corrected to the law the RTL implements (the verdict follows the live lookup until the first beat is accepted) and pinned by a harness arm; the RTL's behaviour is unchanged.

> [!IMPORTANT]
> **Post-review correction (2026-08-28, head `27f55e44`).** The contract gate covers both processor submodules, now parses parameterised instances instead of silently skipping them, and inventories 92 open connections, 43 literal-bound connections, and 80 test-only hierarchical observations. The real RX-filter seam is explicit and its clean 62-check harness rejects a tied-off TCAM lookup; checker self-test 20/20. Earlier figures below are superseded.

Closes #253

Rule 5 of the #248 contract. **Stacked on #278 (Rule 4) → #277 → #276 → #275. Review in that order; this PR's diff is incremental.**

## What changed

| Area | Change |
|---|---|
| Guide | Rule 5 section: what a port contract must state, the two refusals versus the one ratchet, the documented boundary, review checklist. |
| Gate | `scripts/check_port_contracts.py` + `scripts/port_docs.budget`. |
| Cleanup | `hdl/ieee8021q/filtering/rx_mac_filter.sv` — all eighteen undocumented ports given their contract. |
| CI | The gate and its self-test run in `docs.yml`. |

## Two refusals, one ratchet

[...]

PR #278, opened 2026-08-27
[Code quality] Rule 4: put the unit in the name

Mister-M-alt:
> [!IMPORTANT]
> **Post-review correction (2026-08-28, head `ece93aad`).** The naming ratchet now covers both processor submodules and no longer treats ordinary prose such as “a second copy” as a seconds unit. Result: 3,491 ports, 408 raw matches, 255 reasoned exclusions, 153 candidates at the ratchet; self-test 11/11. Earlier figures below are superseded.

Closes #252

Rule 4 of the #248 contract. **Stacked on #277 (Rule 3) → #276 → #275. Review in that order; this PR's diff is incremental.**

## What changed

| Area | Change |
|---|---|
| Guide | Rule 4 section: the qualifier table (time, cycles, size, rate, clock domain, predicate, event), what a qualifier is **not** for, cross-language equivalence, review checklist. |
| Rename | The credit-based shaper configuration interface, end to end through 19 files. |
| Measurement | `scripts/measure_naming.py` + `scripts/naming.budget` — a ratchet, with the false positives measured first. |
| CI | The ratchet and its self-test run in `docs.yml`. |

## The interface, renamed end to end

Three quantities, three different units, none of them stated in a name:

| Was | Now | Unit, from its own `//!` comment |
|---|---|---|
| `o_cbs_hi_credit`, `cbs_hi_credit_i`, `hi_credit_i` | `…_hi_credit_bytes…` | signed bytes |
| `o_cbs_lo_credit`, `cbs_lo_credit_i`, `lo_credit_i` | `…_lo_credit_bytes…` | signed bytes |
| `o_cbs_idle_slope`, `cbs_idle_slope_i`, `idle_slope_i` | `…_idle_slope_bps…` | bits per second |

Every hop it crosses: `milan_csr` → `milan_datapath` / `milan_top` → `traffic_controller_802_1q` → `traffic_shaping_core` → `credit_based_shaper`, plus four testbench wrappers, their harnesses, the CBS reference model, and the register documentation. `hiCredit`/`loCredit` are 802.1Q terms and keep their spelling; only the unit is added.

The rename is pure — no logic moved. LiteX binds none of these names (they are internal to `milan_top`/`milan_datapath`, all past the port list), so the change is contained in RTL, testbenches and docs.

## The false positives were measured before anything was gated

This is the acceptance criterion the issue puts hardest, and the naive check fails it: **180** of 1,757 ports match a unit word in their comment, and most are not findings.

| Excluded | Count | Reason |
[...]

PR #277, opened 2026-08-27
[Code quality] Rule 3: make the RTL the authority for its own source lists

Mister-M-alt:
> [!IMPORTANT]
> **Post-review correction (2026-08-28, head `823bdc99`).** The source closure no longer depends on project naming prefixes: it follows every declared module instantiation plus package and interface/modport dependencies, including the processor submodules. The Vivado consumer now combines literal sources with the real `pp_srcs.py` expansion. Result: 108 files, 5/5 consumers complete, self-test 12/12. Earlier 61-file and 7-arm figures below are superseded.

Closes #251

Rule 3 of the #248 contract. **Stacked on #276 (Rule 2), which is stacked on #275 (Rule 1) — review in that order; this PR's diff is incremental.**

## What changed

| Area | Change |
|---|---|
| Guide | Rule 3 section in `docs/development/CODE_QUALITY.md`: the single-definition rule, the independent-oracle exception, the classified duplication inventory, worked example, review checklist. |
| Consolidation | `scripts/check_rtl_source_lists.py` — new gate. The RTL is the authority; all five source lists are derived consumers. |
| Emit paths | `syn/yosys/ooc.sh --emit-dp` and a `print-srcs` target in `tb/verilator/hostplane/Makefile`, so those two consumers can be *asked* instead of parsed. |
| Named constant | `milan_datapath` passed `16'h88F7` to `ptp_ts_top`'s `ETH_TYPE` while already importing the package that defines it as `ETH_TYPE_PTP`. |
| Completeness | `tb/verilator/milan_dp` now lists `event_counter.sv` explicitly instead of relying on a `+incdir+` side effect. |
| CI | The gate and its self-test run in `docs.yml`, beside the Vivado source-list gate. |

## The duplication, and the receipts

The set of files needed to compile `milan_datapath` was written out five times. Exactly one was guarded. The other four cost three broken Verilator suites the moment Rule 1 extracted a module — one failure per list nobody had told.

The gate walks the module closure of `milan_datapath` through the sources (transitively) and checks all five consumers carry every file in it: **61 files, 5 of 5 consumers.**

Two properties make it a source-of-truth gate and not a sixth copy:

- **It never checks one list against another.** All five could agree and all five be wrong. The first-hop instantiation parser is *imported* from `scripts/check_soc_sources.py` rather than re-written — a gate about single sources of truth that forked its own parser would refute itself.
[...]

PR #276, opened 2026-08-27
[Code quality] Rule 2: flatten the cluster namer into a role table

Mister-M-alt:
> [!IMPORTANT]
> **Post-review correction (2026-08-28, head `8e229639`).** The control-flow measurement now scans both project-owned processor submodules, recognizes Python `match` cases, and publishes measured dispositions for the processor hotspots. At this PR head: 2,322 host functions with 35 at depth 5 or more, and 616 RTL procedural blocks with 507 order-dependent; self-test 15/15. Earlier figures below predate this correction.

Closes #250

Rule 2 of the #248 contract. **Stacked on #275 (Rule 1) — review that first; this PR's diff is incremental.**

## What changed

| Area | Change |
|---|---|
| Guide | Rule 2 section in `docs/development/CODE_QUALITY.md`: wording, what "explicit" means for an RTL priority chain versus a host-side parser, worked example, exceptions, review checklist. |
| Measurement | `scripts/measure_control_flow.py` — nesting depth and decision points per host function; per procedural block in RTL, the nesting and the signals assigned more than once (priority decided by source order). |
| Cleanup | `cluster_names` in `sw/builder/endstation_builder.py`: a four-way `if/elif/elif/else` nested in two loops became one named function per pool role plus a dispatch table. |
| Tests | `test_d10_cluster_namer_paths` (gate 24d) — arms for the two paths the nesting hid. |

## The simplification

| Unit | Lines | Nesting depth | Decision points |
|---|---:|---:|---:|
| `cluster_names` before | 38 | 6 | 14 |
| `cluster_names` after | 20 | 2 | 4 |
| `_name_loopback` (deepest namer) | 9 | 1 | 1 |

## Behaviour is unchanged, and the claim is made with a tool

Every cluster name the builder can produce — 5 configurations × every stream port × both directions — is **byte-identical** before and after. The differential was captured on `origin/dev`, replayed after the refactor, and `diff` is empty.

## Two paths the nesting hid, now with arms

- The loopback namer's **empty receive channel space**: a talker whose entity declares no listener names the cluster by its own offset. That was a `continue` five levels deep and nothing graded it.
- The chain's **bare `else`**: it swallowed any role that was not physical, virtual or pilot and named it as a loopback channel. The table refuses an unknown role by name. The role set is closed today (`legacy_pool` and `role_pool` are the only producers), which is exactly why the refusal needs an arm.

[...]

PR #280, opened 2026-08-27
[Code quality] Rule 6: refuse the build the shield cannot filter

### LINKED ISSUE #254: [Code quality] Rule 6: Fail fast and encode invariants
Parent: #248

## Problem

A build, generator, test, or synthesis script that reports an error but exits
successfully creates false evidence. RTL with undocumented invalid states or
parameter ranges has the same maintainability problem at elaboration/runtime.

## Proposed rule: fail fast and encode invariants

> Invalid parameters, configurations, states, inputs, and tool failures MUST be
> rejected at the nearest responsible boundary and MUST propagate a non-success
> verdict. Important invariants and reset/overflow laws are executable
> assertions or self-checking tests, not comments alone.

Repository interpretation:

- Shell/Tcl/Python commands check return status and never print failure followed
  by exit 0.
- SystemVerilog rejects impossible parameter combinations at elaboration and
  makes illegal/default state behavior explicit.
- Counter wrap/saturation, reset, and update priority are stated and tested.
- An assertion has a useful message naming the violated contract.

Existing defects such as #245 are implementation instances of this rule, not a
reason to duplicate their narrower fixes here.

## Acceptance criteria

- [ ] Add fail-fast examples for RTL elaboration, FSM defaults, Python
      generators, shell pipelines, and Tcl synthesis flows.
- [ ] Inventory commands that can mask child-process failures and assertions
      that only log without controlling the verdict.
- [ ] Fix one representative masked failure and add a deliberately broken
      fixture/self-test proving the gate returns nonzero.
- [ ] Encode and mutation-prove one high-value RTL invariant.
- [ ] Ensure suite tally parsing treats missing/malformed evidence as failure,
      not zero checks.

## Non-goals

- No duplicate implementation work for already-scoped bug tickets.
- No assertion flood for states that are structurally impossible and already
  proven by a stronger boundary check.

Mister-M-alt:
> [!IMPORTANT]
> **Post-review correction (2026-08-29, head `e7e6995b`).** The "Evidence" figures below were measured at the round-1 head; after review round 2 the tool, the budget and the suite changed, and these are the figures at the current head:
> - `scripts/measure_fail_fast.py --selftest` — 90/90; `--check` — PASS (85 ≤ 85 modules without a module-scope elaboration contract, 4 ≤ 4 masked pipelines, 2 ≤ 2 discarded captured verdicts, 2 waived by site).
> - Population: 102 parameterised modules, 85 without a module-scope contract, 17 with; 34 `.sh`, 5 workflow files, 93 Makefiles (recipe lines); Python and Tcl are not measured, and the guide says so.
> - `tb/verilator/rx_filter` — 62 checks 0 failures, binding-negative 5/5, negative 5/5, each refusal graded by the contract's own `USERERROR` line (with the `$error` chain deleted all four arms print `[FAIL]`); `-GTDATA_WIDTH=52` in the build flags → `%Error-USERERROR`, rc 2; `tb/verilator/tcam_csr` 2 checks 0 failures with `-Werror-USERERROR`.
> - `scripts/suite_tally.py --selftest` PASS (a suite whose log tallies a failure while `make` exits 0 is refused by `scripts/run_all_suites.sh`); `ci_events.py --check` OK (230); `lint_rtl.py --check` PASS (99 ≤ 99); `docs_check` 0 / `gen_toc --check` OK / `check_doc_paths` OK.
> - Not enforced, and stated: the sv2v → Yosys path lowers a module-scope `$error` to a `$display` Yosys ignores (`OOC_CHPARAM="TDATA_WIDTH=52" syn/yosys/ooc.sh rx_mac_filter` still passes).

> [!IMPORTANT]
> **Post-review correction (2026-08-28, head `04d0fa5d`).** Fail-fast measurement remains per module across both processor submodules, pipe protection is order-sensitive, and the GitHub Actions default bash contract is modeled without violating bare-metal terminology. Result: 53 parameterised modules, 43 unguarded, zero masked pipelines, two named waivers; self-test 19/19. Earlier figures below are superseded.

Closes #254

Rule 6 of the #248 contract. **Stacked on #279 (Rule 5) → #278 → #277 → #276 → #275.**

## What changed

| Area | Change |
|---|---|
| Guide | Rule 6 section: repository interpretation, the comment-is-not-a-guard example, the two measured populations, review checklist. |
| Invariant | `hdl/ieee8021q/filtering/rx_mac_filter.sv` gains an elaboration contract in the house form. |
| Mutation proof | `make negative` in `tb/verilator/rx_filter` — permanent, four refusals plus one acceptance. |
| Measurement | `scripts/measure_fail_fast.py` + `scripts/fail_fast.budget` (two ratchets). |
| CI | The ratchet and its self-test run in `docs.yml`. |

## A comment is not a guard

`rx_mac_filter` is the receive shield — it decides which frames reach the host. Its banner said the destination compare holds "for TDATA_WIDTH>=48" and **nothing enforced it**. At 32 bits the concatenation building the destination address indexes past the end of the beat, the compare runs on undefined bits, and the shield silently admits or drops the wrong frames. No downstream check can see that: a filter that is wrong looks exactly like a filter that is right until the traffic is inspected on the wire.

Four laws are now executable, in the same module-scope `if (…) $error("one format string", …)` form `milan_datapath` and `KL_media_nco` already use: the datapath must hold the 48-bit destination address in one beat, it must be a whole number of bytes so `tkeep` can describe it, the TCAM needs at least one entry, and an action must distinguish two matches.

## Graded by mutation, permanently

```
[PASS] elaboration refused TDATA_WIDTH=32
[PASS] elaboration refused TDATA_WIDTH=52
[PASS] elaboration refused NUM_ENTRIES=0
[PASS] elaboration refused ACTION_WIDTH=0
[PASS] elaboration accepted the legal default shape

5 checks: 5 PASS, 0 FAIL
```

The last arm is what stops the contract being a ban rather than a guard. The arms read the exit status **directly, never through a pipe** — a pipeline returns its last command's status, which is exactly how a refused build reads as a pass.

## The two populations

| Population | Count | Disposition |
|---|---:|---|
| Parameterised modules with no elaboration contract | **37 of 42** | Ratchet, paid down as modules gain contracts |
| Pipelines that discard their producer's exit status | **0** | Ratchet at zero — it must stay zero |
| Pipelines waived because the consumer *is* the assertion | 27 | Named, with the reason recorded |

`verilator --version \| grep -F "$WANT"` wants grep's status, because grep is the assertion. Those are waived **by name**, not by pattern, so a new one must be added deliberately.

## Both false positives were in the checker, not the tree

On its first run the pipeline scan reported one masked pipeline. It was wrong twice over: a tool name inside a `printf` **format string** read as a command, and a pipe inside a `$(…)` substitution — whose status is never the line's verdict — read as masked. Quoted spans and substitutions are now blanked before the producer search, and both cases have arms. The true count is 0.

## Already satisfied — verified, not rebuilt

The issue also asks that suite tally parsing treat missing or malformed evidence as failure rather than zero checks. `scripts/suite_tally.py` already does this by design (`NOCOUNT`; a skip marker cannot suppress it). Confirmed by running its self-test — `selftest: PASS` — rather than writing a second one.

## Evidence

- `scripts/measure_fail_fast.py --selftest` — **16/16**.
- `scripts/measure_fail_fast.py --check` — PASS (37 ≤ 37, 0 ≤ 0, 27 waived).
- `tb/verilator/rx_filter` — 62 checks 0 failures, plus 5/5 negative arms.
- Direct exit codes: illegal `TDATA_WIDTH=32` → rc 1; illegal `NUM_ENTRIES=0` → rc 1; legal default → rc 0.
- `suite_tally.py --selftest` — PASS.
- `ci_events.py --check` OK (222 items); traceability matrix up to date.
- `docs_check` 0 findings / `gen_toc --check` OK / `check_doc_paths` OK.

## Not covered

- The other 37 modules' contracts; that is the ratchet's job.
- `tsn_fuzz` declared SKIP; builder gate 23h pre-existing on this host and on a pristine `dev` checkout.

### DIFF
--- a/scripts/measure_fail_fast.py
+++ b/scripts/measure_fail_fast.py
@@ -9,8 +9,9 @@ result must be rejected at the nearest responsible boundary and must propagate
 a non-success verdict. Two populations in this tree can break that, and they
 break it silently:
 
-  1. A PARAMETERISED MODULE WITH NO ELABORATION CONTRACT. Forty-three
-     first-party modules declare parameters; four reject an impossible
+  1. A PARAMETERISED MODULE WITH NO ELABORATION CONTRACT. Fifty-three
+     first-party modules across the superproject and its project-owned
+     processor submodules declare parameters; ten reject an impossible
      combination at elaboration. The rest accept any value the caller passes.
      `rx_mac_filter` documented "true for TDATA_WIDTH>=48" in a banner for
      months - a comment does not stop a build, and at 32 bits the destination
@@ -38,7 +39,6 @@ Exit 0 = at or under the ratchet in scripts/fail_fast.budget.
 
 import argparse
 import re
-import subprocess
 import sys
 from pathlib import Path
 
@@ -47,6 +47,7 @@ BUDGET = Path(__file__).resolve().parent / "fail_fast.budget"
 sys.path.insert(0, str(Path(__file__).resolve().parent))
 
 from lint_rtl import LINT_EXCLUDE
+from code_quality_scope import tracked
 
 COMMENT = re.compile(r"//[^\n]*|/\*.*?\*/", re.S)
 PARAM = re.compile(r"\bparameter\s+(?:int|logic|bit|integer|byte)?\s*"
@@ -72,18 +73,27 @@ def blank_comments(text):
                        text)
 
 
-def tracked(*pats):
-    out = subprocess.run(["git", "ls-files", *pats], cwd=REPO,
-                         capture_output=True, text=True, check=True).stdout.split()
-    return [p for p in out
-            if not p.startswith(("third_party/", "external/", "protocol-processor/",
-                                 "gptp-processor/", "gen/", "build/"))]
-
-
 def scan_module(text):
     """(declares_parameters, has_elaboration_check) for one SystemVerilog source."""
     code = blank_comments(text)
     return bool(PARAM.search(code)), bool(ELAB_CHECK.search(code))
+
+
+MODULE = re.compile(r"^\s*module\s+([A-Za-z_]\w*)\b", re.M)
+ENDMODULE = re.compile(r"\bendmodule\b")
+
+
+def scan_modules(text):
+    """Return `(name, has_parameters, has_check)` per module, never per file."""
+    code = blank_comments(text)
+    rows = []
+    for match in MODULE.finditer(code):
+        end = ENDMODULE.search(code, match.end())
+        if end is None:
+            continue
+        has_params, has_check = scan_module(code[match.start():end.end()])
+        rows.append((match.group(1), has_params, has_check))
+    return rows
 
 
 #: quoted spans and command substitutions, blanked before the producer search
@@ -107,31 +117,52 @@ def scan_pipeline(line):
     for needle, why in INTENTIONAL:
         if needle in line:
             return False, why
-    if "pipefail" in line:
-        return False, "the line sets pipefail"
     if line.lstrip().startswith("#"):
         return False, ""
     return bool(PIPE.match(shell_code(line))), ""
 
 
+def scan_pipelines(text, workflow=False):
+    """Return masked and waived pipeline rows, respecting activation order.
+
+    GitHub's default Linux bash shell starts every `run` script with
+    `-o pipefail`; ordinary `.sh` files must enable it before the pipeline.
+    A later mention of pipefail cannot retroactively protect an earlier line.
+    """
+    active = workflow
+    masked, waived = [], []
+    for n, line in enumerate(text.splitlines(), 1):
+        code = shell_code(line)
+        if re.search(r"\bset\s+\+o\s+pipefail\b", code):
+            active = False
+        elif re.search(r"\bset\b[^#\n]*\bpipefail\b", code):
+            active = True
+        is_masked, why = scan_pipeline(line)
+        if why:
+            waived.append((n, why))
+        elif is_masked and not active:
+            masked.append((n, line.strip()[:96]))
+    return masked, waived
+
+
 def audit():
     unguarded, guarded = [], []
     for rel in [p for p in tracked("hdl") if p.endswith(".sv") and p not in LINT_EXCLUDE]:
-        has_params, has_check = scan_module((REPO / rel).read_text(errors="replace"))
-        if not has_params:
-            continue
-        (guarded if has_check else unguarded).append(rel)
+        for name, has_params, has_check in scan_modules(
+                (REPO / rel).read_text(errors="replace")):
+            if not has_params:
+                continue
+            unit = f"{rel}:{name}"
+            (guarded if has_check else unguarded).append(unit)
 
     masked, waived = [], []
     for rel in tracked("*.sh", ".github/workflows/*.yml", "syn/**/*.sh", "harness/**/*.sh"):
         text = (REPO / rel).read_text(errors="replace")
-        pipefail_file = "pipefail" in text
-        for n, line in enumerate(text.splitlines(), 1):
-            is_masked, why = scan_pipeline(line)
-            if why:
-                waived.append((rel, n, why))
-            elif is_masked and not pipefail_file:
-                masked.append((rel, n, line.strip()[:96]))
+        found, exceptions = scan_pipelines(
+            text, workflow=rel.startswith(".github/workflows/") or
+            "/.github/workflows/" in rel)
+        masked.extend((rel, n, line) for n, line in found)
+        waived.extend((rel, n, why) for n, why in exceptions)
     return unguarded, guarded, masked, waived
 
 
@@ -164,6 +195,11 @@ def selftest():
        scan_module("module m #(parameter int W = 8)();\n // $error(\"no\");\n"
                    "endmodule") == (True, False),
        "comments are blanked before the search")
+    rows = scan_modules(
+        "module guarded #(parameter int A=1)(); if (!A) $error(\"A\"); endmodule\n"
+        "module bare #(parameter int B=1)(); endmodule")
+    ck("two modules in one file are measured independently",
+       rows == [("guarded", True, True), ("bare", True, False)], f"{rows}")
 
     ck("a piped gate is masked", scan_pipeline("  python3 scripts/x.py | tee log")[0])
     ck("a piped make is masked", scan_pipeline("  make run | tail -5")[0])
@@ -175,7 +211,12 @@ def selftest():
     ck("a pipe inside a command substitution is not the line's verdict",
        not scan_pipeline('  msg=$(grep ERROR log | head -1); python3 x.py')[0])
     ck("a line that sets pipefail is not masked",
-       not scan_pipeline("  set -o pipefail; python3 x.py | tee log")[0])
+       not scan_pipelines("set -o pipefail\npython3 x.py | tee log")[0])
+    ck("a later pipefail cannot protect an earlier pipeline",
+       scan_pipelines("python3 x.py | tee log\nset -o pipefail")[0] ==
+       [(1, "python3 x.py | tee log")])
+    ck("GitHub's default bash shell carries pipefail",
+       not scan_pipelines("python3 x.py | tee log", workflow=True)[0])
     ck("the version assertion is waived by name, with a reason",
        scan_pipeline('  verilator --version | grep -F "$WANT"') == (False,
        "the grep IS the version assertion; its status is the verdict"))
@@ -190,7 +231,7 @@ def selftest():
     ck("the waiver actually fires on the live tree", bool(waived),
        "no intentional pipeline was seen - the waiver is untested here")
 
-    n = 16
+    n = 19
     print(f"\n{n} checks: {n - failures} PASS, {failures} FAIL")
     return 1 if failures else 0
 
--- a/scripts/measure_fail_fast.py
+++ b/scripts/measure_fail_fast.py
@@ -125,7 +125,7 @@ def scan_pipeline(line):
 def scan_pipelines(text, workflow=False):
     """Return masked and waived pipeline rows, respecting activation order.
 
-    GitHub's default Linux bash shell starts every `run` script with
+    GitHub Actions starts its default bash shell for every `run` script with
     `-o pipefail`; ordinary `.sh` files must enable it before the pipeline.
     A later mention of pipefail cannot retroactively protect an earlier line.
     """
@@ -215,7 +215,7 @@ def selftest():
     ck("a later pipefail cannot protect an earlier pipeline",
        scan_pipelines("python3 x.py | tee log\nset -o pipefail")[0] ==
        [(1, "python3 x.py | tee log")])
-    ck("GitHub's default bash shell carries pipefail",
+    ck("the GitHub Actions default bash shell carries pipefail",
        not scan_pipelines("python3 x.py | tee log", workflow=True)[0])
     ck("the version assertion is waived by name, with a reason",
        scan_pipeline('  verilator --version | grep -F "$WANT"') == (False,
--- a/scripts/measure_fail_fast.py
+++ b/scripts/measure_fail_fast.py
@@ -21,20 +21,43 @@ break it silently:
   2. A PIPELINE THAT DISCARDS ITS PRODUCER'S EXIT CODE. In a shell without
      `pipefail`, `gate | tee log` exits with tee's status, so a failing gate
      reads as a pass. This is scored per-line and only for producers that ARE
-     a verdict - a tool run, a make, a gate script.
+     a verdict - a tool run, a make, a gate script - and, since review, a
+     leftmost `grep`/`cat` that READS A FILE: `grep -A2 "LUTs" util.rpt |
+     head` under `set -e` without pipefail exits 0 when the report does not
+     exist, which is how a synthesis flow whose tools produced nothing was
+     published as measured (gptp-processor/syn/ooc/run.sh, at the pin).
 
-Both are RATCHETS. Neither population can be emptied in one change without
-exactly the churn the governing rule forbids, and the second has legitimate
+  3. A CAPTURED VERDICT WHOSE STATUS IS DISCARDED. `out=$(verilator ...)`
+     followed by a grep of the TEXT decides the verdict from what the tool
+     printed, never from what it returned. Without errexit the assignment's
+     status goes nowhere; with `local`/`export` in front it is the builtin's
+     status even under errexit. A verilator that crashes silently, or is not
+     installed and shadowed by a stub, prints nothing, matches no `%Error`,
+     and reads as LINT OK for every top (protocol-processor/scripts/
+     lint_hdl.sh, at the pin). Only the assignment forms are modelled: an
+     argument-form `echo "k=$(tool)"` is a verdict the CONSUMER decides, and
+     the one such site here (elaborate.yml's scope publication) is fail-closed
+     by the CI contract, which treats an empty publication as RTL-relevant.
+
+All three are RATCHETS. None can be emptied in one change without exactly the
+churn the governing rule forbids, and the pipeline population has legitimate
 members: `verilator --version | grep -F "$WANT"` deliberately wants grep's
 status, because grep IS the assertion. Those are excluded by name, with the
-reason recorded.
+reason recorded. Debt that lives in a pinned processor submodule is counted
+here and fixed upstream; the pin bump lowers the ratchet.
+
+GITHUB'S DEFAULT `run` SHELL IS `bash -e {0}`: errexit, NO pipefail. Only a
+step that declares `shell: bash` gets `-eo pipefail`. The first version of
+this scan assumed every workflow step had pipefail and let one step's
+`set -o pipefail` protect the next; both are modelled per step now, and every
+piped step in this tree sets `set -euo pipefail` itself.
 
 Usage:
     python3 scripts/measure_fail_fast.py            # both inventories
     python3 scripts/measure_fail_fast.py --check    # the ratchet
     python3 scripts/measure_fail_fast.py --selftest # fixture arms
 
-Exit 0 = at or under the ratchet in scripts/fail_fast.budget.
+Exit 0 = at or under all three ratchets in scripts/fail_fast.budget.
 """
 
 import argparse
@@ -60,6 +83,18 @@ ELAB_CHECK = re.compile(r"\$fatal|\$error")
 PIPE = re.compile(
     r"^\s*(?P<lhs>.*?\b(?:python3?|make|verilator|yosys|sv2v|behave|pytest"
     r"|\./run\.sh|\./ooc\.sh|scripts/\S+\.(?:py|sh))\b[^|]*?)\|(?!\|)")
+
+#: A leftmost grep/cat that reads a FILE and feeds a consumer. The file is the
+#: flow's product, so its absence is the failure; grep reading stdin is a
+#: consumer (an assertion) and is not matched here. Matched against
+#: shell_code(), where quoted patterns are already blanked.
+FILE_PIPE = re.compile(r"^\s*(?:grep|egrep|fgrep|cat)\b(?P<args>[^|]*)\|(?!\|)")
+
+#: A captured verdict: `$(tool ...)` for the same producers as PIPE. Matched
+#: on the RAW logical line (shell_code() blanks substitutions on purpose).
+SUBST = re.compile(
+    r"\$\(\s*(?:python3?|make|verilator|yosys|sv2v|vivado|iverilog|behave|pytest"
+    r"|\./run\.sh|\./ooc\.sh|scripts/\S+\.(?:py|sh))\b")
 
 #: Pipelines where the CONSUMER is the assertion and its status is the point.
 #: Each is matched on the whole line, so a new one has to be added deliberately.
@@ -112,6 +147,16 @@ def shell_code(line):
     return _QUOTED.sub(lambda m: " " * len(m.group(0)), line)
 
 
+def _reads_a_file(args):
+    """True when a grep/cat argument list names at least one file operand.
+
+    Options are skipped; for grep the first bare token is the pattern (blank
+    when it was quoted), so a file is the second bare token; for cat any bare
+    token is a file."""
+    bare = [t for t in args.split() if not t.startswith("-")]
+    return len(bare) >= 1
+
+
 def scan_pipeline(line):
     """(is_masked, reason) for one shell/workflow line."""
     for needle, why in INTENTIONAL:
@@ -119,30 +164,140 @@ def scan_pipeline(line):
             return False, why
     if line.lstrip().startswith("#"):
         return False, ""
-    return bool(PIPE.match(shell_code(line))), ""
+    code = shell_code(line)
+    if PIPE.match(code):
+        return True, ""
+    m = FILE_PIPE.match(code)
+    if m:
+        args = m.group("args")
+        bare = [t for t in args.split() if not t.startswith("-")]
+        # grep: pattern then file(s) - a quoted pattern was blanked, so a bare
+        # file token alone means the pattern was quoted; cat: any bare token
+        is_cat = code.lstrip().startswith("cat")
+        quoted_pattern = bool(re.match(r"^\s*(?:grep|egrep|fgrep)\b[^|]*?[\"']", line))
+        needs = 1 if (is_cat or quoted_pattern) else 2
+        return len(bare) >= needs, ""
+    return False, ""
+
+
+_STEP_START = re.compile(r"^\s*-\s+\w+:")
+_SHELL_BASH = re.compile(r"^\s*shell:\s*bash\b")
+
+
+def workflow_steps(text):
+    """[(first_line_no, lines, bash_shell)] - one entry per workflow step.
+
+    Text before the first `- key:` is the preamble (one entry, no shell). A
+    step declares `shell: bash` before or after its `run:` block, so the whole
+    step is read before its lines are scanned."""
+    steps, cur, start = [], [], 1
+    for n, line in enumerate(text.splitlines(), 1):
+        if _STEP_START.match(line) and cur:
+            steps.append((start, cur))
+            cur, start = [], n
+        cur.append(line)
+    if cur:
+        steps.append((start, cur))
+    return [(start, lines, any(_SHELL_BASH.match(l) for l in lines))
+            for start, lines in steps]
+
+
+def _units(text, workflow):
+    """[(first_line_no, lines, pipefail0, errexit0)] - the scan units.
+
+    A `.sh` file is one unit that starts with neither option. A workflow is one
+    unit per step: GitHub's default shell is `bash -e {0}` (errexit, no
+    pipefail), and `shell: bash` is `bash --noprofile --norc -eo pipefail {0}`.
+    Options set inside one step never reach the next - each step is a fresh
+    shell."""
+    if not workflow:
+        return [(1, text.splitlines(), False, False)]
+    return [(start, lines, bash, True) for start, lines, bash in workflow_steps(text)]
 
 
 def scan_pipelines(text, workflow=False):
     """Return masked and waived pipeline rows, respecting activation order.
 
-    GitHub Actions starts its default bash shell for every `run` script with
-    `-o pipefail`; ordinary `.sh` files must enable it before the pipeline.
-    A later mention of pipefail cannot retroactively protect an earlier line.
+    Ordinary `.sh` files must enable pipefail before the pipeline; a later
+    mention cannot retroactively protect an earlier line. A workflow step
+    starts without pipefail unless it declares `shell: bash`.
     """
-    active = workflow
     masked, waived = [], []
-    for n, line in enumerate(text.splitlines(), 1):
-        code = shell_code(line)
-        if re.search(r"\bset\s+\+o\s+pipefail\b", code):
-            active = False
-        elif re.search(r"\bset\b[^#\n]*\bpipefail\b", code):
-            active = True
-        is_masked, why = scan_pipeline(line)
-        if why:
-            waived.append((n, why))
-        elif is_masked and not active:
-            masked.append((n, line.strip()[:96]))
+    for start, lines, active, _ in _units(text, workflow):
+        for n, line in enumerate(lines, start):
+            code = shell_code(line)
+            if re.search(r"\bset\s+\+o\s+pipefail\b", code):
+                active = False
+            elif re.search(r"\bset\b[^#\n]*\bpipefail\b", code):
+                active = True
+            is_masked, why = scan_pipeline(line)
+            if why:
+                waived.append((n, why))
+            elif is_masked and not active:
+                masked.append((n, line.strip()[:96]))
     return masked, waived
+
+
+_ERREXIT_ON = re.compile(r"\bset\s+(?:-[a-df-zA-Z]*e[a-zA-Z]*\b|-o\s+errexit\b|-[a-zA-Z]*\s+-o\s+errexit\b)")
+_ERREXIT_OFF = re.compile(r"\bset\s+(?:\+[a-zA-Z]*e[a-zA-Z]*\b|\+o\s+errexit\b)")
+_GUARD_PREFIX = re.compile(r"^\s*(?:if|elif|while|until)\b|^\s*!\s|^\s*\[|^\s*test\b")
+_BUILTIN_PREFIX = re.compile(r"^\s*(?:local|export|declare|readonly|typeset)\b")
+_ASSIGN_PREFIX = re.compile(r"^\s*[A-Za-z_]\w*\+?=")
+
+
+def logical_lines(lines, start):
+    """Join backslash-continued lines; yields (first_line_no, joined)."""
+    buf, first = [], None
+    for n, line in enumerate(lines, start):
+        if first is None:
+            first = n
+        if line.rstrip().endswith("\\"):
+            buf.append(line.rstrip()[:-1])
+            continue
+        buf.append(line)
+        yield first, " ".join(buf)
+        buf, first = [], None
+    if buf:
+        yield first, " ".join(buf)
+
+
+def scan_substitution(line, errexit):
+    """(is_masked, form) for one logical line holding `$(producer ...)`.
+
+    Forms: `local/export x=$(...)` is masked whatever errexit says - the status
+    is the builtin's. A bare assignment is masked when errexit is off and no
+    `||`/`&&` consults the status. A test (`if`, `!`, `[`) consults it. Any
+    other form is the consumer's verdict and is not modelled here."""
+    if line.lstrip().startswith("#"):
+        return False, "comment"
+    m = SUBST.search(line)
+    if not m:
+        return False, ""
+    if _BUILTIN_PREFIX.match(line):
+        return True, "builtin"
+    if _GUARD_PREFIX.match(line):
+        return False, "test"
+    if _ASSIGN_PREFIX.match(line):
+        tail = line[m.end():]
+        consulted = ("||" in tail) or ("&&" in tail)
+        return (not errexit) and not consulted, "assignment"
+    return False, "argument"
+
+
+def scan_substitutions(text, workflow=False):
+    """Masked captured-verdict rows [(line_no, text)], errexit in line order."""
+    masked = []
+    for start, lines, _, errexit in _units(text, workflow):
+        for n, line in logical_lines(lines, start):
+            code = shell_code(line.replace("$(", "  "))   # keep the $( visible below
+            if _ERREXIT_OFF.search(code):
+                errexit = False
+            elif _ERREXIT_ON.search(code):
+                errexit = True
+            is_masked, _form = scan_substitution(line, errexit)
+            if is_masked:
+                masked.append((n, line.strip()[:96]))
+    return masked
 
 
 def audit():
@@ -155,22 +310,22 @@ def audit():
             unit = f"{rel}:{name}"
             (guarded if has_check else unguarded).append(unit)
 
-    masked, waived = [], []
+    masked, waived, captured = [], [], []
     for rel in tracked("*.sh", ".github/workflows/*.yml", "syn/**/*.sh", "harness/**/*.sh"):
         text = (REPO / rel).read_text(errors="replace")
-        found, exceptions = scan_pipelines(
-            text, workflow=rel.startswith(".github/workflows/") or
-            "/.github/workflows/" in rel)
+        workflow = rel.startswith(".github/workflows/") or "/.github/workflows/" in rel
+        found, exceptions = scan_pipelines(text, workflow=workflow)
         masked.extend((rel, n, line) for n, line in found)
         waived.extend((rel, n, why) for n, why in exceptions)
-    return unguarded, guarded, masked, waived
+        captured.extend((rel, n, line) for n, line in scan_substitutions(text, workflow=workflow))
+    return unguarded, guarded, masked, waived, captured
 
 
 def read_budget():
     if not BUDGET.is_file():
         return None, None
     vals = [int(x) for x in re.findall(r"^\s*(\d+)\s*$", BUDGET.read_text(), re.M)]
-    return (vals + [None, None])[:2]
+    return (vals + [None, None, None])[:3]
 
 
 def selftest():
@@ -215,23 +370,71 @@ def selftest():
     ck("a later pipefail cannot protect an earlier pipeline",
        scan_pipelines("python3 x.py | tee log\nset -o pipefail")[0] ==
        [(1, "python3 x.py | tee log")])
-    ck("the GitHub Actions default bash shell carries pipefail",
-       not scan_pipelines("python3 x.py | tee log", workflow=True)[0])
+    ck("the GitHub Actions default run shell has NO pipefail",
+       scan_pipelines("    - name: s\n      run: |\n        python3 x.py | tee log\n",
+                      workflow=True)[0] == [(3, "python3 x.py | tee log")],
+       "GitHub's default is `bash -e {0}`; only `shell: bash` adds pipefail")
+    ck("a step that declares shell: bash is protected, before or after its run block",
+       not scan_pipelines("    - name: s\n      run: |\n        python3 x.py | tee log\n"
+                          "      shell: bash\n", workflow=True)[0])
+    ck("pipefail set in one step does not protect the next step",
+       scan_pipelines("    - run: |\n        set -euo pipefail\n        python3 a.py | tee a\n"
+                      "    - run: |\n        python3 b.py | tee b\n", workflow=True)[0]
+       == [(5, "python3 b.py | tee b")])
+    ck("a file-reading grep piped to head is a producer whose status is lost",
+       scan_pipelines('set -e\ngrep -A2 "Slice LUTs\\|DSPs" ucpu_util.rpt | head -20\n')[0]
+       == [(2, 'grep -A2 "Slice LUTs\\|DSPs" ucpu_util.rpt | head -20')],
+       "a missing report makes grep fail and head succeed: the flow exits 0")
+    ck("a file-reading cat piped to a consumer is the same shape",
+       scan_pipeline("cat engine_util.rpt | tail -3")[0])
+    ck("grep reading stdin is a consumer, not a producer",
+       not scan_pipeline('echo "$out" | grep -q FAIL')[0])
+    ck("a bare grep pattern with no file reads stdin and is not a producer",
+       not scan_pipeline("grep foo | head -1")[0])
+    ck("the file-reading grep is protected once pipefail is on",
+       not scan_pipelines("set -eo pipefail\ngrep x rpt | head\n")[0])
+    ck("a captured verdict with no errexit and no guard is discarded",
+       scan_substitutions("set -u\nout=$(verilator --lint-only -Wall \\\n  $pkgs $all 2>&1)\n"
+                          "if echo \"$out\" | grep -q Error; then rc=1; fi\n")
+       == [(2, "out=$(verilator --lint-only -Wall    $pkgs $all 2>&1)")],
+       "a silent non-zero verilator prints nothing, matches nothing, reads as OK")
+    ck("the same capture under errexit is not discarded",
+       not scan_substitutions("set -eu\nout=$(verilator --lint-only x.sv 2>&1)\n"))
+    ck("errexit switched off later un-protects a later capture",
+       scan_substitutions("set -e\nset +e\nout=$(make lint)\n") == [(3, "out=$(make lint)")])
+    ck("a local/export assignment discards the status even under errexit",
+       scan_substitutions("set -e\nf() {\n  local out=$(make lint)\n}\n") == [(3, "local out=$(make lint)")]
+       and scan_substitutions("set -e\nexport OUT=$(python3 x.py)\n") == [(2, "export OUT=$(python3 x.py)")],
+       "the builtin's status is what the shell sees")
+    ck("a capture whose status is consulted is not discarded",
+       not scan_substitutions("x=$(python3 y.py) || exit 2\n")
+       and not scan_substitutions("if ! x=$(python3 y.py); then exit 2; fi\n"))
+    ck("an argument-form substitution is the consumer's verdict and is not modelled",
+       not scan_substitutions('echo "rtl=$(python3 scripts/ci_scope.py)" >> "$OUT"\n'))
+    ck("a workflow step starts with errexit, so a bare capture there is not discarded",
+       not scan_substitutions("    - run: |\n        rtl=\"$(python3 scripts/ci_scope.py)\"\n",
+                              workflow=True))
+    ck("a capture inside a comment is not a finding",
+       not scan_substitutions("set -u\n# out=$(verilator x)\n"))
     ck("the version assertion is waived by name, with a reason",
        scan_pipeline('  verilator --version | grep -F "$WANT"') == (False,
        "the grep IS the version assertion; its status is the verdict"))
     ck("an ordinary command is not a pipeline",
        not scan_pipeline("  python3 scripts/x.py")[0])
 
-    unguarded, guarded, masked, waived = audit()
+    unguarded, guarded, masked, waived, captured = audit()
     ck("the live scan reads the tree", len(unguarded) + len(guarded) > 20,
        f"{len(unguarded)} + {len(guarded)} modules with parameters")
     ck("the guarded set is not empty", len(guarded) >= 4,
        "an inert scan would report every module unguarded")
     ck("the waiver actually fires on the live tree", bool(waived),
        "no intentional pipeline was seen - the waiver is untested here")
+    ck("the live scan reaches both processor submodules' shell wrappers",
+       any(rel.startswith("gptp-processor/") for rel, _, _ in masked + captured) and
+       any(rel.startswith("protocol-processor/") for rel, _, _ in masked + captured),
+       f"masked {[r for r, _, _ in masked]}, captured {[r for r, _, _ in captured]}")
 
-    n = 19
+    n = 36
     print(f"\n{n} checks: {n - failures} PASS, {failures} FAIL")
     return 1 if failures else 0
 
@@ -245,7 +448,7 @@ def main():
     if args.selftest:
         return selftest()
 
-    unguarded, guarded, masked, waived = audit()
+    unguarded, guarded, masked, waived, captured = audit()
 
     print(f"parameterised modules with no elaboration contract "
           f"({len(unguarded)} of {len(unguarded) + len(guarded)}):")
@@ -254,6 +457,9 @@ def main():
     print(f"\npipelines that discard their producer's exit code ({len(masked)}):")
     for rel, n, line in masked:
         print(f"   {rel}:{n}  {line}")
+    print(f"\ncaptured verdicts whose exit status is discarded ({len(captured)}):")
+    for rel, n, line in captured:
+        print(f"   {rel}:{n}  {line}")
     print(f"\nwaived, with a reason ({len(waived)}):")
     for rel, n, why in waived:
         print(f"   {rel}:{n}  {why}")
@@ -261,9 +467,9 @@ def main():
     if not args.check:
         return 0
 
-    b_unguarded, b_masked = read_budget()
-    if b_unguarded is None or b_masked is None:
-        print(f"\nNO RATCHET: {BUDGET.relative_to(REPO)} must hold two counts")
+    b_unguarded, b_masked, b_captured = read_budget()
+    if b_unguarded is None or b_masked is None or b_captured is None:
+        print(f"\nNO RATCHET: {BUDGET.relative_to(REPO)} must hold three counts")
         return 1
     bad = False
     if len(unguarded) > b_unguarded:
@@ -274,13 +480,19 @@ def main():
         print(f"\nFAIL: {len(masked)} masked pipeline(s) > ratchet {b_masked}. "
               f"A verdict must not be piped away.")
         bad = True
+    if len(captured) > b_captured:
+        print(f"\nFAIL: {len(captured)} captured verdict(s) discarded > ratchet "
+              f"{b_captured}. A tool's exit status is the verdict; its text is not.")
+        bad = True
     if bad:
         return 1
     print(f"\nFAIL-FAST RATCHET: PASS ({len(unguarded)} <= {b_unguarded} modules "
           f"without an elaboration contract, {len(masked)} <= {b_masked} masked "
-          f"pipeline(s), {len(waived)} waived with a reason)")
-    if len(unguarded) < b_unguarded or len(masked) < b_masked:
-        print(f"  the ratchets can be lowered to {len(unguarded)} and {len(masked)}")
+          f"pipeline(s), {len(captured)} <= {b_captured} discarded captured "
+          f"verdict(s), {len(waived)} waived with a reason)")
+    if len(unguarded) < b_unguarded or len(masked) < b_masked or len(captured) < b_captured:
+        print(f"  the ratchets can be lowered to {len(unguarded)}, {len(masked)} "
+              f"and {len(captured)}")
     return 0
 
 
--- a/scripts/measure_fail_fast.py
+++ b/scripts/measure_fail_fast.py
@@ -6,26 +6,48 @@
 Why this exists. Rule 6 of the maintainability guide
 (docs/development/CODE_QUALITY.md) says an invalid parameter, state or tool
 result must be rejected at the nearest responsible boundary and must propagate
-a non-success verdict. Two populations in this tree can break that, and they
+a non-success verdict. Three populations in this tree can break that, and they
 break it silently:
 
-  1. A PARAMETERISED MODULE WITH NO ELABORATION CONTRACT. Fifty-three
-     first-party modules across the superproject and its project-owned
-     processor submodules declare parameters; ten reject an impossible
-     combination at elaboration. The rest accept any value the caller passes.
+  1. A PARAMETERISED MODULE WITH NO ELABORATION CONTRACT. 102 first-party
+     modules across the superproject and its project-owned processor
+     submodules declare parameters; 18 reject an impossible combination at
+     elaboration. The rest accept any value the caller passes.
      `rx_mac_filter` documented "true for TDATA_WIDTH>=48" in a banner for
      months - a comment does not stop a build, and at 32 bits the destination
      compare would have read past the end of the beat and filtered on
      undefined bits, which nothing downstream can detect.
 
+     WHAT COUNTS AS A CONTRACT: `$error` or `$fatal` at MODULE OR GENERATE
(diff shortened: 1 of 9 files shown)


Mister-M-alt (commented):
[R1] MAJOR Conformance, Tests, Docs — processor lint/OOC wrappers — Rule 6 misses false-success paths in both project-owned processors

Requirement/evidence: Issue #254 requires an inventory of commands that can mask child-process failures, and Rule 6 says a tool failure MUST propagate a non-success verdict. The guide explicitly scopes both project-owned processor submodules, but `measure_fail_fast.py --check` reports zero masked pipelines while both pinned processors contain reproducible false-green paths:

1. `protocol-processor/scripts/lint_hdl.sh:14-20` captures Verilator output but discards the assignment command's exit status and decides only whether the text contains `%Warning` or `%Error`. At pin `3770ae02c56ca712d4a3505f429298b62edd5da8`, exporting a fake `verilator` that returns 7 with no output made the script print `LINT OK` for every discovered top and exit 0.
2. `gptp-processor/syn/ooc/run.sh:15-17` reads each expected report through `grep | head` under `set -e` but without `pipefail`. At pin `903a58125da89ba445c737e6d7db9ea5a8ba25f6`, successful fake Python/Vivado tools that produced no reports caused both `grep: ... No such file or directory` diagnostics, yet the wrapper exited 0.

The scanner cannot see the first form because command substitutions are blanked, and it does not recognize `grep` as a verdict-bearing producer in the second. Thus its zero result is not the inventory required by the Issue.

Impact: an unavailable/crashed lint tool or a missing synthesis report can be published as successful evidence. Green aggregate checks therefore do not prove the processor tooling follows Rule 6.

Required change: account for both failure shapes in the Rule 6 inventory and ensure the responsible wrappers propagate non-zero. Existing debt may be explicitly ratcheted under the governing cleanup policy, but it cannot remain absent from the measured population while the PR claims zero.

Verification: add permanent mutations for (a) a silent non-zero Verilator and (b) successful build commands with absent reports; each processor wrapper and the Rule 6 evidence must return non-zero. Re-run the real processor gates afterward. For control evidence, the real pinned gates were green in this audit: gPTP full `make`; AVB all 30 suites/14,722 checks, real Verilator lint, documentation checks, and Yosys.

Head reviewed: `04d0fa5dedf5db147640b3e3d602eb8c0224be57`.

Lenses applied in this focused round: Conformance, Tests, Docs. RTL and Robustness are not claimed as clean.

Mister-M-alt (commented):
**[R2] Cold-context review round 2** — reviewed at head `04d0fa5d` (the PR head before this round); every finding was reproduced by command and is answered by the commits on the restacked head `e7e6995b` (see the `[A2]` comment).

<details><summary>Full review (five lenses: Conformance / RTL / Robustness / Tests / Docs)</summary>

# R2 cold-context review — PR #280 "[Code quality] Rule 6: refuse the build the shield cannot filter"

Base `253-code-quality-rule-5-explicit-contracts` → head `254-code-quality-rule-6-fail-fast` @ `04d0fa5dedf5db147640b3e3d602eb8c0224be57`; submodules at pin (gptp-processor 903a5812, protocol-processor 3770ae02). Read in order: #254 + #248 acceptance criteria, AGENTS.md §6, CONTRIBUTING.md §3, the full diff (6 files, +463/−4), the three commits, then tests/logs. The R1 items (protocol-processor `lint_hdl.sh` discarding Verilator's status; gptp-processor `syn/ooc/run.sh` `grep | head`) are NOT re-reported below.

---

[R2] BLOCKER Conformance, Tests, Docs — scripts/measure_fail_fast.py:52-54 (`PARAM` regex); scripts/fail_fast.budget:7; docs/development/CODE_QUALITY.md:815 — The module ratchet cannot see the project's own parameter idiom: 49 of 102 parameterised modules are invisible, including every gptp-processor module
Requirement/evidence: #254 "SystemVerilog rejects impossible parameter combinations at elaboration" is measured per module; the guide says the ratchet is "across the superproject and both project-owned processor submodules". `PARAM = \bparameter\s+(?:int|logic|bit|integer|byte)?\s*(?:signed\s+)?(?:\[[^\]]*\]\s*)?(\w+)\s*=` does not match `parameter int unsigned X =`, `parameter string X =`, `parameter type T =`, or a parameter without a default. Hand count (scratchpad/miss.py, same population/comment-blanking as the tool, any `\bparameter\b` per module):
```
counted (has_params) = 53
MISSED by PARAM regex (module has 'parameter' keyword): 49
  gptp-processor/hdl/common/KL_gptp_timer.sv:KL_gptp_timer  check=False   parameter int unsigned CLK_HZ_P = 100_000_000,
  gptp-processor/hdl/top/KL_gptp_engine.sv:KL_gptp_engine   check=True    parameter string       UCODE_HEX_P = "",
  gptp-processor/hdl/ucpu/KL_gptp_ucpu.sv:KL_gptp_ucpu      check=False   parameter string UCODE_HEX_P = ""
  gptp-processor/hdl/wire/KL_gptp_tx_slot.sv:KL_gptp_tx_slot check=False  parameter int unsigned SLOT_BYTES_P = TXSLOT_BYTES_C
  hdl/ieee1722/crf/KL_media_nco.sv:KL_media_nco             check=True    parameter int unsigned CLK_FREQ_HZ_P = 100_000_000,
  ... (22 hdl/, 23 protocol-processor/, 4 gptp-processor/)
```
True population: 102 parameterised modules, 82 without a `$error|$fatal`, 20 with — not "43 of 53". Per processor at the pin: gptp-processor tool=0/0, true=4 parameterised (3 with no check at all); protocol-processor tool=8 unguarded+5 guarded, true=28+8. `KL_media_nco`, which the guide (CODE_QUALITY.md:791) and the RTL comment cite as the house form, is not in the tool's population at all. Direct probe: `m.scan_module('module m #(parameter int unsigned W=8)(); if (W<2) $error("no"); endmodule')` → `(False, True)`; `parameter string` → `(False, False)`; `parameter type` → `(False, False)`; `parameter int W` (no default) → `(False, False)`. Coverage mutation (a): appending `module zz_probe #(parameter int W = 1)(...)` to tracked `gptp-processor/hdl/common/KL_gptp_timer.sv` → `FAIL: 44 module(s) ... > ratchet 43`, rc=1 (same for `protocol-processor/hdl/srp/KL_srp_vlan.sv`); appending the same module with `parameter int unsigned W = 1` (gptp) or `parameter string HEX_P = ""` (pp) → `FAIL-FAST RATCHET: PASS (43 <= 43 ...)`, rc=0, `zz_probe listed? 0`. Submodules restored (`git -C <sub> checkout -- .`, status clean).
Impact: a new module written in the house `_P`/`int unsigned` idiom (Rule 4) never enters the ratchet, so the gate is satisfied vacuously by exactly the code it exists for; the docstring ("Fifty-three ... ten reject"), the guide table (43 of 53), the budget (43) and the PR body's "both processor submodules" are all false — gptp-processor's debt is hidden as zero.
Required change: `PARAM` must accept the SV parameter grammar this tree uses (`int unsigned`, `string`, `type`, `real`, `longint`, no-default), with a self-test arm per form; recount; re-base `fail_fast.budget` to the measured number (82 at this head); correct the docstring and the guide table; the printed inventory must list `KL_gptp_timer`, `KL_gptp_ucpu`, `KL_gptp_tx_slot`.
Verification: `python3 scripts/measure_fail_fast.py` lists 102 parameterised / 82 unguarded at this head with the four gptp modules present; the `int unsigned` and `string` injections above must FAIL `--check`; `--selftest` arms for each form must exist and must fail when the regex is reverted.

[R2] MAJOR RTL, Robustness, Conformance — hdl/ieee8021q/filtering/rx_mac_filter.sv:126-137; tb/verilator/rx_filter/Makefile:13-16; syn/yosys/run.sh:364-376; syn/yosys/ooc.sh:479-551 — The contract refuses the build only under `verilator --lint-only` without `-Wno-fatal`; every flow that actually builds or synthesises the shield accepts the illegal parameters
Requirement/evidence: the PR title/guide (CODE_QUALITY.md:781-799) claim the build is refused. (1) Verilator with `-Wno-fatal` — carried by every consumer of `rx_mac_filter` (`tb/verilator/rx_filter/Makefile` VFLAGS, `milan_dp`, `hostplane`, `tcam_csr`): `verilator --lint-only ... -Wno-fatal -GTDATA_WIDTH=52` → `%Warning-USERERROR: ... TDATA_WIDTH=52 is not a whole number of bytes ...`, rc=0; the suite's own VFLAGS with `--cc --exe --build ... -GTDATA_WIDTH=52` → rc=0 and `obj_dir_mut/Vmut` exists: `BINARY BUILT despite illegal TDATA_WIDTH=52`. (2) sv2v→Yosys (the only path `run.sh`/`ooc.sh` use): `sv2v --top=rx_mac_filter tcam.sv rx_mac_filter.sv` emits 0 `$error`; a minimal probe shows sv2v v0.0.13 lowers a module-scope `$error` to `initial $display("Error [elaboration] ... msg: ", ...)` and `initial ... $fatal` to `$display ... $finish(1)` — simulation prints Yosys ignores. `yosys -p "read_verilog rx.v; chparam -set TDATA_WIDTH 52 rx_mac_filter; hierarchy -check -top rx_mac_filter; synth_xilinx -flatten; stat"` → rc=0, 3616 cells (64 → 3643, 32 → 2712, so chparam re-derives the module); `NUM_ENTRIES 0` and `ACTION_WIDTH 0` → rc=0; even an illegal DEFAULT (`TDATA_WIDTH = 52` edited into a copy) → sv2v rc=0, Yosys rc=0. Real flows: `OOC_CHPARAM="TDATA_WIDTH=52" syn/yosys/ooc.sh rx_mac_filter` → rc=0 (PASS row); control `syn/yosys/run.sh --mode elaborate --top rx_mac_filter` legal default → rc=0. Yosys does enforce it only when fed the original SV (`read_verilog -sv ...; chparam -set TDATA_WIDTH 52` → `ERROR: rx_mac_filter: TDATA_WIDTH=%0d is not ...`, rc=1), which no repository flow does. Verilator without `-Wno-fatal`: all four refused, rc=1 (`make negative` 5/5). xvlog/xelab absent on this host.
Impact: an illegal shield shape builds and simulates in every suite and synthesises in the Yosys gate; the refusal exists only in the negative arm's own invocation. Assignment item 1 answer: enforced by Verilator lint-only; NOT enforced on the sv2v/Yosys path; the OOC flow can override parameters (`OOC_CHPARAM`), `run.sh` cannot; the legal default elaborates in both.
Required change: either the consumer flows must fail on an elaboration `$error` (the suite Makefiles cannot keep `-Wno-fatal` blanket for USERERROR; the Yosys/OOC gate needs its own shape check, as `protocol-processor/tb/timer_map/shape_elab.sh` already does for `protocol_processor_top`), or the guide must state that the contract is enforced by Verilator lint-only and Vivado elaboration and is inert under `-Wno-fatal` and on the sv2v path.
Verification: `make -C tb/verilator/rx_filter run` with `-GTDATA_WIDTH=52` added to VFLAGS must fail; `OOC_CHPARAM="TDATA_WIDTH=52" syn/yosys/ooc.sh rx_mac_filter` must fail or the guide sentence must say why it cannot.

[R2] MAJOR Tests, Docs — tb/verilator/rx_filter/Makefile:56-75 (`negative`); CODE_QUALITY.md:795-799; PR body "four refusals" — Three of the four refusal arms pass with the contract deleted; only `TDATA_WIDTH=52` proves the `$error`
Requirement/evidence: "each new test can fail for the defect it claims to detect." Direct lint of each illegal set at head shows what refuses it: `-GTDATA_WIDTH=32` → USERERROR **and** `%Warning-SELRANGE ... 39:32 outside 31:0` ×2; `-GNUM_ENTRIES=0` → USERERROR and `%Warning-ASCRANGE tcam.sv:57 ... [-1:0]` ×2; `-GACTION_WIDTH=0` → `%Warning-ASCRANGE` ×4, `%Error-ZEROREPL tcam.sv:100`, WIDTHEXPAND; `-GTDATA_WIDTH=52` → USERERROR only. Mutation: deleted lines 126-137 (the whole if/else-if chain; `grep -c '$error(' → 0`), then `make -C tb/verilator/rx_filter negative`:
```
[PASS] elaboration refused TDATA_WIDTH=32
[FAIL] elaboration ACCEPTED illegal TDATA_WIDTH=52
[PASS] elaboration refused NUM_ENTRIES=0
[PASS] elaboration refused ACTION_WIDTH=0
[PASS] elaboration accepted the legal default shape
5 checks: 4 PASS, 1 FAIL   rc=2
```
File restored (`git status` clean). The arms read only the exit status, so they cannot tell the contract's diagnostic from an accidental warning in `tcam.sv`.
Impact: the headline law (TDATA_WIDTH < 48, the one the whole narrative is about) is not mutation-proven; a future `-Wno-SELRANGE`/`-Wno-ASCRANGE` or a tcam.sv change silently turns three arms into no-ops while they keep printing `[PASS]`. The docs and PR body overstate "graded by mutation, permanently".
Required change: each arm must require the module's own diagnostic (`%Warning-USERERROR` carrying `rx_mac_filter: <PARAM>=`) — capture stderr to a file and let `grep -q` on the file be the verdict (no pipe; grep's status IS the assertion), the shape `protocol-processor/tb/timer_map/shape_elab.sh:50` already uses (`grep -q 'F08.4: owner tags OVERLAP'`).
Verification: with the chain deleted all four arms must print `[FAIL]`; with the chain restored 5/5.

[R2] MAJOR Robustness, Tests, Docs — scripts/measure_fail_fast.py:119-139 (`scan_pipelines(..., workflow=True)` starts `active=True`), :220-221 (self-test arm "the GitHub Actions default bash shell carries pipefail"); CODE_QUALITY.md:822-825 — The workflow model assumes `-o pipefail` that GitHub only applies with an explicit `shell: bash`; no workflow in this repository or either processor declares one
Requirement/evidence: GitHub's workflow-syntax reference (fetched 2026-08-28): unspecified shell → "The default shell on non-Windows platforms. Note that this runs a different command to when `bash` is specified explicitly" → `bash -e {0}`; explicit `shell: bash` → `bash --noprofile --norc -eo pipefail {0}`. Grep over `.github/workflows/*.yml`, `gptp-processor/.github` (absent) and `protocol-processor/.github/workflows/hdl.yml`: `shell:` = 0, `defaults:` = 0 in every file; the authors therefore hand-write `set -euo pipefail` at the top of 15 `run:` blocks (docs.yml:40, rtl.yml:51,75,87,...). Shell fact: `bash -e -c 'false | tee /dev/null'` → rc=0; `bash -eo pipefail -c '...'` → rc=1. Mutation: inserted `python3 scripts/measure_naming.py --check | tee ratchet.log` into docs.yml's own "Fail-fast ratchet" `run:` block (line 110) → `FAIL-FAST RATCHET: PASS (... 0 <= 0 masked pipeline(s) ...)`, rc=0. The model also never resets at a `run:` boundary (each step is a fresh shell): `scan_pipelines("run: |\n  set -o pipefail\n  true\nrun: |\n  python3 gate.py | tee log")` → `masked=[]`. Workflow restored.
Impact: any piped gate added to CI is reported safe while `bash -e {0}` masks it — the exact false-green Rule 6 exists to stop; the self-test arm pins the wrong premise; the guide sentence is false.
Required change: per `run:` block, initial state = pipefail iff the step/job/workflow declares `shell: bash` (or the block sets it); reset at each `run:`; invert/replace the self-test arm; correct CODE_QUALITY.md:822-823.
Verification: the docs.yml injection above must FAIL `--check`; the same line under a step with `shell: bash` must pass; the per-step reset arm must pass.

[R2] MAJOR Conformance, Tests, Docs — scripts/measure_fail_fast.py:61-64 (`INTENTIONAL`), :107-110 (`scan_pipeline` waives before any other test); CODE_QUALITY.md:819-821; PR body "waived by name, not by pattern" — The waiver is a substring match on the whole line, so any new `--version | grep` anywhere — including a gate verdict — is auto-waived
Requirement/evidence: `INTENTIONAL = (("--version | grep", ...),)` is tested with `needle in line` before the comment check and before the producer search. Mutation: appended to tracked `scripts/run_all_suites.sh`:
```
yosys --version | grep -q "0.0"
python3 scripts/lint_rtl.py --check --version | grep -c FAIL
```
→ tool output `scripts/run_all_suites.sh:301 the grep IS the version assertion...`, `:302 the grep IS the version assertion...`, `FAIL-FAST RATCHET: PASS (... 4 waived ...)`. File restored. The second line is a lint-gate verdict piped into `grep -c`, which is precisely the masked shape the ratchet is meant to hold at zero.
Impact: the "must be added deliberately" property claimed by the guide and PR body does not exist; the zero-ratchet can be bypassed by adding `--version` to any producer's argument list.
Required change: waive by site (`path` + exact line text, or `path:line`) as the guide describes, and keep the producer test ahead of the waiver.
Verification: the two lines above must be reported as masked (or at least not waived); the two real sites (`rtl-fast.yml:99`, `rtl.yml:184`) must remain waived by name.

[R2] MAJOR Tests, Conformance — scripts/measure_fail_fast.py:55 (`ELAB_CHECK = \$fatal|\$error` anywhere in the module) — A runtime `$error` counts as an elaboration contract, so the module ratchet can be left vacuously
Requirement/evidence: `m.scan_module('module m #(parameter int W=8)(input logic clk); always @(posedge clk) if (W) $error("runtime"); endmodule')` → `(True, True)` ("guarded"). Real instances: `gptp-processor/hdl/top/KL_gptp_engine.sv:735-740` has only runtime PathTrace `$error`s in an `always` block and no contract on `UCODE_HEX_P`/`CLK_HZ_P` — once the BLOCKER is fixed it would be counted guarded; `hdl/ieee1722/avtp/KL_avtp_common_parser.sv:52-53` (counted guarded today) guards via `initial begin assert (PIPELINE_DELAY >= 1) else $error(...)`, a time-0 simulation check Verilator only evaluates with `--assert` and sv2v/Yosys drop. The other nine counted guards are true module-scope generate checks (KL_pcm_tx:185, KL_tdm_capture_master:119, milan_datapath:818-871, KL_aecp_desc_store:237-245, KL_aecp_engine:808-814, KL_aecp_resp_buf:193-201, KL_pp_originator:148, protocol_processor_top:666-684, rx_mac_filter:126-137). Assignment item 3: `initial begin if (...) $fatal(...) end` (KL_pp_acmp_listener:346) and `$error` with a non-literal format are recognised; `localparam`-only modules are correctly not counted; runtime `$error` is wrongly recognised.
Impact: adding any runtime assertion to a parameterised module removes it from the unguarded list and the ratchet offers to lower itself; "has an elaboration contract" is not what is measured.
Required change: recognise only module/generate-scope `$error|$fatal` (outside `always*`/`initial`/`function`/`task`), or state in the guide and docstring that any `$error` counts and why; add a self-test arm for the runtime shape.
Verification: the `always @(posedge clk) ... $error` probe must return `(True, False)`; the counted-guarded list must not change for the nine true guards.

[R2] MAJOR Conformance, Docs — docs/development/CODE_QUALITY.md:760-855 (Rule 6 section); scripts/suite_tally.py:180-215 + scripts/run_all_suites.sh:266-276,285,297-300 — Three #254 acceptance criteria are claimed closed but not delivered at this head
Requirement/evidence: (1) "Add fail-fast examples for RTL elaboration, FSM defaults, Python generators, shell pipelines, and Tcl synthesis flows" — the section contains the RTL elaboration example (rx_mac_filter) and the shell pipeline example (`tee`/`pipefail`, `--version | grep`) only; there is no FSM-default example, no Python-generator example and no Tcl-synthesis example (the "Repository interpretation" bullets name Tcl/Python without an example; the whole section text is in the diff). (2) "Inventory commands that can mask child-process failures **and assertions that only log without controlling the verdict**" — only the first half is measured; the second class is neither inventoried nor gated: a suite log carrying `[FAIL] elaboration ACCEPTED illegal TDATA_WIDTH=52` beside `5 checks: 5 PASS, 0 FAIL` → `suite_tally.py` prints `in-suite failures: 0`, rc=0; a log whose tally says `5 checks: 4 PASS, 1 FAIL` → `in-suite failures: 1`, rc=0; `run_all_suites.sh` gates only on `make` rc (:266-276) and on `tally_rc` (:299, exit 90), never on `[FAIL]` markers or tally failures, so a harness that prints FAIL and exits 0 passes the sweep. (3) "Fix one representative masked failure" — the diff fixes none (the PR reports zero found; the two real ones are the R1 items being fixed outside this PR). Criteria 4 (invariant + mutation) and 5 (missing/malformed tally → failure: truncated `62 checks: 62 PASS, 0 FA` → `UNPARSED`, rc=1; log with no tally → `NOCOUNT`, rc=1; `--selftest` PASS) are delivered, subject to the two Tests findings above.
Impact: "Closes #254" is not true at this head; the log-only-assertion class is the one the rx_filter `negative` target itself would fall into if its `test $$fail -eq 0` were ever dropped.
Required change: add the three missing example classes (a `default:`/illegal-state arm, a generator that raises on bad input, a Tcl flow that checks `PROGRESS`/`exec` status — `bd/build.tcl:76-93` and `syn/ooc/milan_datapath_ooc.tcl:167` are usable in-tree examples); inventory or gate log-only assertions (at minimum make `run_all_suites.sh` fail when a log's tally reports failures or carries `[FAIL]` while the suite exited 0, or record why not); either fix a representative masked failure in this PR or drop the criterion from the close claim.
Verification: the section names all five example classes with a repository example each; `run_all_suites.sh` (or `suite_tally.py`) exits non-zero on the two contradictory logs above; the criterion checklist in #254 matches the diff.

[R2] MINOR Robustness, Tests — scripts/measure_fail_fast.py:76-139 (`shell_code`, `PIPE`, `scan_pipelines`) — Latent scanner misses reproduced on synthetic lines (none present in the tree today)
Requirement/evidence: `scan_pipelines(...)` results: `"# note: set -o pipefail is not used here\npython3 gate.py | tee log"` → `masked=[]` (a comment activates protection: `shell_code` blanks quotes but not `#` comments, and the `set ... pipefail` regex runs before `scan_pipeline`'s comment test); `"cat <<EOF\nset -o pipefail\nEOF\npython3 gate.py | tee log"` → `[]`; `"(set -o pipefail; make | tee a)\nmake | tee b"` → `[]` (subshell scope leaks); `"make -C x \\\n  | tee log"` → `[]` (continuation line); `"$MAKE -C x | tee log"`, `"./obj_dir/Vsim | tee log"`, `"./lint_hdl.sh | tee log"` (only `./run.sh`/`./ooc.sh` are listed), `"xvlog ... | tee"`, `"vivado ... | tee"`, `"iverilog ... | tee"` → `[]`; false positive: `"#!/bin/bash -eo pipefail\nmake | tee log"` → masked. Correct: `python3 gate.py 2>&1 | tee` masked; `if ! make | tee log; then` masked; `bash scripts/x.sh | tee` masked; `$(...)` and quoted spans blanked. Tree grep: no shebang/subshell/comment pipefail shapes and no such producers piped exist today (`syn/yosys/check_tap_purity.sh:94,109` and `scripts/check_tied_inputs.sh:46-57` continuation pipes are grep/awk text chains inside `$(...)`).
Impact: the zero can be preserved by accident (a comment) or missed by a shape the repo's Makefiles already use (`./obj_dir/V*`, `verilator_coverage`).
Required change: blank comments before the activation search; scope activation to the line's shell (subshell/step); join `\`-continued lines; extend producers to sim binaries, `$MAKE`, `*.sh` invocations and the Vivado/xvlog tools; add arms.
Verification: each snippet above reports the expected masked list in `--selftest`.

[R2] MINOR Conformance, Docs — scripts/measure_fail_fast.py:143 (population `*.sh`, workflows, `syn/**/*.sh`, `harness/**/*.sh`); CODE_QUALITY.md:808-818 — The pipeline population excludes 93 Makefiles (36 in the processors), 161 `.py` and 14 `.tcl` files, and the guide's scope statement does not say so
Requirement/evidence: `git ls-files --recurse-submodules` (first-party, excluding third_party/external/historical/docs-overhaul): Makefiles/.mk 93 (processors 36), .py 161 (11), .tcl 14 (4), .sh 34 (7), workflows 5 (1); no extension-less shell scripts. Makefile recipes run under `/bin/sh -c` per line with no `SHELL`/`.SHELLFLAGS`/`.POSIX` anywhere (grep 0), so a recipe pipe is masked; the only recipe pipes in the tree are `verilator_coverage --annotate ... \ | tail -2` at `tb/verilator/avtp_rxmon/Makefile:56-57` and `tb/verilator/maap/Makefile:27-28` (`coverage` targets, not in CI), backstopped by `cov_gate.py` (`/nonexistent_dir` and an empty dir → `MISSING ... FAIL`, rc=1). No `-`-prefixed recipes, `.IGNORE` or `|| true` in any Makefile. Python: `os.system` 0, `shell=True` 0; `check=False` sites are in `scripts/act_ci.py` (returncode kept), `bd/bringup.py:128` (returned through `raise SystemExit(main())`), and a `make clean` in `protocol-processor/tb/nvm_port/measure_figures.py:605`. Tcl: `exec` raises on non-zero (`syn/ooc/milan_datapath_ooc.tcl:42,167`, `pp_shadow_ooc.tcl:61`); `bd/build.tcl:78,93` check `PROGRESS` after `wait_on_run`; the `catch` sites are Vivado-generated `bd/milan-dma.tcl`. No unbackstopped masked path was found outside the scanned population at this head.
Impact: the guide reads as if "pipelines that discard their producer's exit status" were measured tree-wide; they are measured in `.sh` and workflow files only.
Required change: state the measured population in the guide and docstring, or extend the scan to Makefile recipes (`\t`-lines, `\`-joined) where the repo's sim binaries actually run.
Verification: the guide's scope sentence names the file classes scanned; if extended, `tb/verilator/maap/Makefile:27-28` appears in the inventory or is waived by site.

[R2] MINOR Docs — PR #280 body "Evidence" — Figures that do not match this head
Requirement/evidence: "`ci_events.py --check` OK (222 items)" → at head: `ci_events: OK (230 contract item(s) across 4 workflow files and docs/testing/CI_WORKFLOWS.md)`, rc=0. The superseded "37 of 42 / 27 waived / 16/16" are correctly marked superseded; the corrected "53 / 43 / 0 / 2 / 19" match the tool's output at head but are themselves wrong per the BLOCKER (true 102/82/20). Re-measured and matching: `--selftest` 19/19 (19 `ck()` calls = literal `n = 19`); `make -C tb/verilator/rx_filter` rc=0 with `checks: 62 failures: 0`, `1 checks: 1 PASS, 0 FAIL`, `5 checks: 5 PASS, 0 FAIL` (tally 68/0, rc=0); `docs_check: 0 finding(s)`; `TOC gate: OK`; `doc path gate: OK`; `suite_tally --selftest` PASS; `lint_rtl.py --check` PASS (99 ≤ 99).
Impact: a cold reader re-measuring the body finds a mismatch and cannot tell which figures were re-run after the correction.
Required change: restate the evidence block at the corrected head, or drop the numbers that were not re-run.
Verification: every figure in the body reproduces at the head the body names.

[R2] SUGGESTION RTL — hdl/ieee8021q/filtering/rx_mac_filter.sv:126-137 — Unnamed generate blocks add four `GENUNNAMED` warnings to every `-Wall` build of the shield
Requirement/evidence: rx_filter build log: `%Warning-GENUNNAMED ... rx_mac_filter.sv:127`, `:130`, `:133`, `:136` (4 new; not counted by the lint ratchet, which passes 99 ≤ 99). `KL_media_nco.sv:128,132` names its guard blocks (`gen_guard_hi`/`gen_guard_lo`); `milan_datapath.sv:818-871` does not — the "house form" the guide cites is two forms. Also the `else if` chain reports only the first violated law of a set that violates several.
Required change: name the blocks (`begin : gen_guard_width ... end`) and say which form is the house form.
Verification: the rx_filter build log carries no GENUNNAMED for rx_mac_filter.sv.

[R2] SUGGESTION Robustness — scripts/code_quality_scope.py:32; scripts/measure_fail_fast.py:232-234 — Refusals and tallies that are correct but rough
Requirement/evidence: off-pin gptp-processor (`git -C gptp-processor checkout --detach HEAD~1`, `git submodule status` → `+5d2c0f31...`) → `--check` rc=1 with a Python traceback ending `RuntimeError: project submodules must be initialized at their pinned commits before a code-quality scan (+5d2c0f31... gptp-processor ...)` — the refusal is right, the surface is a traceback rather than the named-refusal shape `xvlog_gate.py` uses; pin restored (` 903a58125...`). Budget missing or holding one number → `NO RATCHET: scripts/fail_fast.budget must hold two counts`, rc=1 (good). `selftest()` prints a literal `n = 19` rather than counting its arms, so an added arm without bumping `n` makes the `N checks:` line — the one `suite_tally.py` sums — lie.
Required change: catch the scope error and print the message only; derive `n` from the `ck()` calls.
Verification: off-pin run prints one line and exits 2/1; adding an arm changes the printed total without editing a literal.

---

## Lens coverage (no lens is clean at this head; what each examined)

- `Conformance` — #254's five criteria vs the diff (BLOCKER, MAJOR ×3, MINOR); processor items (a) coverage mutations in `gptp-processor/hdl/common/KL_gptp_timer.sv`, `protocol-processor/hdl/srp/KL_srp_vlan.sv`, `gptp-processor/syn/ooc/run.sh` (→ `FAIL: 1 masked pipeline(s) > ratchet 0`), `protocol-processor/scripts/lint_hdl.sh` (→ same) — the scan reads processor `.sh` and `.sv` files; (b) per-processor inventory at the pin: gptp 0/0 reported vs 4 parameterised (3 with no check) real, pp 8+5 reported vs 28+8 real, pipelines 0 masked / 0 waived in both (7 processor `.sh` files scanned); (c) processor tooling: `protocol-processor/scripts/run_suites.sh` (explicit `if !` on every step, `UNREADABLE` → exit 90) and `tb/timer_map/shape_elab.sh` (`$?` read immediately, negative arm greps the guard's own message) obey the rule and are the model for M2; `gptp-processor/bench/arty/build.sh` and `syn/ooc/run.sh` run under `set -e` (Vivado absent here, not exercised); gptp-processor has no `.github/` workflow at the pin, protocol-processor's `hdl.yml` declares no `shell:` but contains no pipelines; (d) guide scope/counts/budget vs both processors: wrong per BLOCKER, M3, M4, m2.
- `RTL` — `git diff` of `rx_mac_filter.sv` (only the generate-if chain added; no logic touched), `milan_datapath.sv:2893` instantiates at `TDATA_WIDTH=64` (legal), rx_filter suite 62/0 at head, `syn/yosys/run.sh --mode elaborate --top rx_mac_filter` PASS, `ooc.sh rx_mac_filter` PASS, lint ratchet PASS; enforcement scope per flow (MAJOR #2); GENUNNAMED (SUGGESTION).
- `Robustness` — `--check` exit codes (missing/one-number budget → 1; off-pin → 1; over-budget → 1), quoted/`$(...)`/comment/continuation/subshell/here-doc/shebang shapes (MINOR), workflow shell model (MAJOR), population boundaries (MINOR), `tracked()` refusal.
- `Tests` — contract-deletion mutation of `make negative` (MAJOR), `--selftest` arm-by-arm (19/19, premise of arm 14 false), runtime-`$error` classification (MAJOR), waiver mutation (MAJOR), suite_tally crafted logs (truncated → UNPARSED rc=1, empty → NOCOUNT rc=1, contradictory → rc=0), processor injection mutations ×6, workflow injection ×1.
- `Docs` — CODE_QUALITY.md Rule 6 section line by line vs tool output and vs GitHub's shell documentation; TOC entry and "Rules 1 to 6" consistent; no stale `rx_filter` target references in docs/testing; docs.yml step wiring correct (checkout + `git submodule update --init third_party/verilog-axis protocol-processor gptp-processor` at :32 precede the step at :106; `ci_events.py --check` OK 230); PR body figures (MINOR).

Head reviewed: 04d0fa5dedf5db147640b3e3d602eb8c0224be57

## Commands run (WORKDIR=/tmp/mfr-pr280; tree and both submodules restored after every mutation, `git status` clean except `REVIEW.md` and the ignored `tb/verilator/rx_filter/obj_dir/`)
- `git log --oneline BASE..HEAD`; `git submodule status`; `git diff --stat BASE..HEAD`; read of the full diff
- `python3 scripts/measure_fail_fast.py` / `--check` / `--selftest`; scratchpad `miss.py` (per-module hand count with the tool's own `blank_comments`/`MODULE`); `scan_module`/`scan_pipelines` probes via `python3 -c`
- `make -C tb/verilator/rx_filter negative`; direct `verilator --lint-only --top-module rx_mac_filter ... -G<param>` ×4 and legal default (with/without `-Wno-fatal`); full VFLAGS `--cc --exe --build ... -GTDATA_WIDTH=52 -Mdir obj_dir_mut` (removed); `make -C tb/verilator/rx_filter` (default target, log tallied with `suite_tally.py`)
- contract-deletion mutation (lines 126-137) + `make negative`; restore from copy; `git status`
- `sv2v --top=rx_mac_filter ...`; sv2v on a 3-form minimal probe; `yosys -p "read_verilog ...; chparam -set ...; hierarchy -check; proc"` ×4; `read_verilog -sv` control; `synth_xilinx -flatten; stat` at 64/52/32; `syn/yosys/run.sh --mode elaborate --top rx_mac_filter`; `OOC_CHPARAM="TDATA_WIDTH=52" syn/yosys/ooc.sh rx_mac_filter` and the legal control
- processor injections (`printf >>` into tracked files) ×6, workflow injection (`sed -i` docs.yml), waiver injection (`printf >>` run_all_suites.sh), each followed by `--check` and `git [-C sub] checkout -- <path>`
- `bash -e -c 'false | tee /dev/null'` vs `bash -eo pipefail -c ...`; WebFetch of docs.github.com workflow-syntax (shell table)
- budget file moved / rewritten with one number → `--check`; `git -C gptp-processor checkout --detach HEAD~1` → `--check` → checkout `903a5812`
- `python3 scripts/suite_tally.py --selftest` and on four crafted log dirs; `python3 scripts/ci_events.py --check`; `python3 scripts/lint_rtl.py --check`; `python3 scripts/docs_check.py`; `python3 scripts/gen_toc.py --check`; `python3 scripts/check_doc_paths.py`
- greps over `git ls-files --recurse-submodules` populations: Makefile recipe pipes (`grep -nP '^\t...'`), `SHELL`/`.SHELLFLAGS`, `|| true`/`set +e`, Python `os.system`/`shell=True`/`check=False`/`Popen`, Tcl `catch`/`exec`/`launch_runs`/`PROGRESS`, `.sh`/`.yml` producers outside the regex, continuation pipes, `xargs`/`find -exec`, pipefail shebang/subshell/comment shapes, workflow `shell:`/`defaults:`; reads of the four processor shell scripts, `syn/yosys/run.sh`/`ooc.sh`, `run_all_suites.sh`, `suite_tally.py`, `code_quality_scope.py`, `lint_rtl.py` (`LINT_EXCLUDE`), `docs.yml`, `bd/bringup.py`, `cov_gate.py`

</details>

Mister-M-alt:
**[A2] Answer to review round 2** — branch `254-code-quality-rule-6-fail-fast` restacked onto `253-code-quality-rule-5-explicit-contracts` @ `32c2d961`, head now `e7e6995b` (force-pushed; every commit is one line, no trailers, per CONTRIBUTING.md).

Commits on this PR (`git log base..254-code-quality-rule-6-fail-fast`):

```
    e7e6995b Refuse a suite that logs a failure and exits 0, and give Rule 6 its five boundaries: suite_tally.py --verdict reads a passing suite's log back for a [F…
    3b252045 Make the shield's contract bite in every Verilator build and grade it by its own diagnostic: -Werror-USERERROR joins the rx_filter and tcam_csr flags b…
    2e3ca352 Read every parameter form through the shared header parser and count only module-scope contracts in the fail-fast ratchet: the first regex saw 53 of 10…
    cd629aaf Model the two false-green shapes review reproduced in the pinned processors and correct the workflow shell model: a captured verdict decided from text …
    6793155e Keep the fail-fast fixture within bare-metal terminology
    0049bd11 Measure fail-fast contracts per module across processor RTL and make pipefail protection order-sensitive
    49622449 Refuse the build the shield cannot filter: Rule 6 gives rx_mac_filter an elaboration contract graded by mutation, and ratchets the two ways a failure c…
```

Gate, self-test and doc-gate outputs at `e7e6995b` (each line: command, exit code, last line printed):

```
    python3 scripts/measure_cohesion.py --selftest rc=0 :: 18 checks: 18 PASS, 0 FAIL
    python3 scripts/measure_control_flow.py --selftest rc=0 :: 45 checks: 45 PASS, 0 FAIL
    python3 scripts/check_rtl_source_lists.py      rc=0 :: RTL source-list gate: OK (109 files in the milan_datapath closure, 5 of 5 consumer list(s) carry all of them; protocol-p
    python3 scripts/check_rtl_source_lists.py --selftest rc=0 :: 40 checks: 40 PASS, 0 FAIL
    python3 scripts/measure_naming.py --check      rc=0 :: NAMING RATCHET: PASS (102 candidate(s), all recorded by identity; 102 recorded)
    python3 scripts/measure_naming.py --selftest   rc=0 :: 50 checks: 50 PASS, 0 FAIL
    python3 scripts/check_port_contracts.py        rc=0 ::   review inventory: 92 open and 43 literal-bound named connection(s), 76 without a local rationale (all recorded; none m
    python3 scripts/check_port_contracts.py --selftest rc=0 :: 70 checks: 70 PASS, 0 FAIL
    python3 scripts/measure_fail_fast.py --check   rc=0 :: FAIL-FAST RATCHET: PASS (85 <= 85 modules without an elaboration contract, 4 <= 4 masked pipeline(s), 2 <= 2 discarded c
    python3 scripts/measure_fail_fast.py --selftest rc=0 :: 90 checks: 90 PASS, 0 FAIL
    python3 scripts/suite_tally.py --selftest      rc=0 :: selftest: PASS
    python3 scripts/docs_check.py                  rc=0 :: docs_check: 0 finding(s) across 182 md files + 804 scrubbed text files, scrub self-test 17/17 [git ls-files]
    python3 scripts/gen_toc.py --check             rc=0 :: TOC gate: OK (132 page(s) carry an annotated contents list, 18 below the 3-section threshold)
    python3 scripts/check_doc_paths.py             rc=0 :: doc path gate: OK (1208 cited paths all resolve, 3 allowlisted, 1 ledger page(s) skipped; 89 line anchor(s) within their
    python3 scripts/ci_events.py --check           rc=0 :: ci_events: OK (230 contract item(s) across 4 workflow files and docs/testing/CI_WORKFLOWS.md)
    python3 scripts/lint_rtl.py --check            rc=0 :: LINT GATE: PASS (99 violation(s) <= ratchet 99; 22 waived, 0 justified lint_off)
    python3 scripts/pp_srcs.py --check             rc=0 ::   prose    tests/steps/aecp_engine_steps.py: BDD steps citing the RTL they mirror
    python3 scripts/check_soc_sources.py           rc=0 :: SoC source gate: OK (41 instantiated modules all registered, 75 sources all present)
    syn/yosys/check_tap_purity.sh                  rc=0 :: TAP-PURITY RESULT: PASS
    cd sw/builder && python3 test_builder.py       rc=0 :: ALL GATES PASS EXCEPT 1 NOT RUN
```

Commit hashes quoted inside the fix report are the pre-restack ones (the stack was rebased and every message reduced to one line afterwards); the commit list above gives the current hash for the same one-line subject.

<details><summary>Fix report — every finding's disposition with before/after reproduction</summary>

# FIX_REPORT — PR #280 "[Code quality] Rule 6: refuse the build the shield cannot filter", round 2

Worktree `/tmp/mfr-wt-254`, branch `254-code-quality-rule-6-fail-fast`. Original head `24770e7b`
(the maintainer's rebased `8f292850`, which already answered R1 and the workflow-shell MAJOR).
Final head `32a794e4`. Submodules at pin throughout (gptp-processor `903a5812`, protocol-processor
`3770ae02`); every probe below was restored with `git checkout -- <path>` / `git -C <sub> checkout -- .`.

Commits (each ends with the two required trailers):

- `53401dd3` — scanner: shared header parser for parameters, module-scope-only contracts, site
  waivers, comment/here-doc/subshell/continuation/shebang model, Makefile recipes, extended
  producers, derived arm count; `scripts/fail_fast.budget` re-based to 85 / 4 / 2 with reasons;
  the stale docs.yml step comment.
- `b43c4667` — `-Werror-USERERROR` in `tb/verilator/rx_filter` and `tb/verilator/tcam_csr`;
  negative arms graded by the contract's own diagnostic from a captured stderr file; the four
  guard blocks named `gen_guard_*`; the RTL banner states where the contract is and is not enforced.
- `32a794e4` — `suite_tally.py --verdict` + `run_all_suites.sh` refuse a suite that logs a failure
  and exits 0 (self-test arms through the CLI); the Rule 6 guide section rewritten: five in-tree
  examples, enforcement scope, measured populations with their scope stated, site waivers.

## Findings, in review order

### [R2] BLOCKER Conformance, Tests, Docs — scripts/measure_fail_fast.py:52-54 (`PARAM` regex); scripts/fail_fast.budget:7; docs/development/CODE_QUALITY.md:815 — The module ratchet cannot see the project's own parameter idiom
FIXED — `53401dd3`. Which modules have parameters is now read with `scripts/sv_ports.py`
`declarations()` (the header parser the Rule 4/5 gates share; Rule 3 says import, not re-implement),
so `int unsigned`, `string`, `type`, `real`, `longint`, a packed range, no-default and multi-name
lists all count; one arm per form (nine) and a live arm requiring `KL_gptp_timer`, `KL_gptp_ucpu`,
`KL_gptp_tx_slot` in the printed inventory. Budget 43 → 85 with the reason in the file; docstring
and guide table corrected. The count is 85 rather than the review's 82 because the next finding
moves the three runtime/`initial`-only sites into the unguarded set (102 = 17 guarded + 85).
- before: `parameterised modules with no elaboration contract (43 of 53)`; the `int unsigned`
  injection into `gptp-processor/hdl/common/KL_gptp_timer.sv` → `FAIL-FAST RATCHET: PASS (43 <= 43 ...)`, rc 0.
- after: `parameterised modules with no elaboration contract (85 of 102)`, the four gptp modules listed;
  `int unsigned` probe module appended to `KL_gptp_timer.sv` → `gptp-processor/hdl/common/KL_gptp_timer.sv:zz_probe`,
  `FAIL: 86 module(s) without an elaboration contract > ratchet 85`, rc 1; `string` probe appended to
  `protocol-processor/hdl/srp/KL_srp_vlan.sv` → same, rc 1. Reverting to the old regex in-process makes
  10 arms fail (the scratchpad mutation harness exec'd the mutated source and ran `selftest()`).

### [R2] MAJOR RTL, Robustness, Conformance — hdl/ieee8021q/filtering/rx_mac_filter.sv:126-137; tb/verilator/rx_filter/Makefile:13-16; syn/yosys/run.sh:364-376; syn/yosys/ooc.sh:479-551 — The contract refuses the build only under `verilator --lint-only` without `-Wno-fatal`
FIXED — `b43c4667` (RTL banner, Makefiles) and `32a794e4` (guide). Both halves of the required
change: (a) `-Werror-USERERROR` added to the rx_filter `VFLAGS` and to the tcam_csr build;
`tb/verilator/milan_dp` (Makefile:64, :202) and `tb/verilator/hostplane` (Makefile:35) already carry
`-Wno-fatal -Werror-USERERROR`, so nothing to add there — the review listed them as `-Wno-fatal`
consumers, but the flag beside it makes the contract fatal (verified: `-Wno-fatal -Werror-USERERROR
-GTDATA_WIDTH=52` → `%Error-USERERROR ... Exiting due to 1 error(s)`, rc 1). (b) The RTL banner and
the guide state exactly where the contract is enforced (Verilator with USERERROR fatal — its default
or `-Werror-USERERROR`; Vivado elaboration) and where it is not (sv2v lowers a module-scope `$error`
to an `initial $display` Yosys ignores, so `OOC_CHPARAM="TDATA_WIDTH=52" syn/yosys/ooc.sh
rx_mac_filter` still reports PASS), and cite `protocol-processor/tb/timer_map/shape_elab.sh` as the
shape a Yosys-side check would take, fed the original SystemVerilog. That Yosys-side check itself is
deliberately not built here (the review's "or" clause); the guide says so.
- before: the suite's own VFLAGS with `-GTDATA_WIDTH=52` → rc 0, binary built (review).
- after: `make -C tb/verilator/rx_filter run VERILATOR='verilator -GTDATA_WIDTH=52'` →
  `%Error-USERERROR: .../rx_mac_filter.sv:147:5: rx_mac_filter: TDATA_WIDTH=52 is not a whole number of bytes ...`, rc 2.
  `make -C tb/verilator/tcam_csr` with the flag → `tcam_csr: 2 checks, 0 failures`, rc 0.

### [R2] MAJOR Tests, Docs — tb/verilator/rx_filter/Makefile:56-75 (`negative`); CODE_QUALITY.md:795-799; PR body "four refusals" — Three of the four refusal arms pass with the contract deleted
FIXED — `b43c4667`. Each arm redirects stderr to `obj_dir/neg_<PARAM>_<value>.log` and
`grep -q "USERERROR.*rx_mac_filter: <PARAM>=<value> " <log>` is the assertion (a refusal without that
line is `[FAIL] ... refused ..., but NOT by the contract`); no pipe anywhere; `NEG_LINT` carries
`-Werror-USERERROR` so the arm does not depend on Verilator's default fatal policy.
- before (review, chain deleted): `[PASS] refused TDATA_WIDTH=32 / [FAIL] ACCEPTED 52 / [PASS] NUM_ENTRIES=0 / [PASS] ACTION_WIDTH=0`, 4 PASS 1 FAIL.
- after (chain deleted, lines `gen_guard_dmac_in_beat0` … `gen_guard_action_width` removed, `$error(` count 0):
  `[FAIL] elaboration refused TDATA_WIDTH=32, but NOT by the contract` / `[FAIL] elaboration ACCEPTED illegal TDATA_WIDTH=52` /
  `[FAIL] ... NUM_ENTRIES=0, but NOT by the contract` / `[FAIL] ... ACTION_WIDTH=0, but NOT by the contract` /
  `[PASS] elaboration accepted the legal default shape` — `5 checks: 1 PASS, 4 FAIL`, rc 2.
  Restored: `5 checks: 5 PASS, 0 FAIL`, each refusal printed `with the contract's own diagnostic`.

### [R2] MAJOR Robustness, Tests, Docs — scripts/measure_fail_fast.py:119-139 (workflow model), :220-221; CODE_QUALITY.md:822-825 — The workflow model assumes `-o pipefail`
FIXED — by the maintainer's `24770e7b` (= `8f292850`) before this round; verified here and kept
through the rewrite (per-step units, `shell: bash` before or after `run:`, per-step reset; the arms
"the GitHub Actions default run shell has NO pipefail", "a step that declares shell: bash is
protected", "pipefail set in one step does not protect the next step"). Mutating the default back to
pipefail makes 5 arms fail.
- before (review): docs.yml injection → `FAIL-FAST RATCHET: PASS (... 0 <= 0 masked ...)`, rc 0.
- after: `python3 scripts/measure_naming.py --check | tee ratchet.log` injected into the "Fail-fast ratchet"
  run block → `.github/workflows/docs.yml:111 python3 scripts/measure_naming.py --check | tee ratchet.log`,
  `FAIL: 5 masked pipeline(s) > ratchet 4`, rc 1 (docs.yml restored).

### [R2] MAJOR Conformance, Tests, Docs — scripts/measure_fail_fast.py:61-64 (`INTENTIONAL`), :107-110; CODE_QUALITY.md:819-821 — The waiver is a substring match on the whole line
FIXED — `53401dd3`. `INTENTIONAL` is `(path, exact whitespace-squashed line, reason)`; it is consulted
only after the producer test has found the line masked with pipefail off; a waiver whose line no longer
exists fails the live self-test arm; the guide describes it as by site.
- before (review): the two appended lines → `... the grep IS the version assertion ...` ×2, `PASS (... 4 waived ...)`.
- after: appended to `scripts/run_all_suites.sh` → `scripts/run_all_suites.sh:318 yosys --version | grep -q "0.0"`,
  `:319 python3 scripts/lint_rtl.py --check --version | grep -c FAIL`, `FAIL: 6 masked pipeline(s) > ratchet 4`, rc 1;
  the two real sites stay waived: `.github/workflows/rtl-fast.yml:99`, `.github/workflows/rtl.yml:184`.
  Mutating the lookup back to `"--version | grep" in text` makes 3 arms fail.

### [R2] MAJOR Tests, Conformance — scripts/measure_fail_fast.py:55 (`ELAB_CHECK` anywhere in the module) — A runtime `$error` counts as an elaboration contract
FIXED — `53401dd3`. Only a `$error`/`$fatal` at module or generate scope counts: strings, imports and
function/task bodies are blanked, then every `always*`/`initial`/`final` statement is spanned by a
small statement parser (`begin…end`, `fork…join`, `case…endcase`, `if…else`, loops, simple
statements to `;`). `initial begin … $fatal` (`KL_pp_acmp_listener`) and `initial … assert … else
$error` (`KL_avtp_common_parser`) are simulation checks every synthesis flow drops, so they are
counted unguarded — the docstring and the guide say so and why. Arms: always, always_ff block,
initial $fatal, initial assert, function body, named generate, generate/endgenerate, and four
span-termination shapes (a guard after an always block, after a single-statement always, after an
if/else, after `initial $readmemh`), plus "always" inside a message.
- probe: `scan_module('... always @(posedge clk) if (W) $error("runtime"); ...')` → before `(True, True)`, after `(True, False)`.
- the nine true guards stay guarded (KL_pcm_tx, KL_tdm_capture_master, milan_datapath, KL_aecp_desc_store,
  KL_aecp_engine, KL_aecp_resp_buf, KL_pp_originator, protocol_processor_top, rx_mac_filter), with
  KL_media_nco, KL_aaf_packetizer, KL_pair_blend, KL_pair_zero_fill, KL_media_clock_restart,
  KL_talker_diag_ctx, KL_pp_release_merge, KL_srp_encoder = 17 guarded. Mutating the spans away makes 4 arms fail.

### [R2] MAJOR Conformance, Docs — CODE_QUALITY.md:760-855; scripts/suite_tally.py:180-215 + scripts/run_all_suites.sh — Three #254 acceptance criteria claimed closed but not delivered
FIXED — `32a794e4`. (1) The section now has a table with one in-tree example per boundary: RTL
elaboration (`rx_mac_filter` `gen_guard_*`), FSM default (`hdl/ieee1722/aaf/KL_aes3_rx.sv`
`default: st_r <= HUNT_S;`), Python generator (`scripts/pp_srcs.py` raising on an empty list and its
consumers taking the status), shell pipeline (`gate | tee`), Tcl flow (`bd/build.tcl` `PROGRESS`
check and `syn/ooc/milan_datapath_ooc.tcl`'s `exec`/ROM refusal). (2) Assertions that only log are
inventoried and gated: `suite_tally.py --verdict <log>` (`log_reports_failure()`) refuses a log whose
tally counts a failure or that carries a line starting with `[FAIL]`; `run_all_suites.sh` runs it for
every suite whose `make` exited 0 and prints `FAIL <suite> (exited 0, but its log reports a failure -
a masked verdict)`, counted among the failed suites. Every `[FAIL]` printer in the tree sits on a
counted failure branch and the two suites that run deliberate failures (`nvm_backend`,
`tsn_fuzz/test_grade_tx.py`) keep that output out of their logs, so a green sweep stays green (the
full sweep itself was not run, per the brief). (3) That sweep acceptance is the representative masked
failure fixed here; the deliberately broken fixtures are the `verdict-*` and `cli-verdict-*` arms.
- before (review): `[FAIL] ...` beside `5 checks: 5 PASS, 0 FAIL` → `in-suite failures: 0`, rc 0; `4 PASS, 1 FAIL` → rc 0.
- after: `suite_tally.py --verdict` on the marker log → `VERDICT CONTRADICTED -- it carries 1 [FAIL] line(s) ...`, rc 1;
  on the `4 PASS, 1 FAIL` log → `VERDICT CONTRADICTED -- its tallies report 1 failure(s) across 5 checks`, rc 1;
  the sweep loop snippet on those logs → `FAIL bad (exited 0, but its log reports a failure - a masked verdict)`, `PASS good`.
  `python3 scripts/suite_tally.py --selftest` → `selftest: PASS` (10 new arms: 6 function, 3 CLI, 1 unreadable log → rc 2).

### [R2] MINOR Robustness, Tests — scripts/measure_fail_fast.py:76-139 — Latent scanner misses
FIXED — `53401dd3`, one arm per shape: comments blanked before the `set` search; here-document
bodies skipped (an unterminated one is `UNREADABLE POPULATION`, exit 2 — a partial population is
not a pass); `set` inside `( … )` scoped to the subshell, same-line and multi-line; backslash
continuations joined for pipelines (found at their first line); producers extended to `$MAKE`,
`./obj_dir/V*`, any `.sh` invocation, `xvlog`, `xelab`, `vivado`, `iverilog`, `verilator_coverage`,
tool variables (`$VERILATOR`, `$PY`) and path-qualified tools; a shebang `-eo pipefail` protects the
script and a plain shebang does not. Each snippet from the finding is an arm with the expected list.

### [R2] MINOR Conformance, Docs — scripts/measure_fail_fast.py:143 (population); CODE_QUALITY.md:808-818 — The pipeline population excludes Makefiles, `.py`, `.tcl`
FIXED — `53401dd3` + `32a794e4`. The scan reads every recipe line of the 93 first-party Makefiles
(vendor `third_party/`/`external/` excluded — `tracked()` recurses into verilog-axis — with a live
arm), each logical recipe line its own `/bin/sh -c`, `.SHELLFLAGS` honoured, `$(VERILATOR)`/`$(PY)`/
`$(MAKE)` expanded from the file's own definitions, `$$` lowered; `{ tool; echo $? > f; } | tee`
(tsn_fuzz) keeps its verdict. `tb/verilator/maap/Makefile:27` and `tb/verilator/avtp_rxmon/Makefile:56`
now appear in the inventory and the budget is 4 with the reason (their fix is to drop the pipe; left
for the suites' owners, outside this PR's scope). The tool prints `measured shell population: 34 shell
scripts, 5 workflow files, 93 Makefiles (recipe lines); Python and Tcl are not measured`, and the guide
states the population and why Python/Tcl are not measured (the review's survey, summarised).

### [R2] MINOR Docs — PR #280 body "Evidence" — Figures that do not match this head
NOT FIXED here — the PR body lives on GitHub, which this lane does not touch. The evidence block at
the final head, for the maintainer to paste, is in the last section below.

### [R2] SUGGESTION RTL — hdl/ieee8021q/filtering/rx_mac_filter.sv:126-137 — Unnamed generate blocks
FIXED — `b43c4667`. `gen_guard_dmac_in_beat0`, `gen_guard_byte_lanes`, `gen_guard_tcam_entries`,
`gen_guard_action_width`; `-Wall` lint of the legal default: GENUNNAMED 6 → 0 (the remaining
PINCONNECTEMPTY pair is the suite's existing waiver). The guide names the named form as the house
form and notes `milan_datapath`'s unnamed blocks. The `else if` chain (first violated law reported)
is kept as is: not a functional change the review required, and it is the `KL_media_nco` shape.

### [R2] SUGGESTION Robustness — scripts/code_quality_scope.py:32; scripts/measure_fail_fast.py:232-234 — Refusals and tallies rough
PARTIAL — `53401dd3` derives the printed arm total from the `ck()` calls (90 at this head; adding an
arm moves the total without editing a literal). The off-pin traceback in `code_quality_scope.py` is
left to the maintainer, as assigned.

## Gate, self-test and suite outputs at the final head `32a794e4` (last line of each)

- `python3 scripts/measure_fail_fast.py` → `waived by site, with a reason (2):` … `.github/workflows/rtl.yml:184 the grep IS the version assertion; its status is the verdict`
  (inventory: 85 of 102 modules; 4 masked pipelines; 2 captured verdicts; population 34 / 5 / 93)
- `python3 scripts/measure_fail_fast.py --check` → `FAIL-FAST RATCHET: PASS (85 <= 85 modules without an elaboration contract, 4 <= 4 masked pipeline(s), 2 <= 2 discarded captured verdict(s), 2 waived by site)` rc 0
- `python3 scripts/measure_fail_fast.py --selftest` → `90 checks: 90 PASS, 0 FAIL` rc 0
- `python3 scripts/suite_tally.py --selftest` → `selftest: PASS` rc 0
- `python3 scripts/docs_check.py` → `docs_check: 0 finding(s) across 182 md files + 802 scrubbed text files, scrub self-test 17/17 [git ls-files]` rc 0
- `python3 scripts/gen_toc.py --check` → `TOC gate: OK (132 page(s) carry an annotated contents list, 18 below the 3-section threshold)` rc 0
- `python3 scripts/check_doc_paths.py` → `doc path gate: OK (1158 cited paths all resolve, 3 allowlisted, 1 ledger page(s) skipped; 89 line anchor(s) within their file)` rc 0 (its pre-existing note about the stale `scripts/ci_scope.py` ALLOW entry is unrelated to this PR)
- `python3 scripts/ci_events.py --check` → `ci_events: OK (230 contract item(s) across 4 workflow files and docs/testing/CI_WORKFLOWS.md)` rc 0
- `python3 scripts/lint_rtl.py --check` → `LINT GATE: PASS (99 violation(s) <= ratchet 99; 22 waived, 0 justified lint_off)` rc 0 (same before the RTL edit)
- `make -C tb/verilator/rx_filter` → `checks: 62   failures: 0` / `RESULT: PASS` / `1 checks: 1 PASS, 0 FAIL` / `5 checks: 5 PASS, 0 FAIL` rc 0
- `make -C tb/verilator/tcam_csr` → `tcam_csr: 2 checks, 0 failures` rc 0
- `python3 scripts/measure_naming.py --selftest` → `46 checks: 46 PASS, 0 FAIL`; `python3 scripts/check_port_contracts.py --selftest` → `20 checks: 20 PASS, 0 FAIL` (the other `sv_ports` importers, untouched)
- `git status --short` → only this untracked `FIX_REPORT.md`; `git submodule status` → all four prefixes are spaces.

## Corrected evidence block for the PR body (all figures at `32a794e4`)

- `scripts/measure_fail_fast.py --selftest` — 90/90; `--check` — PASS (85 ≤ 85, 4 ≤ 4, 2 ≤ 2, 2 waived by site).
- Population: 102 parameterised modules, 85 without a module-scope contract, 17 with; 34 `.sh`, 5 workflow files, 93 Makefiles (recipe lines); Python and Tcl not measured.
- `tb/verilator/rx_filter` — 62 checks 0 failures, binding 1/1, negative 5/5 each by the contract's own USERERROR line; with the `$error` chain deleted 1 PASS 4 FAIL (rc 2); `-GTDATA_WIDTH=52` in the build flags → `%Error-USERERROR`, rc 2.
- `tb/verilator/tcam_csr` — 2 checks 0 failures with `-Werror-USERERROR`.
- `suite_tally.py --selftest` — PASS (log-verdict arms included); `ci_events.py --check` — OK (230 items); `lint_rtl.py --check` — PASS (99 ≤ 99); `docs_check` 0 findings / `gen_toc --check` OK / `check_doc_paths` OK.
- Not enforced and said so: the sv2v → Yosys path (`OOC_CHPARAM="TDATA_WIDTH=52" syn/yosys/ooc.sh rx_mac_filter` still PASS).

## git log --oneline 24770e7b..HEAD

</details>

Mister-M-alt:
**[A2 follow-up]** Head moved `e7e6995b` → `df40ca57` (base `32074fb0`): CI's Verilator shard 0/4 failed `rx_filter` on every head from #279 up. `tb/verilator/rx_filter/binding_mutant.py` reads the clean build recipe through `make -s print-vflags` (Rule 3: one copy of the flags), and under the runner's GNU make 4.3 the sub-make also prints its `make[1]: Entering directory …` line — make 4.4, used locally, suppresses it under `-s` — so the driver executed that line as the Verilator command (`FileNotFoundError`). Fixed on the 253 branch (`32074fb0`: `--no-print-directory`, and make's own lines are dropped), reproduced and proven locally with `MAKEFLAGS=w` (before: the traceback; after: 5/5 arms, suite log accepted by `suite_tally.py --verdict` and its accounting). One testbench driver file changed; every gate, self-test and doc gate re-run green at the new head, RTL unchanged since the reported full sweep and Yosys gate.

scripts/measure_fail_fast.py
```diff
@@ -0,0 +1,982 @@
...
+    `code` is one module with comments blanked. Strings go too (a message may
+    say "always"), then imports (a DPI `import ... function` has no
+    `endfunction`), then function and task bodies; what is left inside a
+    procedural span is a simulation check, and everything else is the
+    contract."""
+    code = STRING.sub(_blank, code)
+    code = IMPORT.sub(_blank, code)
+    code = _FUNC_TASK.sub(_blank, code)
+    spans = procedural_spans(code)
+    return any(not any(a <= m.start() < b for a, b in spans)
```
Mister-M-alt: [P2] Do not count concurrent assertions as elaboration guards

Only procedural spans are excluded here, so a module-scope concurrent assertion is counted as an elaboration contract. Reproduced with `assert property (@(posedge clk) P) else $error;`: `scan_module()` returns `(True, True)`. Its action block fires during simulation at a clock edge and cannot refuse an illegal parameter at elaboration, so the unguarded-module ratchet is undercounted. Please exclude concurrent assertion bodies, or positively recognize module/generate elaboration checks, and add this fixture.

Mister-M-alt: Fixed in `601a2e66`. Guard recognition is now positive: `elaboration_checks()` descends only into what elaboration evaluates (generate, if/else, for, case, named blocks) and steps over every other item whole — `always*`/`initial`/`final`, any `assert`/`assume`/`cover`/`restrict` with its action block (immediate, deferred `#0`/`final`, concurrent `property`/`sequence`), and declarations. Your fixture: `scan_module()` before `(True, True)`, after `(True, False)`; `assert #0 (...) else $error` is likewise not a contract (documented); a concurrent assertion beside a real `if (W<2) $error` stays guarded. 16 arms added (106/106; the old scanner fails 6 of them). Re-measured: no first-party module writes a concurrent or deferred assertion today, so the inventory is byte-identical (85 of 102 unguarded, same 17 guarded) and the budget stays 85 / 4 / 2 with the re-measurement recorded in its comment.

Issue #287, opened 2026-08-28
[TASK] Pin the Yosys the portability gate is certified with, and make its cells= record real


Mister-M-alt:
## Objective / problem

The Yosys portability gate certifies that the RTL maps to a generic,
device-independent cell library. CI runs that certification with an
**unpinned** Yosys, and the evidence it publishes for it is empty.

Two coupled defects, both measured below:

1. `rtl.yml` and `rtl-fast.yml` install Yosys with
   `sudo apt-get install -y --no-install-recommends yosys` in three places.
   Every other tool in the same class is pinned — Verilator to `v5.050`
   (built from source *and* verified), `sv2v` to `v0.0.12`, `sbt` to
   `v1.10.7`. Yosys is the only one that floats, and it is the one the gate
   is named after.
2. The `cells=` column of every per-top result record is `?` in CI — all 46
   of them — because the extractor in `syn/yosys/run.sh` parses a Yosys
   output format the CI Yosys does not produce. `scripts/yosys_tally.py`
   excludes `cells` from its required keys, so nothing notices.

Make the tool the gate is certified with explicit and verified, and make the
recorded evidence either real or absent — not silently `?`.

### CI runs Yosys 0.33; a developer box runs 0.66

From [run 33146002383](https://github.com/kebag-logic/milan-fpga/actions/runs/33146002383)'s
own `yosys -V` step, on both the Yosys and Verilator workers:

```
Setting up yosys-abc (0.33-5build2) ...
Setting up yosys (0.33-5build2) ...
Yosys 0.33 (git sha1 2584903a060)
```

`sv2v` immediately below it in the same job is pinned and prints `sv2v
v0.0.12`. Verilator is pinned by `VERILATOR_VERSION: v5.050`, built from a
tag, and then **verified**:

```yaml
/opt/verilator/bin/verilator --version | grep -F "${VERILATOR_VERSION#v}"
```

There is no equivalent line for Yosys. `apt-get install yosys` is whatever
the `ubuntu-latest` archive resolves to on the day, and it resolved to a
release that is 33 versions behind the one a bench box has.

### The version changes the gate's answer, not just its formatting

Measured by fetching the **exact** Ubuntu binaries CI installs
(`yosys_0.33-5build2_amd64.deb` + `yosys-abc_0.33-5build2_amd64.deb`,
`git sha1 2584903a060`) and running the gate's own program against the
identical sv2v output as the locally installed 0.66:

[...]

Issue #286, opened 2026-08-28
[TASK] Measure the single-process Yosys cost: allocator, front end, build flags, host tuning


Mister-M-alt:
## Objective / problem

Establish, by measurement on this repository's own RTL, what a **single Yosys
process** actually costs and which of the four commonly-cited accelerators
moves it: a Verific front end, a replacement `malloc`, a natively-compiled
Yosys, and host/firmware tuning. Publish the comparison and a per-lever
go/no-go so the CI-speed work in #270 spends effort where the time is.

#270 covers *how many* Yosys processes run and how they are scheduled. #271
evaluated *replacing* Yosys and was closed. Neither answers *what one Yosys
process spends its time on*, which is the binding constraint: the whole
portability gate cannot finish faster than its single longest top.

The first pass of that measurement is below. It already takes the local gate
from **16m56s to 5m58s** with the full 46-top inventory still passing, and it
retires two of the four candidate levers on evidence. What this ticket asks
for is to land that measurement as a reproducible harness, confirm it, and
turn each lever into a public go/no-go.

This ticket is evidence-producing. It does not authorise changing the gate's
pass criteria, its `cells=` records, or the toolchain CI installs.

## Evidence / motivation

All figures below were taken on one bench machine with the locally installed
toolchain (`yosys 0.66` = upstream `v0.66`, `abc 1.01` built 2026-06-19,
`sv2v 0.0.13`) on the current `dev` head, running the exact program
`syn/yosys/run.sh` issues in `full` mode. They are a first pass, they are
stated with the concurrency each trial ran at because a lone run and a batched
run are not comparable, and the box had a second tenant for part of the
window. Reproducing them under the harness this ticket asks for is the first
deliverable, not an afterthought. Note that CI does not run this Yosys — see
the defects at the end.

### The gate's critical path is one process, and it is optimisation-bound

`milan_datapath` elaborates to **1,137,752 wires / 1,597,718 cells** and is
synthesised by a single-threaded Yosys process. `yosys -d` on that run
(585 s wall, 711 s CPU, 3,051 MiB peak RSS, near-idle machine):

| pass | share | calls | seconds |
| --- | --- | --- | --- |
| `opt_dff` | 25% | 90 | 183.5 |
| `opt_clean` | 23% | 96 | 164.2 |
| `abc` | 16% | 1 | 116.4 |
| `opt_expr` | 14% | 446 | 99.5 |
| `opt_merge` | 6% | 94 | 43.6 |
[...]

PR #289, opened 2026-08-28
Preload jemalloc for the Yosys portability gate, and record why -defer is not the lever here

### LINKED ISSUE #288: [TASK] Speed up the Yosys portability gate without changing what it proves: jemalloc preload and read_verilog -defer
## Objective / problem

Make the Yosys portability gate faster without changing what it proves, using
the two levers #286 identified that cost nothing in evidence:

1. run the gate's `yosys` invocations under a **jemalloc preload** when the
   library is present;
2. use **`read_verilog -defer`** where the gate actually reads modules it will
   not use — and document, with numbers, where it does not, so the flag is not
   cargo-culted into the places it cannot help.

(1) is #286's decision 1, option 1, taken by the maintainer. #286 stays what it
is — the measurement — and this ticket is the implementation it said a cleared
lever would need.

Neither change may alter what the gate proves. The preload must not add a hard
dependency to a gate whose documented requirement is "`yosys` and `sv2v` on
`PATH`", and both must be visible in the output so a reader knows which
configuration produced a timing.

## Evidence / motivation

From #286, on one bench box, `yosys 0.66`:

| scope | glibc | jemalloc | change |
| --- | ---: | ---: | ---: |
| `milan_datapath`, one process | 656.3 s | 361.5 s | **−44.9%** |
| `KL_pp_shadow`, one process | 273.9 s | 171.8 s | −37.3% |
| whole gate, 8 shards, 46/46 pass | 616.2 s | 357.8 s | **−41.9%** |
| whole gate, serial → 8 shards + jemalloc | 1015.9 s | 357.8 s | −64.8% |

Peak RSS **falls** on both heavy tops (3033 → 2786 MiB on `milan_datapath`),
so this does not spend the memory headroom #270's shard pool needs.

The output is identical, and that was measured rather than assumed: `write_rtlil`
after the gate's own program on `KL_gptp_shadow` is **byte-identical**
(`sha256 cc79dc89adbce892…`, 43,920,712 bytes) under glibc, tcmalloc, jemalloc
and mimalloc.

jemalloc was the fastest of the three replacements measured (jemalloc >
mimalloc > tcmalloc > glibc) at every design size.

### `read_verilog -defer`: where it pays, and where it cannot

`-defer` postpones turning a parsed module into RTLIL until something
instantiates it. It can only win back time spent elaborating modules the
selected top does not use, so its prize is bounded by the front end's share of
a run — and this gate is optimisation-bound (#286: `opt*` ≈ 74%,
`read_verilog` ≈ 0.5% on `milan_datapath`).

That bound is not uniform across the tree, which is the whole point of
documenting it:

- **`syn/yosys/run.sh`** hands every top its *own* `sv2v` output built from
  that top's own source list, so most tops have nothing to defer. The
  exceptions are the tops whose list is `$PP_SRCS` — the entire
  protocol-processor tree — of which they instantiate a subset. Measured in
  this ticket's PR.
- **`protocol-processor/syn/yosys/run.sh`** lowers its whole `hdl/` tree into
  one `all.v` and then re-reads it once per top, 32 times. There `-defer` is
  39.65 s → 10.31 s (**−74%**), and with a bounded pool 1.25 s. That is a
  different repository and is tracked in its issue #25 — including the trap
  that the redundant re-parse is currently the only front-end coverage six of
  its modules get, so the tops array has to be completed before `-defer` is
  applied there.

## Authoritative references

- #286 — the measurement, the netlist-invariance proof, and the decision this
  implements
- `syn/yosys/run.sh` — the loop that invokes `yosys`
- `syn/yosys/README.md` — the documented tool requirement this must not break
- `scripts/yosys_tally.py` — `REQUIRED_KEYS` plus `{"mode", "cells"}`; the
  result record schema is closed, so the allocator must not become a new key
  in it without changing that validator
- #270 — the CI trust rules any environment the gate exports must respect

## Context entry points

Read first:

- `syn/yosys/run.sh`
- `syn/yosys/README.md`
- `scripts/yosys_tally.py`
- #286

## Scope

In scope:

### 1. Allocator preload

- Detect a usable jemalloc and run the gate's `yosys` child processes under it.
- Apply it to the `yosys` invocation only, not to the script as a whole:
  `sv2v` is a GHC binary and `python3` runs the source-list and shard
  helpers; neither was measured and neither needs it.
- Keep it optional in both directions. Absent library ⇒ the gate runs exactly
  as it does today, with no failure and no warning-as-error. An explicit
  environment override selects a different library or turns the preload off,
  so a bisect or a reproduction can pin the allocator.
- Print which allocator the run used, in the gate's own header line, so a
  timing quoted from the output is attributable.
- Add a self-check for the selection logic that runs without jemalloc
  installed, so the "library absent" path is proved rather than assumed.

### 2. `read_verilog -defer`

- Measure `-defer` on the tops whose source list is the whole
  protocol-processor tree (`KL_pp_shadow`, `milan_datapath`) against the same
  tops without it, same allocator, one matched batch, and adopt it only where
  the measurement says it pays.
- Prove the netlist is unchanged by the flag before adopting it anywhere:
  `-defer` changes *when* a module is elaborated, and a gate that stops
  elaborating something is a gate that stops checking it.
- If it does not pay on this gate, say so in the documentation with the
  number, and do not add the flag. A flag that buys nothing here is a flag the
  next reader has to re-derive the meaning of.

### 3. Documentation

- Document both levers in `syn/yosys/README.md`: what each one does, the
  measured effect with its source (#286 and this PR), and — for `-defer` — the
  rule for when it helps at all, so it is applied to the next script by
  evidence rather than by habit.
- State plainly that both change speed and not results, and name the proof.

Out of scope:

- `syn/yosys/ooc.sh` and `protocol-processor/syn/yosys/run.sh`. The same lever
  applies to both and both should get it, but one script at a time — and the
  protocol-processor one is a different repository with its own gate
  (see its issue #25).
- Making the preload mandatory, installing jemalloc, or adding it to a CI
  workflow. CI's Yosys is not even pinned yet (#287); exporting an allocator
  into a floating toolchain is the wrong order.
- Any change to the top inventory, the `synth` + `hierarchy -check`
  criterion, the `cells=` record, the shard selector or the structural checks.
- Applying `-defer` to the protocol-processor gate. Its own issue #25 requires
  six missing tops to be added first, and that is a different repository.
- Adding a key to the per-top result records. `scripts/yosys_tally.py` rejects
  unknown keys, and widening that schema is #287's business, not this one.

## Executor

Agent A [A1]

## Independent reviewer

Unassigned

## Acceptance criteria

- [ ] `syn/yosys/run.sh` runs `yosys` under jemalloc when the library is
      present, and unchanged when it is not.
- [ ] With the library removed from the search path, the gate still passes all
      46 tops and prints no error.
- [ ] An environment override selects an explicit library path, and a
      documented value disables the preload entirely.
- [ ] The preload reaches `yosys` and nothing else — `sv2v` and `python3` in
      the same script are not run under it.
- [ ] The header line names the allocator actually in effect for that run.
- [ ] A full local `syn/yosys/run.sh` run is green with the preload active,
      with an unchanged top count of 46 and unchanged `cells=` values for the
      tops that report one.
- [ ] The netlist is proved unchanged, not assumed: `write_rtlil` on at least
      one top is byte-identical with and without the preload, and the digests
      are in the PR.
- [ ] `python3 scripts/lint_rtl.py --check`, `python3 scripts/pp_srcs.py
      --check` and `python3 scripts/yosys_shards.py --selftest` stay green.
- [ ] The selection logic has a self-check that runs on a machine without
      jemalloc.
- [ ] `-defer` is measured on `KL_pp_shadow` and `milan_datapath` with the
      numbers in the PR, and is adopted **only** where the measurement
      justifies it. A null result is an acceptable outcome and must be
      documented rather than papered over.
- [ ] Wherever `-defer` is adopted, the emitted netlist is byte-identical to
      the same run without it, with the digests in the PR.
- [ ] `syn/yosys/README.md` explains when `-defer` helps and when it cannot,
      with the measured numbers, so the next script is changed on evidence.

## Validation

```sh
# selection logic, both directions
syn/yosys/run.sh --selftest-alloc
YOSYS_MALLOC=none syn/yosys/run.sh --list >/dev/null

# the gate, with and without
syn/yosys/run.sh --no-structural --results /tmp/with
YOSYS_MALLOC=none syn/yosys/run.sh --no-structural --results /tmp/without
diff -r /tmp/with /tmp/without          # records must be identical

# netlist invariance, in the PR
for a in none /usr/lib/libjemalloc.so.2; do
  YOSYS_MALLOC=$a yosys -p "read_verilog work/$TOP.v; synth -top $TOP; \
    hierarchy -check; write_rtlil out/$TOP.$a.il"
done
sha256sum out/$TOP.*.il

python3 scripts/pp_srcs.py --check
python3 scripts/yosys_shards.py --selftest
```

Expected result: 46/46 tops pass in both configurations, the two result trees
are identical, the two netlist digests are equal, and the only difference is
wall time.

## Dependencies / lane conflicts

Depends on:

- #286 for the measurement and the netlist-invariance evidence.

May conflict with:

- #287, which will change how `syn/yosys/run.sh` records evidence and how CI
  obtains Yosys. Both touch `syn/yosys/run.sh`; land this one first, it is
  smaller.
- #270 and any concurrent edit to `syn/yosys/run.sh`.

## Open design/specification decision?

No — scope is settled and may become Ready

## Decision details (if applicable)

Decision owner: maintainer. Taken: #286 decision 1, option 1 — adopt the
allocator preload, behind an environment knob, with the netlist digests as the
gate. Options 3 (a pinned rebuilt Yosys) and 5 (a result cache) remain open in
#286 and are unaffected by this.

## Discovered from

#286, and the maintainer's decision to implement its first lever and to
document and apply `read_verilog -defer` where it earns its place.

Mister-M-alt:
Closes #288.

`syn/yosys/run.sh` now preloads jemalloc for its `yosys` children when the
library is installed, and `syn/yosys/README.md` gains a **Runtime levers**
section that records why that is the lever worth having here — and why
`read_verilog -defer`, measured on the same tops, is not.

## What changed

- `syn/yosys/run.sh`: resolve an allocator once, after the tool check, and
  export `LD_PRELOAD` inside the per-top subshell so it reaches `yosys` and
  the `abc` it spawns, and nothing else. New `--selftest-alloc`. New
  `YOSYS_MALLOC` environment knob. The header line names the allocator in
  effect.
- `syn/yosys/README.md`: the pass profile that makes this gate
  optimisation-bound, the allocator numbers, and the `-defer` null result with
  the rule for when that flag *does* pay.

No RTL, no top inventory, no `synth` + `hierarchy -check` criterion, no
`cells=` record, no shard selector, no workflow.

## Why: the gate is allocation-bound, not parse-bound

`yosys -d` on `milan_datapath` (1,137,752 wires / 1,597,718 cells, one
single-threaded process): `opt_dff` 25%, `opt_clean` 23%, `abc` 16%,
`opt_expr` 14% — and `read_verilog` **0.5%**.

| scope | glibc | jemalloc |
| --- | ---: | ---: |
| `milan_datapath`, one process | 656.3 s | **361.5 s** (−44.9%) |
| `KL_pp_shadow`, one process | 273.9 s | 171.8 s (−37.3%) |
| whole gate, 8 shards, 46/46 pass | 616.2 s | **357.8 s** (−41.9%) |

Peak RSS falls too — 3033 → 2786 MiB on `milan_datapath` — so this does not
spend the headroom a parallel shard run needs. Full measurement in #286.

## It changes speed and not results

Not asserted, measured. `write_rtlil` after this script's own program on
`KL_gptp_shadow` is **byte-identical** across allocators:

```
distro + glibc      cc79dc89adbce892  43920712 bytes
distro + jemalloc   cc79dc89adbce892  43920712 bytes
distro + tcmalloc   cc79dc89adbce892  43920712 bytes
distro + mimalloc   cc79dc89adbce892  43920712 bytes
```

And the gate's own evidence is identical with and without the preload — the
full 46-top run was done twice on this branch, once each way, and the
`--results` trees compared byte for byte:

```
$ for i in 0..7; do ./syn/yosys/run.sh --shard $i/8 --no-structural --results with     ; done
$ for i in 0..7; do YOSYS_MALLOC=none ./syn/yosys/run.sh --shard $i/8 --no-structural --results without ; done
with:    46 tops 46 pass 0 fail
without: 46 tops 46 pass 0 fail
$ diff -r with without
IDENTICAL
```

## The other half: `read_verilog -defer` is deliberately NOT used here

The ticket asked for `-defer` to be applied and documented. Measured on the
two tops with the largest surplus of unused modules — both are handed the
entire protocol-processor tree and instantiate a subset — same allocator, one
matched batch:

| top | plain | `-defer` | `read_verilog` step | cells |
| --- | ---: | ---: | ---: | ---: |
| `milan_datapath` | 353.5 s | 350.0 s (−1.0%) | 2.30 s → 0.26 s | 1,597,718 → 1,598,979 |
| `KL_pp_shadow` | 165.9 s | 166.4 s (+0.3%) | 0.89 s → 0.16 s | 1,133,318 → 1,148,270 |

`-defer` does exactly what it claims — it removes 82–89% of the
`read_verilog` step — and that step is 0.5% of the run, so the gate does not
get faster. It also **changes the netlist**: +0.08% cells on `milan_datapath`
and +1.3% on `KL_pp_shadow`, because deferring changes which parameterised
specialisations exist by the time the `opt*` passes run. The control rules out
noise: the same binary run twice on `milan_datapath` reproduces its own digest
(`0d541ffab0a4cef7`), while the `-defer` run gives `ffcca31dd036de26`.

Buying nothing and changing the answer is a no-go, so the flag is not added
and the reason is in the README instead of being re-derived by the next
reader. Where it *does* belong is stated there too: a script that reads one
file holding many modules once per top. That is the control plane's own
portability gate (39.65 s → 10.31 s, and 1.25 s with a pool), tracked in that
repository's issue #25 — which should complete its `tops` array first, since
six of its modules are in no top at all.

One thing `-defer` does **not** cost, and this corrects an assumption I made
in #25 before measuring it: **it still parses every module.** Four fault
classes injected into a module that no top instantiates — a syntax error, a
negative-width vector, an undefined submodule and a `$fatal` — are caught
identically with and without the flag. `-defer` postpones elaboration, not
parsing. So the objection to `-defer` here is purely that it buys nothing and
moves the netlist; it is not a coverage objection.

## Compatibility

The documented tool requirement is unchanged: `yosys` and `sv2v` on `PATH`. A
machine without jemalloc runs exactly as before and prints
`yosys allocator: system`. A **named** library that is absent refuses rather
than falling back, so a run asked to reproduce one allocator cannot quietly
report another one's timing.

CI is untouched. Its Yosys is not even pinned yet (#287); exporting an
allocator into a floating toolchain is the wrong order.

## Validation

```
$ ./syn/yosys/run.sh --selftest-alloc
allocator selection self-test: PASS

$ python3 scripts/yosys_shards.py --selftest
selftest: PASS

$ python3 scripts/lint_rtl.py --check
LINT GATE: PASS (99 violation(s) <= ratchet 99; 22 waived, 0 justified lint_off)

$ python3 scripts/pp_srcs.py --check        # rc=0
$ python3 scripts/gen_toc.py --check
TOC gate: OK (131 page(s) carry an annotated contents list, 18 below the 3-section threshold)

docs_check: 0 finding(s) across 181 md files + 785 scrubbed text files, scrub self-test 17/17

$ ./syn/yosys/run.sh --list | wc -l
46
```

`xvlog_gate.py` was not run: this change touches no `.sv` or `.v` file.

### DIFF
--- a/syn/yosys/README.md
+++ b/syn/yosys/README.md
@@ -132,11 +132,13 @@ pool. Before applying it there, read that repository's issue #25: six of its
 modules are not in its `tops` array at all, so nothing ever elaborates them as
 a top, and that array should be completed first.
 
-One thing `-defer` does **not** cost, measured rather than assumed: it still
-parses every module. Four fault classes injected into a module no top
-instantiates — a syntax error, a negative-width vector, an undefined
-submodule, and a `$fatal` — are caught identically with and without the flag.
-`-defer` postpones elaboration, not parsing.
+One thing `-defer` does **not** cost, controlled on that gate by injecting a
+fault into a module no top instantiates and running both forms over it:
+coverage. A syntax error is caught either way — by `sv2v`, before Yosys reads
+anything — and an instance of an undefined module is missed either way,
+because nothing elaborates a module that is in no top. `-defer` postpones
+elaboration; what provides elaboration coverage is the `tops` array, not the
+reading mode.
 
 The rule to carry to the next script: **`-defer` is a fix for re-reading, not
 a fix for synthesis.** Check the `read_verilog` line in `yosys -d` output
--- a/syn/yosys/README.md
+++ b/syn/yosys/README.md
@@ -128,9 +128,15 @@ nothing. The flag belongs in the control plane's own portability gate
 (`syn/yosys/run.sh` inside the `protocol-processor` submodule), which lowers
 its whole `hdl/` tree into a single `all.v` and then re-reads it once per top,
 32 times — there it is 39.65 s → 10.31 s, and 1.25 s with a bounded process
-pool. Before applying it there, read that repository's issue #25: the
-redundant re-parse is currently the only front-end coverage six of its modules
-get, so the tops array has to be completed first.
+pool. Before applying it there, read that repository's issue #25: six of its
+modules are not in its `tops` array at all, so nothing ever elaborates them as
+a top, and that array should be completed first.
+
+One thing `-defer` does **not** cost, measured rather than assumed: it still
+parses every module. Four fault classes injected into a module no top
+instantiates — a syntax error, a negative-width vector, an undefined
+submodule, and a `$fatal` — are caught identically with and without the flag.
+`-defer` postpones elaboration, not parsing.
 
 The rule to carry to the next script: **`-defer` is a fix for re-reading, not
 a fix for synthesis.** Check the `read_verilog` line in `yosys -d` output
--- a/syn/yosys/README.md
+++ b/syn/yosys/README.md
@@ -21,6 +21,7 @@ make ecp5       # map to a real non-Xilinx device: Lattice ECP5 (TRELLIS_FF/LUT4
 
 - **[How it works](#how-it-works)** — The two-stage pipeline and why each stage is there: sv2v converts the SystemVerilog Yosys cannot parse (interfaces, packages, assignment patterns), then `hierarchy -check` is what makes a PASS mean something — it fails on any surviving vendor primitive, so green = fully mapped to generic logic with nothing Xilinx-specific left.
 - **[Tooling](#tooling)** — The two binaries you need and where to get them. No Xilinx tools are required, which is the point — this flow is the evidence that the RTL is not tied to one vendor's toolchain.
+- **[Runtime levers](#runtime-levers)** — Why this gate is optimisation-bound and not parse-bound, which is the one fact that decides every speed question here: the jemalloc preload it now runs under (−45% on the heaviest top, −42% on the whole sharded gate, byte-identical netlist) and how to turn it off, and the measured reason `read_verilog -defer` is *not* used — it removes 85% of a step that is 0.5% of the run, and changes the cell count by 1.3% while doing it.
 - **[Coverage](#coverage)** — What the tops actually span, and the standing rule that the `tops=()` array is the count while this prose is not. Also names the one deliberate gap: `milan_top`, which pulls in the RGMII SelectIO and PS block design.
 - **[Notes](#notes)** — Two facts that stop you misreading the output: the concrete non-Xilinx targets (`synth_ecp5`, `synth_ice40`) with real cell counts, and why `axis_fifo` looks enormous — its 4096-deep default, which no instance in the design uses.
 - **[ooc.sh — AREA measurement (a different question from run.sh)](#oocsh--area-measurement-a-different-question-from-runsh)** — `run.sh` asks *does it map*; this asks *what does it cost*, which is the only number an area lever may be judged on. Three traps it exists to avoid, and the third is the sharpest: `-flatten` can read a genuinely deleted block as **−1 LUT / −0 FF**, so a structural lever needs the hierarchy-preserving form as well. Ends with the honest caveat — these are estimates with a yosys→Vivado ratio between 0.25 and 0.86, and control sets are not measurable here at all.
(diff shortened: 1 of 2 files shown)


Mister-M-alt (commented):
I found two correctness issues in the allocator override behavior:

1. **[P1] Resolve and validate explicit allocator paths before reporting them** (`syn/yosys/run.sh:132-136`)

`select_malloc` only checks an explicit path with `-e` and returns it unchanged. Yosys later runs after `cd "$TMP"`, so a valid relative path no longer points to the selected library. On this head, running from `/usr/lib` with `YOSYS_MALLOC=./libjemalloc.so.2` printed that allocator and passed, although the dynamic loader ignored the preload. Existing non-libraries such as `/etc/hosts` are likewise accepted and the gate passes; the loader diagnostic is redirected into the successful per-top log and never shown. This can silently attribute system-allocator timings to the requested allocator. Please canonicalize explicit paths before the directory change and verify that selected candidates are loadable, refusing them otherwise.

2. **[P2] Clear inherited LD_PRELOAD when YOSYS_MALLOC=none** (`syn/yosys/run.sh:489-492`)

When selection returns an empty `MALLOC_LIB`, the conditional does nothing, so Yosys inherits any pre-existing `LD_PRELOAD`. I reproduced this with `LD_PRELOAD=/usr/lib/libjemalloc.so.2 YOSYS_MALLOC=none`: the header printed `yosys allocator: system`, but Yosys still inherited jemalloc. This breaks the documented disable mode and invalidates with-versus-without timing comparisons in such an environment. The empty or `none` branch should explicitly unset `LD_PRELOAD` inside the Yosys subshell.

All 20 GitHub checks pass. I also ran the allocator self-test, Bash syntax check, one-top default and override elaboration checks, lint, pp-srcs, shard self-test, documentation check, TOC check, and hermetic-list check successfully. These cases are not covered by the current self-test.

Mister-M-alt:
[A1] Both findings confirmed and fixed at `943c08d0`. I reproduced each one on the reviewed head before changing anything — thank you, they are real and they defeat the exact property this change was supposed to have.

## P1 — an explicit allocator was believed without being resolved or verified

Reproduced on `d0ec3525`, both halves:

```
$ (cd /usr/lib && YOSYS_MALLOC=./libjemalloc.so.2 …/run.sh --top cdc_pulse --no-structural)
   yosys allocator: ./libjemalloc.so.2        <- resolved against /usr/lib, but yosys runs from $TMP
  [PASS] cdc_pulse

$ YOSYS_MALLOC=/etc/hosts ./syn/yosys/run.sh --top cdc_pulse --no-structural
   yosys allocator: /etc/hosts
  [PASS] cdc_pulse

$ LD_PRELOAD=/etc/hosts /bin/true ; echo $?
ERROR: ld.so: object '/etc/hosts' from LD_PRELOAD cannot be preloaded (invalid ELF header): ignored.
0
```

That last line is the root of it: **the loader rejects a bad preload by ignoring it, and still exits 0.** Exit status can never detect this, and the diagnostic goes to stderr, which for a yosys child is redirected into that top's log and never read.

Fixed in three parts:

- existence is checked **before** canonicalisation, so a relative request is judged against the caller's directory — where they meant it — and the message says what is actually wrong;
- the path is then made **absolute** (`abs_path`, `readlink -f` with a `dirname`/`pwd` fallback), because `yosys` runs after `cd "$TMP"`;
- loadability is **probed**, and the test is *stderr stayed empty*, not *the command succeeded*: `err="$(LD_PRELOAD="$lib" "$TRUE_BIN" 2>&1 >/dev/null)"; [ -z "$err" ]`.

An explicitly named library that fails any of these is **refused** (exit 2) and the loader's own words are printed. An **auto-detected** candidate that fails is skipped and the search continues, because the documented default is "use jemalloc when it is installed" and a broken one is not installed.

```
$ (cd /usr/lib && YOSYS_MALLOC=./libjemalloc.so.2 …/run.sh --top cdc_pulse --no-structural)
   yosys allocator: /usr/lib/libjemalloc.so.2
$ YOSYS_MALLOC=/etc/hosts ./syn/yosys/run.sh …
YOSYS_MALLOC=/etc/hosts: the loader will not preload it (/etc/hosts)
  ERROR: ld.so: object '/etc/hosts' from LD_PRELOAD cannot be preloaded (invalid ELF header): ignored.
rc=2
$ YOSYS_MALLOC=/no/such/lib.so ./syn/yosys/run.sh …
YOSYS_MALLOC=/no/such/lib.so: no such file
rc=2
```

## P2 — `none` declined to set `LD_PRELOAD` instead of unsetting it

Reproduced with a preload shim that records `program_invocation_short_name` for every process it enters:

```
$ LD_PRELOAD=<probe.so> YOSYS_MALLOC=none ./syn/yosys/run.sh --top cdc_pulse --no-structural
   yosys allocator: system
  processes that inherited it:  abc awk bash dirname env git mktemp python3(5) rm sh sv2v yosys
```

`yosys` and `abc` in that list is the bug: the header said `system` and it was not. Now:

```
  processes that inherited it:  awk bash dirname env git mktemp python3(5) rm sv2v
  --> yosys present? 0   abc present? 0
```

The decision moved into one function, `apply_malloc_env`, which the per-top subshell and the self-test both call, so the test cannot drift from the code:

```sh
apply_malloc_env() {
  local lib="${1-}"
  if [ -n "$lib" ]; then export LD_PRELOAD="$lib"; else unset LD_PRELOAD; fi
}
```

`sv2v` and the `python3` helpers still inherit whatever the caller exports — deliberately; the gate only claims to control the allocator `yosys` runs under, and that is now what the README says.

## Self-test

You were right that none of this was covered. `--selftest-alloc` now has arms for: `none` selects nothing; an **existing but unloadable** file is refused; a missing file is refused; the default never names something the loader would drop; a **relative** request comes back absolute; an inherited `LD_PRELOAD` does not survive an empty selection; and a selected library does reach the child environment. The relative-path and usable-default arms **announce a SKIP** on a machine with no preloadable jemalloc rather than passing vacuously.

## Re-validation

The full 46-top gate, eight shards, run twice — and the `none` side deliberately run **with an ambient `LD_PRELOAD=libtcmalloc_minimal.so.4`**, which is the P2 case at full scale:

```
with:    46 tops 46 pass 0 fail   yosys allocator: /usr/lib/libjemalloc.so.2
without: 46 tops 46 pass 0 fail   yosys allocator: system
diff -r  ->  IDENTICAL
```

Plus `--selftest-alloc`, `bash -n`, `lint_rtl --check`, `pp_srcs --check`, `yosys_shards --selftest`, `docs_check`, `gen_toc --check`, `check_baremetal_only --check/--selftest`, and the no-git docs shape.

No measurement in the PR body is affected: my shell had no ambient `LD_PRELOAD` (checked), and every quoted figure came from an explicit absolute path or from the default.

Issue #190, opened 2026-08-21
dev is red: the yosys-portability aggregate cannot run run.sh --list without the submodule it says it does not need


Mister-M-alt:
`dev` is red. The `yosys-portability` aggregate has failed on every run since PR #176 merged, in 7 seconds, before it tallies anything.

## The failure

```
/home/runner/work/milan-fpga/milan-fpga/protocol-processor/hdl not found.
The protocol-processor submodule is not checked out; run `git submodule update --init protocol-processor`.
This script refuses to emit an empty list, because an empty list builds cleanly and proves nothing.
##[error]Process completed with exit code 2.
```

Observed on run `32473235503`, job `yosys-portability`, step "Reconcile the live inventory and structural gates". **All four Yosys shards passed**; only the aggregate fails.

## Cause, and it is two things that only bite together

1. **The aggregate job never checks out the submodule.** `.github/workflows/rtl.yml`'s `yosys-portability` job runs `actions/checkout@v4`, `actions/download-artifact`, then `syn/yosys/run.sh --list`. The four shard jobs do `git submodule update --init third_party/verilog-axis protocol-processor gptp-processor`; the aggregate does not, because it was not thought to need sources. It only wants top NAMES.

2. **`run.sh --list` needs the submodule anyway, contrary to its own documented contract.** `syn/yosys/run.sh:89` runs `PP_DERIVED="$(python3 "$R/scripts/pp_srcs.py" --prefix "$PP")" || exit 2` unconditionally, and `--list` is not handled until line 194. Its own header says the opposite in two places:

   - line 14: `# Requires yosys and sv2v except for --list.`
   - line 24: `--list                 print selected top names without requiring tools`

   Both were true before the source list became derived. `pp_srcs.py` is not a "tool" in the sense line 24 means, so nobody re-read that sentence when the derivation landed.

This is mine as much as #176's: I made `run.sh` derive its sources through `pp_srcs.py` in #175, and `--list` silently inherited a requirement its documentation denies. #176 is the first caller to use `--list` from a job without submodules, so it is the first to hit it.

## Why no gate caught it

The aggregate is correctly fail-closed for evidence problems, which is what it was reviewed for. Nothing exercises the job's own environment, and `--list` is not called by the local serial path in a submodule-free tree, so the only place this shape exists is that one CI job.

## The fix

Two options, and the first is better because it makes an existing false statement true rather than working around it:

[...]

PR #284, opened 2026-08-27
[Code quality] Rule 10: say what the block is

Mister-M-alt:
> [!IMPORTANT]
> **Post-review correction (2026-08-28, head `e70fd6ad`).** SystemVerilog idiom enforcement covers both processor submodules, refuses ordinary `.v`, catches ANSI `output reg`, and ignores comments/strings correctly. Result: 120 tracked HDL files, zero ordinary `.v`, zero generic `always`, 58 legacy regs, zero untyped parameters; self-test 22/22. Earlier figures below are superseded.

Closes #258

Rule 10 of the #248 contract — **the last one**. Stacked on #283 (Rule 9) → #282 → #281 → #280 → #279 → #278 → #277 → #276 → #275.

## What changed

| Area | Change |
|---|---|
| Guide | Rule 10 section, and the guide's closing note now that all ten rules have landed. |
| Gate | `scripts/check_sv_idiom.py` + `scripts/sv_idiom.budget`. |
| Cleanup | The tree's one generic `always @`, converted. |
| CI | The gate and its self-test run in `docs.yml`. |

## One refusal, two ratchets

| Finding | Count | Disposition |
|---|---:|---|
| Generic `always @` in synthesizable first-party HDL | **0** | Refused outright |
| `reg` declarations | 48 across 12 modules | Ratchet |
| Untyped `parameter` | 4, all in `axi_stream_if.sv` | Ratchet |
| `.v` files in the gated set | 0 | — |

A generic `always @` is the only construct here that discards a real check: `always_ff` asks every tool in the path to enforce a single driver and to reject a block that is not sequential; `always @` asks for neither, and the difference is silent.

## A `wire` is deliberately not a finding

A net type is the correct model for module, primitive, continuous-assignment and multiple-driver connectivity. Mechanically rewriting `wire` to `logic` would touch nearly every file and change nothing a reader or a tool can use. The issue that asked for this rule says so in as many words. The gate does not look for it, and the guide records that decision rather than leaving it as an omission.

## The one generic `always`, and why converting it was delicate

`hdl/common/csr/milan_csr.sv` carried exactly one — the **reset-epoch canary**: two flops that count datapath reset *releases*, and therefore must have no reset clause at all so their bitstream initialisation survives every axis reset. Adding a reset would silently destroy the only thing they measure. This is precisely the `reset_less` class of observer `CONTRIBUTING.md` warns about.

[...]

PR #275, opened 2026-08-27
[Code quality] Rule 1: cut the LTAP observation adapter out of milan_datapath

Mister-M-alt:
> [!IMPORTANT]
> **Post-review correction (2026-08-28, head `d6e9d3e8`).** The cohesion measurement now scans the superproject plus pinned `protocol-processor` and `gptp-processor`, refuses absent/off-pin processor checkouts, and measures each module separately in a multi-module file. Current Rule 1 result: 110 modules, 106 with procedural blocks, 55 multi-group; self-test 18/18. Earlier figures below predate this correction; this block is authoritative where they differ.

Closes #249

Rule 1 of the #248 contract, landed with the guide it needs, the measurement that picked its target, and one representative cleanup.

## What changed

| Area | Change |
|---|---|
| Guide | `docs/development/CODE_QUALITY.md` — new. Governing (Boy Scout) rule, scope, Rule 1 wording, repository interpretation, worked example, counter-example, exceptions, review checklist. Linked from `CONTRIBUTING.md` and the docs index. |
| Measurement | `scripts/measure_cohesion.py` — disjoint state groups, not LOC. Two `always_ff` blocks are one owner when they share a written signal or one reads what the other writes; coupling is transitive. |
| Cleanup | The AAF latency-tap observation adapter moved out of `hdl/milan/milan_datapath.sv` into `hdl/ieee1722/aaf/KL_aaf_latency_tap_bank.sv`. |
| Tests | `tb/verilator/aaf_latency_tap_bank/` — two builds, 39 + 7 checks, graded through the AXIS observation points rather than the pre-derived stage vector. |
| Gate | `sw/builder/test_builder.py` gate 23d now follows a prune parameter one hop, into the child module it is bound to. |

## Why this seam

`milan_datapath.sv` owns 22 disjoint state groups. The tap adapter is one of them and it has a contract a test can grade: AXIS observation points in, the sixteen LTAP words at base `0x870` out. Every port is an input except the two register rails, which makes the observer-purity property `syn/yosys/check_tap_purity.sh` checks structural rather than an argument about a 7,000-line body.

| | Lines | `always_ff` | Disjoint state groups |
|---|---:|---:|---:|
| `milan_datapath.sv` before | 7238 | 25 | 22 |
| `milan_datapath.sv` after | 7152 | 23 | 20 |

## Evidence

All run at this head.

- `scripts/lint_rtl.py --check` — **PASS**, 99 violations ≤ ratchet 99. No new lint.
[...]

PR #291, opened 2026-08-28
Run syn/yosys/ooc.sh under the same allocator, and put the rules in one file

### LINKED ISSUE #290: [TASK] Run syn/yosys/ooc.sh under the same allocator as the portability gate
## Objective / problem

#288 / PR #289 put the gate's `yosys` children under a jemalloc preload when
the library is installed. `syn/yosys/ooc.sh` is the repository's other Yosys
consumer and was deliberately left out of that change, one script at a time.
Bring it under the same knob, with the same properties, and no change to what
it measures.

## Evidence / motivation

`ooc.sh` runs the heavier flow — `synth_xilinx -family xc7 -flatten` — so it
gains more than the portability gate did. Measured on one machine, `yosys
0.66`, same head, the 1722.1 control plane as `KL_pp_shadow`:

| | wall |
| --- | ---: |
| `./syn/yosys/ooc.sh KL_pp_shadow` | 241.51 s |
| same, under `LD_PRELOAD=/usr/lib/libjemalloc.so.2` | **171.00 s** (−29.2%) |

The area figures — which are the entire point of this script — are identical
in both runs:

```
KL_pp_shadow 58209 6008 64217 28937 15 4 4 2019
```

That is consistent with #286's netlist proof: `write_rtlil` after the gate's
own program is byte-identical under glibc, tcmalloc, jemalloc and mimalloc.

## Authoritative references

- #288 / PR #289 — the merged mechanism this reuses, and the review that
  established its safety properties
- #286 — the measurement and the netlist-invariance proof
- `syn/yosys/run.sh` — `select_malloc`, `apply_malloc_env`, `--selftest-alloc`
- `syn/yosys/ooc.sh` — the yosys invocation at its per-top run directory
- `syn/yosys/README.md` — "Runtime levers"

## Context entry points

Read first:

- `syn/yosys/run.sh` (the allocator block and its self-test)
- `syn/yosys/ooc.sh`
- `syn/yosys/README.md`
- #288, #286

## Scope

In scope:

- Select an allocator in `ooc.sh` on the same `YOSYS_MALLOC` contract as
  `run.sh`: an explicit path is refused when absent or when the loader will
  not take it; a relative path is made absolute before any directory change;
  `none` unsets `LD_PRELOAD` rather than declining to set it; an
  auto-detected candidate that will not load is skipped.
- Apply it to the `yosys` invocation only, leaving `sv2v` and the `python3`
  ROM generators alone, exactly as `run.sh` does.
- Name the allocator in the script's own output so an area run's wall time is
  attributable.
- A self-test on the same shape as `run.sh --selftest-alloc`, runnable without
  jemalloc.
- Prove the area numbers are unchanged, per top, with and without.

Out of scope:

- Changing the tops list, the ROM digest discipline, the exclusive read-only
  image copies, the refusal contract of #245, or any reported figure.
- The protocol-processor gate — its own repository, its own PR.
- Adding the allocator to any CI workflow: CI's Yosys is still unpinned
  (#287), and that ordering has not changed.

## Executor

Agent A [A1]

## Independent reviewer

Unassigned

## Acceptance criteria

- [ ] `ooc.sh` runs `yosys` under jemalloc when the library is present and
      unchanged when it is not; a machine without it sees no error.
- [ ] `YOSYS_MALLOC=none`, `YOSYS_MALLOC=<path>` and the absent-library default
      behave as they do in `run.sh`, including the refusal cases.
- [ ] `YOSYS_MALLOC=none` clears an inherited `LD_PRELOAD` for the `yosys`
      child; proved, not asserted.
- [ ] The preload reaches `yosys` and not `sv2v` or `python3`.
- [ ] Per-top area figures are byte-identical with and without the preload, for
      every top the PR reports, and the digests/records are in the PR.
- [ ] A self-test covers the selection rules and runs on a machine with no
      jemalloc, announcing skips rather than passing vacuously.
- [ ] `syn/yosys/run.sh` (all shards), the tied-input gate, `lint_rtl --check`,
      `pp_srcs --check`, `yosys_shards --selftest`, `docs_check`,
      `gen_toc --check` and `check_baremetal_only --check/--selftest` stay green.

## Validation

```sh
syn/yosys/ooc.sh --selftest-alloc
./syn/yosys/ooc.sh KL_pp_shadow                       # under the default
YOSYS_MALLOC=none ./syn/yosys/ooc.sh KL_pp_shadow     # under glibc
# the reported area rows must match line for line
```

Expected result: identical area rows both ways, a wall-time reduction on the
order measured above, and every listed gate green.

## Dependencies / lane conflicts

Depends on:

- #288 / PR #289, merged, for the mechanism and its reviewed properties.

May conflict with:

- concurrent edits to `syn/yosys/ooc.sh` or `syn/ooc/dp_srcs.py`.
- #287, which will pin the toolchain these figures were taken against.

## Open design/specification decision?

No — scope is settled and may become Ready

## Decision details (if applicable)

None. This applies an already-decided lever (#286 decision 1, option 1) to the
second consumer, on the contract PR #289 already landed and a reviewer already
exercised.

## Discovered from

#288, and a follow-up sweep of the remaining Yosys consumers after PR #289
merged.

Mister-M-alt:
Closes #290.

`syn/yosys/ooc.sh` now runs its `yosys` under the same `YOSYS_MALLOC` contract
PR #289 landed for the portability gate, and the rules moved into
`syn/yosys/malloc.sh` so the second consumer does not carry a copy.

## Measured

`./syn/yosys/ooc.sh KL_pp_shadow KL_crf_rx tcam`, one machine, `yosys 0.66`:

| | wall |
| --- | ---: |
| `YOSYS_MALLOC=none` | 235.09 s |
| default (jemalloc found) | **177.85 s** (−24.3%) |

`synth_xilinx -flatten` is the heavier program, so this flow gains more than
the portability gate did.

## The area rows do not move

They are the entire point of this script, so they are the acceptance test, and
they are identical across all three tops:

```
KL_crf_rx        355     0    355    509    0    1    0   125
KL_pp_shadow   58209  6008  64217  28937   15    4    4  2019
tcam             678     0    678   1680    0    0    0     0

diff <(rows under jemalloc) <(rows under none)  ->  IDENTICAL
```

Consistent with #286, where `write_rtlil` after the gate's own program was
byte-identical under glibc, tcmalloc, jemalloc and mimalloc.

## One file, not two copies

`syn/yosys/malloc.sh` holds the selection, the child-environment decision and
the self-test; `run.sh` and `ooc.sh` both source it. It defines functions and
sets one variable, starts nothing and changes no directory.

**`run.sh --list` does not source it**, deliberately and with a comment saying
why: `check_list_hermetic.sh` proves `--list` reads nothing but `run.sh` and
`scripts/yosys_shards.py`, in a tree built to hold only those two files (#190).
An unconditional `source` would have broken that contract, so the source is
guarded and `--selftest-alloc` and `--list` are now mutually exclusive, on the
same grounds as `--emit` and `--list`. `check_list_hermetic.sh` passes,
including its negative control.

## Properties carried over from #289's review

Because they were defects there first:

- an explicitly named library is **refused** (exit 2) if absent, or if the
  loader will not take it — the loader rejects an unusable `LD_PRELOAD` by
  *ignoring* it, exiting 0 with the complaint on stderr, so the test is
  "stderr stayed empty";
- a relative `YOSYS_MALLOC` is made absolute before it is used;
- `YOSYS_MALLOC=none` **unsets** `LD_PRELOAD` rather than declining to set it;
- an auto-detected candidate the loader will not take is skipped, not refused.

Proved for `ooc.sh` with a shim that records `program_invocation_short_name`:

```
$ LD_PRELOAD=<probe.so> YOSYS_MALLOC=none ./syn/yosys/ooc.sh tcam
   yosys allocator: system
   yosys=0  abc=0  sv2v=1
```

`sv2v` runs *inside* this script's per-top loop, which is why the preload is
applied in the `yosys` subshell rather than exported once — the ROM generators
and `sv2v` keep the caller's environment, as the README says.

## The self-test anchors moved, and I want that flagged

`ooc_selftest.py` plants its mutations by matching `ooc.sh`'s text. Two arms
anchor on `(cd "$rundir" && yosys`, which is now
`(cd "$rundir" && apply_malloc_env "$MALLOC_LIB" && yosys`. Both anchors were
updated; **neither mutation changed in meaning** — `cwd-escape` still runs the
synthesis outside the exclusive per-top directory and `consume-shared-dir`
still moves it back to the shared published one, and both must still be caught.
The suite refused loudly rather than passing when the anchor moved, which is
the behaviour you want from it. 75/75 arms pass (75 since [A1] added the allocator-scoping arms).

## Validation

```
ooc.sh refusal self-test: 75 arm(s) passed
allocator selection self-test: PASS            (run.sh and ooc.sh)
check_list_hermetic: OK                        (positive and negative control)
dp_srcs self-test: 35 arm(s) passed
LINT GATE: PASS (99 <= ratchet 99)
docs_check: 0 finding(s) across 181 md + 786 scrubbed text files
baremetal-only: OK (0 findings across 784 tracked first-party file(s))
TOC gate: OK
yosys_shards --selftest: PASS
pp_srcs --check: rc=0
```

The two file-census gates were re-run **after** `git add` of the new file, so
their counts include it (783→784 tracked, 785→786 scrubbed).

CI is untouched: its Yosys is still unpinned (#287), and that ordering has not
changed.

### DIFF
--- a/syn/yosys/README.md
+++ b/syn/yosys/README.md
@@ -110,6 +110,15 @@ figure quoted from this output is only reproducible with it. The preload
 reaches `yosys` and the `abc` it spawns; `sv2v` and the `python3` helpers are
 left alone and keep whatever the caller's environment gives them.
 
+The rules live in one place, [`malloc.sh`](malloc.sh), and `ooc.sh` below
+sources the same file and honours the same `YOSYS_MALLOC` values. That flow
+gains more, because `synth_xilinx -flatten` is the heavier program — 235.09 s
+to 177.85 s over `KL_pp_shadow`, `KL_crf_rx` and `tcam` together (−24.3%),
+with every reported area row identical to the digit. `run.sh --list` does
+**not** source it: `check_list_hermetic.sh` proves `--list` reads nothing but
+`run.sh` and `scripts/yosys_shards.py`, and that contract outranks the tidiness
+of an unconditional `source`.
+
 ### `read_verilog -defer`: right instrument, wrong script
 
 `-defer` postpones turning a parsed module into RTLIL until something
--- a/syn/yosys/malloc.sh
+++ b/syn/yosys/malloc.sh
@@ -1,0 +1,187 @@
+# SPDX-FileCopyrightText: 2026 Kebag Logic
+# SPDX-License-Identifier: CERN-OHL-W-2.0
+#
+# Allocator selection for this directory's Yosys flows, sourced by run.sh
+# (portability) and ooc.sh (area). It defines functions and sets one variable
+# (TRUE_BIN); it starts nothing and changes no directory, so a caller may
+# source it before or after its own `cd`.
+#
+# It is NOT sourced on run.sh's `--list` path. syn/yosys/check_list_hermetic.sh
+# proves `--list` reads nothing but run.sh and scripts/yosys_shards.py, in a
+# tree built to hold only those two files (#190), and an unconditional source
+# here would break that contract rather than the other way round.
+
+# THESE FLOWS ARE ALLOCATION-BOUND, NOT PARSE-BOUND (#286, #288). On the heaviest
+# top Yosys spends about three quarters of its time in the `opt*` family and
+# 0.5% in `read_verilog`, and a large share of that is glibc's allocator:
+# preloading jemalloc takes milan_datapath from 656 s to 361 s and the whole
+# eight-way sharded gate from 616 s to 358 s, on the same machine, with all 46
+# tops still passing.
+#
+# It is a SPEED knob and never a correctness one: `write_rtlil` after this
+# script's own program is BYTE-IDENTICAL under glibc, tcmalloc, jemalloc and
+# mimalloc (#286). So it is optional in both directions - the documented tool
+# requirement stays "yosys and sv2v on PATH" (README "Tooling") and a machine
+# without jemalloc runs exactly as it did before, with no warning to silence.
+#
+#   YOSYS_MALLOC=<path>   preload that library
+#   YOSYS_MALLOC=none     run yosys under the system allocator
+#   unset                 use jemalloc when it is installed
+#
+# A NAMED library that is absent REFUSES rather than falling back. Falling back
+# would let a run that was asked to reproduce one allocator quietly report
+# another one's timing, which is the measurement defect this whole lane exists
+# to remove.
+ldconfig_path() {
+  local p
+  for p in ldconfig /usr/sbin/ldconfig /sbin/ldconfig; do
+    command -v "$p" >/dev/null 2>&1 && { printf '%s\n' "$p"; return 0; }
+  done
+  return 1
+}
+
+# EVERY PATH IS MADE ABSOLUTE BEFORE IT IS BELIEVED. yosys runs after `cd
+# "$TMP"`, so a relative LD_PRELOAD the caller checked from their own directory
+# resolves against $TMP instead, where it is not there: the loader drops it and
+# the header line reports an allocator that never ran.
+abs_path() {
+  local p="$1" out d b
+  if out="$(readlink -f -- "$p" 2>/dev/null)" && [ -n "$out" ]; then
+    printf '%s\n' "$out"
+    return 0
+  fi
+  d="$(cd -- "$(dirname -- "$p")" 2>/dev/null && pwd)" || return 1
+  b="$(basename -- "$p")"
+  printf '%s/%s\n' "$d" "$b"
+}
+
+# EXISTING IS NOT LOADABLE. The loader takes an unusable LD_PRELOAD by IGNORING
+# it: the process still exits 0 and the complaint goes to stderr, which for a
+# yosys child lands in that top's log and is never read - so `/etc/hosts` was
+# accepted, the gate passed, and the header named it. The test is therefore
+# "stderr stayed empty", never "the command succeeded".
+TRUE_BIN="$(type -P true 2>/dev/null || printf '')"
+preload_is_usable() {
+  local lib="$1" err
+  [ -n "$TRUE_BIN" ] && [ -x "$TRUE_BIN" ] || return 1
+  err="$(LD_PRELOAD="$lib" "$TRUE_BIN" 2>&1 >/dev/null)" || return 1
+  [ -z "$err" ]
+}
+preload_refusal() {
+  [ -n "$TRUE_BIN" ] && [ -x "$TRUE_BIN" ] || { printf 'no external true(1) to probe with\n'; return 0; }
+  LD_PRELOAD="$1" "$TRUE_BIN" 2>&1 >/dev/null | head -1
+}
+
+select_malloc() {
+  local want="${YOSYS_MALLOC-}" cand lc abs
+  case "$want" in
+    none) return 0 ;;
+    "")   ;;
+    *)    # Existence is checked BEFORE canonicalisation so a relative request
+          # is judged against the caller's directory, which is where they meant
+          # it, and so the message says what is actually wrong.
+          [ -e "$want" ] || {
+            echo "YOSYS_MALLOC=$want: no such file" >&2
+            return 2
+          }
+          abs="$(abs_path "$want")" || {
+            echo "YOSYS_MALLOC=$want: cannot be resolved to an absolute path" >&2
+            return 2
+          }
+          preload_is_usable "$abs" || {
+            echo "YOSYS_MALLOC=$want: the loader will not preload it ($abs)" >&2
+            echo "  $(preload_refusal "$abs")" >&2
+            return 2
+          }
+          printf '%s\n' "$abs"
+          return 0 ;;
+  esac
+  # ldconfig answers for the loader's own search path, which is the only
+  # authority on where a distribution put the library - multiarch layouts
+  # included, which is why no such path is spelled below. The literal list is
+  # only the fallback for a machine whose ldconfig a normal user cannot run,
+  # and an empty answer is a valid one: the gate then runs unchanged.
+  # A candidate that the loader will not take is SKIPPED here rather than
+  # refused: the documented default is "use jemalloc when it is installed", and
+  # a broken one is simply not installed. An explicit request is refused above,
+  # because there the caller named it and is owed an answer.
+  if lc="$(ldconfig_path)"; then
+    while IFS= read -r cand; do
+      [ -e "$cand" ] && preload_is_usable "$cand" && { printf '%s\n' "$cand"; return 0; }
+    done < <("$lc" -p 2>/dev/null | sed -n 's/^.* => //p' | grep -F libjemalloc.so.2)
+  fi
+  for cand in /usr/lib/libjemalloc.so.2 /usr/lib64/libjemalloc.so.2 \
+              /usr/local/lib/libjemalloc.so.2; do
+    [ -e "$cand" ] && preload_is_usable "$cand" && { printf '%s\n' "$cand"; return 0; }
+  done
+  return 0
+}
+
+# THE ONLY PLACE that decides what a yosys child sees, so the self-test below
+# exercises the same code the gate runs. An EMPTY selection must actively UNSET
+# LD_PRELOAD rather than merely decline to set it: a caller who already exported
+# one would otherwise have it inherited while the header line said "system",
+# which is the same silent misattribution the refusal above exists to prevent -
+# and it would invalidate any with-versus-without comparison run in that shell.
+apply_malloc_env() {
+  local lib="${1-}"
+  if [ -n "$lib" ]; then
+    export LD_PRELOAD="$lib"
+  else
+    unset LD_PRELOAD
+  fi
+}
+
+# The interesting case is the machine WITHOUT jemalloc, because that is the one
+# the documented tool requirement covers and the one nobody runs by accident.
+# This self-test needs neither jemalloc, nor yosys, nor sv2v, nor a submodule.
+selftest_alloc() {
+  local rc=0 skipped=0 out probe found dir base
+
+  out="$(YOSYS_MALLOC=none select_malloc)" || rc=1
+  [ -z "$out" ] || { echo "selftest: YOSYS_MALLOC=none selected '$out'" >&2; rc=1; }
+
+  # An EXISTING file the loader will not take must be refused, not reported as
+  # the allocator in effect. An empty temp file is exactly that case.
+  probe="$(mktemp)"
+  if out="$(YOSYS_MALLOC="$probe" select_malloc 2>/dev/null)"; then
+    echo "selftest: an unloadable YOSYS_MALLOC was accepted as '$out'" >&2; rc=1
+  fi
+  if out="$(YOSYS_MALLOC="$probe.absent" select_malloc 2>/dev/null)"; then
+    echo "selftest: a missing YOSYS_MALLOC was accepted as '$out'" >&2; rc=1
+  fi
+  rm -f "$probe"
+
+  # The default never names something the loader would drop.
+  found="$(unset YOSYS_MALLOC; select_malloc)" || rc=1
+  if [ -n "$found" ]; then
+    { [ -e "$found" ] && preload_is_usable "$found"; } || {
+      echo "selftest: the default selected an unusable '$found'" >&2; rc=1
+    }
+    # A RELATIVE request must come back absolute, because yosys runs from $TMP.
+    dir="$(dirname -- "$found")"; base="$(basename -- "$found")"
+    out="$(cd "$dir" && YOSYS_MALLOC="./$base" select_malloc)" || rc=1
+    [ "$out" = "$found" ] || {
+      echo "selftest: relative path resolved to '$out', expected '$found'" >&2; rc=1
+    }
+  else
+    echo "selftest: SKIPPED the relative-path and usable-default arms" \
+         "(no preloadable jemalloc on this machine)"
+    skipped=1
+  fi
+
+  # The child environment, through the same function the per-top run uses.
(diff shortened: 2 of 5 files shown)


Mister-M-alt (commented):
**[R1] Cold-context review** — head `e6ed004e` (branch `290-task-run-synyosysoocsh-under-the-same-allocator-as-the-portability-gate`, base `dev` @ 88f58ba6; also merged locally onto today's `dev` b5b4359c, where every code-quality gate of #275–#284 stays green). Reproduced on one machine: Yosys 0.66, jemalloc at `/usr/lib/libjemalloc.so.2`, GNU make 4.4.

## Findings

**[R1] MAJOR Tests, Conformance — syn/yosys/ooc.sh:562-567; syn/yosys/ooc_selftest.py:897,1066; syn/yosys/malloc.sh `selftest_alloc` — the two properties issue #290 says must be "proved, not asserted" have no regression arm**

Requirement/evidence: #290 acceptance criteria: "`YOSYS_MALLOC=none` clears an inherited `LD_PRELOAD` for the `yosys` child; proved, not asserted" and "The preload reaches `yosys` and not `sv2v` or `python3`". The PR proves both once, by hand, with a shim (PR body). I reproduced that at the head with my own shim (a constructor recording `program_invocation_short_name`): `YOSYS_MALLOC=none` with an inherited `LD_PRELOAD` → `yosys`/`abc` absent from the record, `sv2v=1 python3=10 sha256sum=9 …` keep it; `YOSYS_MALLOC=<probe.so>` → `yosys=1 abc=1 sh=1` only. Then I planted the obvious regression — one line, `apply_malloc_env "$MALLOC_LIB"` at top level directly after `MALLOC_LIB="$(select_malloc)" || exit 2`, the subshell call left untouched — and the preload now reaches `sha256sum=9 python3=9 awk=7 rm=6 … sv2v=1` as well, while `syn/yosys/ooc.sh --selftest-alloc` prints `allocator selection self-test: PASS` and `python3 syn/yosys/ooc_selftest.py` prints `ooc.sh refusal self-test: 72 arm(s) passed`. `selftest_alloc` grades the two functions in isolation; the two anchored arms grade only the subshell's working directory.

Impact: the scoping is held by the code's shape and a comment. The first "simplification" that exports the preload once puts `sv2v` and the ROM generators under an allocator nobody measured them with — the exact misattribution class this lane exists to remove — and every gate stays green.

Required change: an arm that bites, runnable without a compiler or jemalloc. Run `ooc.sh` on the cheapest top (`tcam`, ~4 s) with `YOSYS_MALLOC=none` and an inherited `LD_PRELOAD=/nonexistent.so`: the loader prints `ERROR: ld.so: object '/nonexistent.so' from LD_PRELOAD cannot be preloaded` on the stderr of every child that inherits it, so assert that `$TMP/tcam.ooc.log` (the `yosys` child's stderr) carries zero such lines while the script's own stderr carries at least one (`sv2v`/`python3` kept it). For "reaches yosys", `LD_DEBUG=libs` with the selected library when a usable one exists, skip announced otherwise — or the shim, if `ooc_selftest.py` may assume a C compiler. Prove the arm bites with the top-level-export mutant above.

Verification: the mutant fails the new arm and the head passes it; `ooc_selftest.py` reports 73+ arms.

**[R1] MINOR Docs, Conformance — syn/yosys/ooc.sh:565 — the comment names the anchor shape this PR moved away from**

Requirement/evidence: the comment reads "The `(cd "$rundir" && yosys` shape is load-bearing - the cwd-escape arm of ooc_selftest.py plants its mutation on exactly that text", but the line below is now `(cd "$rundir" && apply_malloc_env "$MALLOC_LIB" && yosys …` and both arms anchor on `(cd "$rundir" && apply_malloc_env` (ooc_selftest.py:897, :1066) — the PR body describes this move itself.

Impact: that comment is what the next editor reads before touching the line; it sends them to preserve text that no longer exists, so the anchor moves again and the suite refuses (loudly, as the PR notes — but the comment is there to prevent that trip, not to cause it).

Required change: name the actual anchor text.

Verification: `grep -n 'apply_malloc_env' syn/yosys/ooc.sh syn/yosys/ooc_selftest.py` agree.

**[R1] SUGGESTION Tests — syn/yosys/malloc.sh `selftest_alloc` — the relative-path and usable-default arms skip on a jemalloc-less machine although any preloadable library would serve them**: `libc.so.6` (via `ldconfig -p`) is always preloadable, so the relative→absolute arm can run everywhere and the skip can be narrowed to the jemalloc-specific default. Skips are announced today, so this is not vacuous — just narrower than it needs to be.

**[R1] SUGGESTION Robustness — syn/yosys/malloc.sh `preload_is_usable` — "stderr stayed empty" also rejects a library that loads and merely warns** (a debug jemalloc, a `MALLOC_CONF` message): the auto-detect path then skips it silently and the explicit path refuses it with the loader's line, which is the right diagnostic; name the limit in the comment.

## Clean lenses

**[R1] PASS Conformance (issue #290 criteria)** — default selects jemalloc when present (`yosys allocator: /usr/lib/libjemalloc.so.2` in the header) and `none` → `system`; `YOSYS_MALLOC=<absent>` → `no such file`, exit 2; an existing empty file and `/etc/hosts` → `the loader will not preload it` plus the loader's own `ERROR: ld.so: …` line, exit 2; `YOSYS_MALLOC=./probe.so` from another directory resolves to the absolute path and reaches `yosys`; the self-test runs with the submodules deinitialised and announces its skips; CI untouched (`rtl-fast.yml` runs `ooc_selftest.py` only, allocator not added); tops list, ROM digest discipline and the #245 refusal untouched (72/72 arms).

**[R1] PASS RTL / measurement** — no RTL. Area rows byte-identical under jemalloc and the system allocator on every top the PR reports, re-measured here: `tcam 678 0 678 1680 0 0 0 0`; `KL_crf_rx 355 0 355 509 0 1 0 125`; `KL_pp_shadow 58209 6008 64217 28937 15 4 4 2019`. Wall on `KL_pp_shadow`: 169 s (jemalloc) vs 234 s (system), −27.8%, consistent with the PR's −24.3% over three tops.

**[R1] PASS Robustness** — `run.sh --list` output identical to `dev`'s (46 lines) and `check_list_hermetic.sh: OK` including its negative control; `run.sh --selftest-alloc --list` → `mutually exclusive`, exit 2; `malloc.sh` is sourced under `set -u` by both scripts, defines functions and one variable, starts nothing (`type -P`); `ooc.sh`'s main shell never changes directory (grep), so the caller-relative existence check holds; `bash -n` clean on all three scripts.

**[R1] PASS Docs** — README "Runtime levers", the `ooc.sh` header block and `run.sh` usage agree with each other and with the measurements; `docs_check` 0 findings, `gen_toc --check` OK, `check_doc_paths` OK, `ci_events --check` OK (230), `pp_srcs --check` OK, `yosys_shards --selftest` PASS, `lint_rtl --check` PASS (99 ≤ 99).

**[R1] PASS Integration with today's `dev`** — merged locally onto b5b4359c (fc8218d3, no conflicts beyond auto-merge): measure_cohesion 18/18, measure_control_flow 45/45, check_rtl_source_lists OK + 49/49, measure_naming 108 recorded + 66/66, check_port_contracts OK + 80/80, measure_fail_fast PASS (85/4/2) + 106/106, suite_tally PASS, check_todo_ownership OK + 45/45, measure_test_evidence PASS + 93/93, check_hygiene PASS (3 populations) + 35/35, check_sv_idiom OK + 54/54, docs gates (with and without submodules) green, lint 99 ≤ 99, check_soc_sources OK, tap purity PASS, both `--selftest-alloc` PASS, hermetic OK.

Verdict: mergeable once the MAJOR has its arm; the MINOR is a one-line comment fix.

Mister-M-alt:
**[A1] Answer to review [R1]** — head `e6ed004e` → `63dab47e` (one commit on the PR branch, one-line message, no trailers).

**[R1] MAJOR Tests, Conformance — no regression arm for the preload's scope — FIXED (`63dab47e`).** `ooc_selftest.py` now carries the shim as arms, without a compiler and without jemalloc: the honest `yosys` and `sv2v` models gain twins (`YOSYS_ENV`, `SV2V_ENV`) that record the `LD_PRELOAD` they ran under into `$HOME`, and `libc.so.6` — which every dynamically linked process already maps, so the loader preloads it anywhere — is the fixture library (found through `ldconfig -p`, verified with `true` the way `malloc.sh` does; if no preloadable library exists the three arms are announced as SKIPPED, never passed vacuously).
- Arm 73 `alloc-scoped-to-yosys`: `YOSYS_MALLOC=<libc>` → the yosys model records libc, the sv2v model records `<unset>`, the header names the library.
- Arm 74 `alloc-none-clears-child`: `YOSYS_MALLOC=none` with `LD_PRELOAD=<libc>` inherited → the yosys model records `<unset>`, the sv2v model still records libc (the caller's own environment is kept), the header says `system`.
- Arm 75 `mut-alloc-exported-once`: the regression the review planted (`apply_malloc_env "$MALLOC_LIB"` at top level, subshell untouched) as a `_mutant`, required to be caught by arm 73's check.

Before/after: at `e6ed004e` the export-once mutant passed `--selftest-alloc` and `72 arm(s) passed`; at `63dab47e` the same mutant fails with `SELF-TEST FAILED [alloc-scoped-to-yosys]: the preload leaked to sv2v: ['/usr/lib/libc.so.6']` (and arm 74 with `sv2v lost the caller's own environment`), and a `malloc.sh` that declines to unset fails with `[alloc-none-clears-child]: YOSYS_MALLOC=none left the inherited preload on yosys`. Clean tree: `ooc.sh refusal self-test: 75 arm(s) passed`. The PR body's "72/72" is now 75/75.

**[R1] MINOR Docs — stale anchor comment at `ooc.sh:565` — FIXED (`63dab47e`).** The comment now names the anchor the arms actually use — and does so without quoting it verbatim, because the first rewording that did quote it made the mutant pattern hit twice and the suite refused the run (`mutation 'cwd-escape': pattern hit 2 times`); the comment says so, so the next editor does not repeat it.

**SUGGESTIONs** (relative-path arm via libc; "any stderr = unusable" as a named limit): not taken in this round — both are `malloc.sh` refinements the maintainer may want to weigh against #289's reviewed wording; noted here so they are not lost.

Also caught on the way: a multiarch fallback path in the new helper carried a retired-stack term and tripped `check_baremetal_only` — dropped, `ldconfig` answers for multiarch layouts as `malloc.sh` already says.

Gates at `63dab47e` (all rc 0): `ooc_selftest.py` 75/75; `ooc.sh --selftest-alloc` and `run.sh --selftest-alloc` PASS; `check_list_hermetic` OK; `docs_check` 0; `gen_toc --check` OK; `check_doc_paths` OK; `ci_events --check` OK (230); `pp_srcs --check` OK; `lint_rtl --check` PASS (99 ≤ 99); `check_baremetal_only --check` OK (784 files); `yosys_shards --selftest` PASS. Merged locally onto today's `dev` (b5b4359c → 38a5c07a): every code-quality gate of #275–#284 green (31 gate/self-test commands, 0 non-zero), tree clean.

Mister-M-alt:
**[R2] Review at `63dab47e` → fix `d33b714a`** — reproduced on the hosted runner's layout with `act` (`catthehacker/ubuntu:act-latest`, Ubuntu 24.04, Yosys 0.33, gates run as the unprivileged user) and on Arch (Yosys 0.66, jemalloc present).

**[R2] MAJOR Tests — syn/yosys/ooc_selftest.py:317 — arm 73 compared the fixture library uncanonicalised, so `rtl-fast` was red on the hosted runner — FIXED (`d33b714a`).** `_preloadable_library` returned the `ldconfig -p` spelling (`/lib/<triplet>/libc.so.6` on a merged-`/usr` layout) while `select_malloc` hands an explicit `YOSYS_MALLOC` through `abs_path` (`readlink -f`) and so gives the yosys child and the header line `/usr/lib/<triplet>/libc.so.6`; `scoped_to_yosys` compared the two. Arch's `ldconfig` already answers the canonical path, which is why 75/75 held locally. Reproduced verbatim under act at `63dab47e`: `SELF-TEST FAILED [alloc-scoped-to-yosys]: yosys ran under ['/usr/lib/x86_64-linux-gnu/libc.so.6'], not the selected /lib/x86_64-linux-gnu/libc.so.6`. Arm 75 (the export-once mutant) passed on that layout only vacuously — `scoped_to_yosys` failed on the path, not on the leak. Fix: the fixture is `os.path.realpath`-ed where it is chosen, the same resolution `abs_path` performs, so arms 73–75 grade the property they name on both layouts. One file, +10/−1.

Clean lenses at `d33b714a`: Conformance (#290 criteria unchanged; the three allocator arms now bite for the right reason on both layouts), Robustness (`malloc.sh`, `ooc.sh`, `run.sh` untouched), Docs (`docs_check` 0, `gen_toc --check` OK; PR body's 72/72 corrected to 75/75).

Validation — candidate merge `611df13a` (tree `1ebe36a8`, parents `d33b714a` + `dev` `b5b4359c`, auto-merge, submodules unchanged): full `yosys-elaboration` job under act green (elaborate PASS, dp_srcs 35/35, tcl 40/40, `ooc.sh refusal self-test: 75 arm(s) passed`); on the host `ooc_selftest` 75/75, both `--selftest-alloc` PASS, `check_list_hermetic` OK, `check_baremetal_only --check` OK (811), `docs_check` 0, `yosys_shards --selftest` PASS, `pp_srcs --check` rc 0, `dp_srcs --selftest` 35/35. Hosted CI at `d33b714a`: `rtl-fast` green including `yosys-elaboration`; `rtl-full` shards in flight at the time of writing.

Verdict: POSITIVE at `d33b714a`.

Mister-M-alt:
**[R2]** The act-first refusal noted above (`remote refs changed between metadata lookup and fetch` on a PR whose base moved after it was opened) is tracked as #292; the base re-target used here was the workaround, not the fix.

PR #264, opened 2026-08-25
syn/ooc: milan_datapath_ooc.tcl generates and requires both control-plane ROMs before synthesis

Mister-M-alt:
[A3]

## Contents

- **[Status](#status)**
- **[Linked Issue / roles](#linked-issue--roles)**
- **[Description](#description)**
- **[The re-measured area: old vs new](#the-re-measured-area-old-vs-new)**
- **[The public area figures on #231 / #232](#the-public-area-figures-on-231--232)**
- **[How to get into the same state](#how-to-get-into-the-same-state)**
- **[How to validate](#how-to-validate)**
- **[Known limitations / out of scope](#known-limitations--out-of-scope)**
- **[Definition of Done](#definition-of-done)**

## Status

GREEN at head `2cace1e8`, answering [R-parallel]'s round-three pair on `e5885c4a` (non-nibble width truncation; the table's date), on top of [R0]'s six on `4eb75c25` and [R-parallel]'s three on `c41a52d6`. Base `dev` `4b381c09`, unchanged since the last round. All figures below were measured at this head or at the named historical commits; full transcripts in the [A3] comments.

`246-synoocmilan_datapath_ooctcl-never-generates-ucodehex-...` -> `dev`.

## Linked Issue / roles

Closes #246
Relates to #235 / PR #240 (whose refusal pattern and self-test this extends) and #245 / PR #262 (the Yosys member; separate branch, disjoint diffs: as of this round this PR's net diff vs `dev` is exactly `syn/ooc/milan_datapath_ooc.tcl` + `syn/ooc/ooc_tcl_selftest.py`, and the earlier shared `scripts/pp_srcs.py` edit is reverted, so the previous shared-file note is retired).

Executor: `[A3]`
External reviewers: `[R-parallel]` (round one), `[R0]` (round two) - both NEGATIVE rounds answered at this head.

## Description

`syn/ooc/milan_datapath_ooc.tcl` never generated `ucode.hex`: Vivado read the absent image as an all-zero ROM behind one `CRITICAL WARNING [Synth 8-4445]`, constant-folded the AECP uCPU, and completed rc=0 with a plausible, wrong utilization report. Two review rounds then closed the survivors of the first two fixes. The recipe now:

[...]

PR #305, opened 2026-08-31
syn/ooc: derive the include path and defines from the record, not by hand

Mister-M-alt:
## Contents

- **[Status](#status)**
- **[Linked Issue / roles](#linked-issue--roles)**
- **[Description](#description)**
- **[The defect: which design was being measured](#the-defect-which-design-was-being-measured)**
- **[Why three review rounds missed it](#why-three-review-rounds-missed-it)**
- **[The rest of the findings](#the-rest-of-the-findings)**
- **[How to get into the same state](#how-to-get-into-the-same-state)**
- **[How to validate](#how-to-validate)**
- **[Known limitations / out of scope](#known-limitations--out-of-scope)**
- **[Definition of Done](#definition-of-done)**

## Status

🟢 GREEN at `a8d79890`. Self-test **55/55** (was 40/40), full local bar passes, and the suite is proven RED on three separate replanted defects: the pre-fix include path (the exact recipe the previous 40 arms certified), a clock guard hoisted above `synth_design`, and `-sv` moved to the Verilog list. Base `dev` `8043f299`. No Vivado run (see limitations).

Round two answers a three-agent review of `bc7b62fb`, which found two majors in the fix itself and two in the test it shipped — all closed at this head, in the second commit.

`246-synoocmilan_datapath_ooctcl-never-generates-ucodehex-...` → `dev`.

## Linked Issue / roles

Follow-up to #246 / PR #264 (merged `0d4fff94`), from a four-agent post-merge review of that PR. #264's branch was deleted on merge, so this is the same branch name recreated from `dev`; it does not revert or re-litigate #264, it closes what the review found still open on `dev`.

Relates to #235 (the "two different strings" class) and #116 (the default-on gPTP fabric, whose sources joined the record after #264 merged and make `define=SYNTHESIS` live).

## Description

`syn/ooc/milan_datapath_ooc.tcl` derived its **sources** from `dp_srcs.py`'s record and then spelled by hand the one list that decides **which design gets measured**. This PR makes the recipe consume the whole record and gives the self-test eyes on the whole `synth_design` call.

| piece | change |
|---|---|
| `syn/ooc/dp_srcs.py` | `--record` prints the whole consumable record (`top=`/`define=`/`incdir=`/`src=`) in authority order. Default output is byte-identical, so the suite's read-set equality is unaffected. |
| `syn/ooc/milan_datapath_ooc.tcl` | consumes all four keys; refuses an unrecognized key, zero-or-many tops, an empty include or source list. No hand list remains. Plus 8 smaller refusals below. |
| `syn/ooc/ooc_tcl_selftest.py` | 40 → 49 arms. The `synth_design` stub records `-top`/`-part`/`-mode`/`-include_dirs`/`-verilog_define` and whether `-sv` was used. |
| `.gitignore` | `syn/ooc/.ooc-mut-*` — the entries named `syn/yosys/` while the mutants land in `syn/ooc/`. |

## The defect: which design was being measured

`syn/yosys/run.sh:133-138` puts `configs/generated/endstation_arty_current` **first** for `milan_datapath` — `incdirs_for()` exists for no other reason, and says so. The recipe had it **last**, and *both* `hdl/common/csr/gen/` and `configs/generated/*/gen/` carry an `adp_shape_defaults.svh`. Include search is first-match-wins, so the two orders elaborate two different entities — both cleanly, neither warning:

```
AUTHORITY order (run.sh)   ADP_TALKER_SRC_C=1  AEM_NAME_ENTRIES_C=29  TALKER_WIRE_CHANS_C=2
RECIPE order   ($INCS)     ADP_TALKER_SRC_C=2  AEM_NAME_ENTRIES_C=31  TALKER_WIRE_CHANS_C=8
```

Reproduced independently with `sv2v` and with `yosys` (whose parse error names the file it actually opened). Those constants are not decorative: `ADP_TALKER_SRC_C` → `ACMP_SRC_C` (`milan_datapath.sv:1535`) feeds `.N_STREAM_IN_P`/`.N_STREAM_OUT_P` at `:6854-6855`, and `AEM_NAME_ENTRIES_C` feeds `.DESC_NAME_ENTRIES_P` at `:6865` — they size the processor, the dominant block this instrument exists to price. `hdl/common/csr/gen/` is the last-writer-wins artifact `endstation_builder.py --write-rtl` overwrites, and it currently holds the **ax7101_1x1_tdm8** shape.

So the recipe reported a complete, plausible utilization figure for a design the portability gate never elaborates. That is #246's own class one layer up.

### What this does NOT fix

`milan_csr.sv` sits beside its **own** `hdl/common/csr/gen/` copy, and a quoted `` `include `` resolves against the including file's directory before any `-I`. So under the authority order `milan_csr` still sees 2/31 while `milan_datapath` sees 1/29 — the two-includer invariant asserted at `milan_datapath.sv:1526` (*"milan_csr \`include-s the SAME file … are ONE constant"*) does **not** hold on this tree, and does not hold for the Yosys gate either. The pre-fix order was coherent *and wrong*; the authority order is *right for `milan_datapath`* and still incoherent. Matching the gate is this recipe's job; making the tracked copy agree with the config is `endstation_builder.py --write-rtl`'s, and is not attempted here. This is stated in the recipe and is the single most important thing a reviewer should push back on.

Second half of the same defect: the record's `define=SYNTHESIS` was dropped entirely. That is **not** inert on today's `dev` — the gPTP sources joined the record after #264 merged, and `gptp-processor/hdl/top/KL_gptp_engine.sv` gates simulation-only `$error` blocks and one extra register on `` `ifndef SYNTHESIS `` at **five** sites (`:695,725,732,795,808`). It is the only file in the 115-source record that does. The gate defines it; the recipe did not.

Both halves are fixed by deriving, not spelling. `$INCS` and `$DEFINES` are now the record's own, in the record's own order.

## Why three review rounds missed it

The self-test's `synth_design` stub inspected **only** `-generic`:

```tcl
if {[lindex $args $i] ne "-generic"} continue
```

`-top`, `-part`, `-mode`, `-include_dirs` were never read, and `read_verilog` explicitly discarded `-sv`. Five separate one-line mutations of that call each left the suite at `40 arm(s) passed`. The stub now records all of them, the positive arm asserts the include path equals the record's **in order** and the defines equal the record's, and five mutation arms prove the detector fires.

## The rest of the findings

Each was confirmed by reproduction, and each now has an arm or a guard:

- **`exec` aborts on child stderr even at exit 0.** One Python warning from a generator killed the run — with the warning as the entire diagnostic, `$img` already deleted, and `$tmp` orphaned because cleanup was only on the validation path. `-ignorestderr` + `catch`; the exit *status* is the contract, and `$tmp` is cleaned on both failure paths.
- **`UPC_W_C` reached `expr` with no guard**: `UPC_W_C = 0` asked for a one-word ucode image — "the same lie as an absent one" by the recipe's own contract, admitted through the single geometry constant that skipped a check.
- **Tcl read a leading zero as octal**, so `UPC_W_C = 011` was 9 in the recipe and 11 in the RTL. `pkg_num` now returns a decimal.
- **`TAG` went through `expr`**, which coerces numerically: `TAG=007` wrote `util_7.rpt` and collided with a `TAG=7` run.
- **The `$TAG` reports were the one artifact nothing invalidated** — written only on success, so a refused run left the previous run's numbers under the same name. Now deleted before `synth_design`, like the images.
- **`get_ports -quiet`** on a renamed port would hand `create_clock` an empty object list and define a *virtual* clock: an unconstrained design reported as a timing result, at rc=0.
- **`rom_check` stripped only `//`** while `pkg_num` stripped both forms, so a generator banner written as `/* */` was a false refusal.
- **`one_source` and `positive_depth` had zero arm coverage**; disabling either left the suite green.
- **The malformed-image arms had no `check`**, so nothing asserted the publish-after-validate ordering.
- **`set_msg_config`'s model was a two-option whitelist**, blind to `-suppress`. It now models it and fails closed on an option it does not know.
- **`derived_sources()` ran `dp_srcs.py` with a bare `check=True`**, so a missing submodule aborted the suite with a traceback that *also discarded every arm failure already collected*. The record is now proven available at arm zero and names itself when it is not.
- **Arm 27 asserted the bare substring `SRC_LINES`**, which Tcl's own undefined-variable error satisfied by echoing the source line. It now plants an empty record and requires a refusal the recipe authors.
- The recipe quoted **7,923 LUT**, the issue title's first estimate; #264's measured figure at its own head is **8,012**.

```sh
git fetch origin && git checkout 246-synoocmilan_datapath_ooctcl-never-generates-ucodehex-it-reports-30925-lut-for-a-design-that-is-38848-and-exits-0
git submodule update --init third_party/verilog-axis protocol-processor gptp-processor
```

Needs `tclsh`, `python3` and pinned `sv2v` on `PATH` (README.md's tool table) — the same set the Yosys gate needs.

## How to validate

```sh
python3 syn/ooc/ooc_tcl_selftest.py          # 55 arms, ~7 min
```

Prove the new arms bite — restore the pre-fix hand-spelled include path and the suite must go RED on `dp-empty-dir-generates-and-passes`, naming the order mismatch:

```sh
python3 - <<'PY'
import io; p="syn/ooc/milan_datapath_ooc.tcl"; s=io.open(p).read()
io.open(p,"w").write(s.replace("set DEFARGS {}\n", "set DEFARGS {}\nset INCS [list "
  "$REPO/hdl/common $REPO/hdl/common/csr $REPO/hdl/common/eth_event_counter "
  "$REPO/hdl/ieee17221/adp $REPO/hdl/ieee8021q/ts $REPO/hdl/ieee8021as/ptp_timestamp "
  "$REPO/hdl/ieee1722/avtp $REPO/configs/generated/endstation_arty_current]\n", 1))
PY
python3 syn/ooc/ooc_tcl_selftest.py ; git checkout -- syn/ooc/milan_datapath_ooc.tcl
```

See the shape divergence directly, with the front end the gate itself uses:

```sh
printf 'module p;\n`include "gen/adp_shape_defaults.svh"\nendmodule\n' > /tmp/p.sv
sv2v -Iconfigs/generated/endstation_arty_current -Ihdl/common -Ihdl/common/csr /tmp/p.sv | grep TALKER_SRC
sv2v -Ihdl/common -Ihdl/common/csr -Iconfigs/generated/endstation_arty_current /tmp/p.sv | grep TALKER_SRC
```

Plus the standard bar: `scripts/lint_rtl.py --check`, `scripts/pp_srcs.py --check --selftest`, `scripts/docs_check.py`, `scripts/check_doc_paths.py`, `scripts/gen_toc.py --check`, `scripts/ci_events.py --check`, `syn/ooc/dp_srcs.py --selftest`, `git diff --check`.

- **No Vivado run.** The include-order consequence is demonstrated with `sv2v` and `yosys` and rests on first-match-wins include search — documented behaviour, and corroborated in-project by `incdirs_for()` existing only to prepend that directory. It is not confirmed against the tool. **A real Vivado run should re-baseline the absolute figures: every number this recipe has produced was for the other shape**, including #264's 39,776 and the `docs/findings/PP_SHADOW_AREA_0812.md` pair. No figure in this PR is offered as a measurement.
- **`syn/yosys/ooc.sh:48` has THIS PR's defect, live and unfixed.** Its `INC` is one fixed string for every top — `hdl/common/csr` present, **no `configs/generated/*` at all** — and `milan_datapath` is one of its tops, so it prices the ax7101_1x1_tdm8 entity too. `docs/design/AREA_BUDGET.md` quotes its figures as the campaign's LUT method. This is the more important sibling and it deserves its own PR; the round-one text of this section named `pp_shadow_ooc.tcl` instead, which is wrong on this axis — that recipe's record has zero `` `include ``s and zero `SYNTHESIS` gating, so its missing `-include_dirs`/`-verilog_define` bite nothing today.
- **`syn/yosys/run.sh:130`'s comment is false** — *"milan_datapath and nothing else \`include's the elaboration-shape header"* — `milan_csr.sv:1251` does too, and `incdirs_for()` special-cases only `milan_datapath`, so `run.sh`'s own `milan_csr` top is elaborated on the tracked shape. Not fixed here.
- **`syn/ooc/pp_shadow_ooc.tcl` still has #246's ROM hole** and is NOT fixed here. Its preflight is `file exists` + `size != 0` only — no generation, no geometry check, no `Synth 8-4445` promotion — so a truncated `ucode.hex` passes it, `$readmemh` part-fills 2,047 words with X, and it exits 0 with a full report. It also still does the unguarded `get_ports clk_i` this PR just fixed in the datapath recipe. `milan_datapath_ooc.tcl`'s header now states which properties the two recipes no longer share.
- **Nothing gates the authority's own include ORDER.** The positive arm asserts *recipe == record*; flip `incdirs_for()` from prepend to append and the recipe follows it into the wrong shape with the suite still green. `milan_datapath.sv:1553`'s elaboration guard accepts both 1 and 2 at the default `N_STREAMS=1`, so it does not catch it either.
- **Geometry remains content-blind.** A geometrically perfect all-zero image passes every check. #264 justified this with "content doesn't move utilization"; the counter-argument is that #246's own number *was* a content effect — the all-zero ROM is what constant-folded the uCPU. Closing it means hashing generator output; not attempted here.
- **The stub's canonical geometry map is still test knowledge** and still duplicates the pinned packages' numbers. A comment now names the seam, so a coherent ROM change fails with a message that points at the right file.
- **`$REPO/protocol-processor/.../gen_*.py` remain hand-spelled.** The generators are not in the record (`dp_srcs.py` emits only `.sv`/`.v`), so there is nothing to derive them from today. A pin bump that relocates one fails loudly.
- **Runtime is ~5 min**, ~90 % of it redundant `dp_srcs.py` runs (one per datapath arm). A per-process cache brings it to ~23 s; not done here because the arms that sabotage the record must keep the real generator, and splitting them is its own change.
- CI runs the self-test, not Vivado (`.github/workflows/rtl-fast.yml:167`, `timeout-minutes: 30`).

## Definition of Done

- [ ] Linked Issue acceptance criteria are satisfied 
- [x] New or changed behavior has self-checking tests
- [x] Self-test evidence is posted in a PR comment
- [x] No undocumented requirement or interface change remains
- [ ] Candidate merge result is validated per [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] Documentation is updated where needed 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,12 +132,11 @@ syn/yosys/.ooc-mut-*
 syn/yosys/.ooc-pkg-*
 syn/yosys/rom_digests.tsv.new.*
 
-# syn/ooc/ooc_tcl_selftest.py does the same beside ITSELF, for the same
-# reason (milan_datapath_ooc.tcl derives $REPO from [info script], so a
-# mutant must live in syn/ooc/). These lines were missing while the mutants
-# were being written -- only the syn/yosys/ suite was covered.
-syn/ooc/.ooc-mut-*
-syn/ooc/.ooc-pkg-*
+# syn/ooc/ooc_tcl_selftest.py writes its mutated recipes beside ITSELF for
+# the same reason (milan_datapath_ooc.tcl derives $REPO from [info script],
+# so a mutant must live in syn/ooc/). This line was missing while those
+# mutants were being written -- only the syn/yosys/ suite was covered. Its
+# package mutants take no dir= and land in $TMPDIR, so they need no entry.
 
 # local agent scratch, never shipped
 .claude/
--- a/.gitignore
+++ b/.gitignore
@@ -132,6 +132,13 @@ syn/yosys/.ooc-mut-*
 syn/yosys/.ooc-pkg-*
 syn/yosys/rom_digests.tsv.new.*
 
+# syn/ooc/ooc_tcl_selftest.py does the same beside ITSELF, for the same
+# reason (milan_datapath_ooc.tcl derives $REPO from [info script], so a
+# mutant must live in syn/ooc/). These lines were missing while the mutants
+# were being written -- only the syn/yosys/ suite was covered.
+syn/ooc/.ooc-mut-*
+syn/ooc/.ooc-pkg-*
+
 # local agent scratch, never shipped
 .claude/
 
--- a/docs/design/AREA_80_CAMPAIGN.md
+++ b/docs/design/AREA_80_CAMPAIGN.md
@@ -18,7 +18,7 @@ method is *static-conversion first* (this page), block diets second.
 > **substitution**, and it is measured, in context, both sides on the same
 > instrument, in
 > [`../findings/PP_SHADOW_AREA_0812.md`](../findings/PP_SHADOW_AREA_0812.md)
-> — **+6,956 LUT for the processor plane against −15,474 LUT removed.**
+> — **+6,956 LUT for the processor plane against −15,474 LUT removed.** (the `+6,956` pair was measured on the `ax7101_1x1_tdm8` shape -- see the shape-provenance note in [PP_SHADOW_AREA_0812.md](../findings/PP_SHADOW_AREA_0812.md) -- and is owed a re-measurement)
 >
 > Read the rest of this page as the *analysis record* it always was. Rows
 > below that name a deleted module are marked; the razors themselves still
--- a/docs/design/AREA_BUDGET.md
+++ b/docs/design/AREA_BUDGET.md
@@ -38,7 +38,7 @@ The current command-surface claim is checked against the
 > * the single biggest line item below, `u_bld` (`KL_aecp_response_builder`,
 >   5,300 LUT here), was part of an AECP plane that measured **8,645 LUT
 >   in context** for `KL_aecp_top` alone — larger than the entire replacement
->   plane costs in context (**+6,956**);
+>   plane costs in context (**+6,956**); (the `+6,956` pair was measured on the `ax7101_1x1_tdm8` shape -- see the shape-provenance note in [PP_SHADOW_AREA_0812.md](../findings/PP_SHADOW_AREA_0812.md) -- and is owed a re-measurement)
 > * the tier-3 "core, not reclaimable by parametrisation" verdict was
 >   correct about *parametrisation* and was overtaken by **deletion**: three
 >   of its five entries are gone.
--- a/docs/findings/PP_SHADOW_AREA_0812.md
+++ b/docs/findings/PP_SHADOW_AREA_0812.md
@@ -199,6 +199,16 @@ own interconnect to each plane. `syn/ooc/milan_datapath_ooc.tcl` closes that
 gap by synthesising the whole `milan_datapath` at the 1×1 shape, baseline
 versus plane-ON, same instrument:
 
+> **Shape provenance (PR #305).** The `milan_datapath` figures below were
+> taken with `syn/ooc/milan_datapath_ooc.tcl` before its include path was
+> derived from the authority record. That recipe placed the
+> elaboration-shape config dir LAST, so `milan_datapath` elaborated as
+> `endstation_ax7101_1x1_tdm8` (2 talker sources / 31 name entries / 8 wire
+> channels) rather than `endstation_arty_current` (1 / 29 / 2). The A/B
+> delta is still same-instrument -- both runs used the same shape -- but the
+> ABSOLUTE numbers are for the other entity and are owed a re-measurement.
+> `syn/yosys/ooc.sh` has the same defect and is still unfixed.
+
 | milan_datapath, 1×1 | LUT | FF | BRAM | WNS @ 100 MHz | Failing |
 |---|---|---|---|---|---|
 | baseline (`PP_PLANE_P=0`) | 33,704 | 31,321 | 24.5 | +0.825 ns | 0 |
@@ -416,7 +426,7 @@ macro-EXPANDED one that must both expand clean so the group cannot pass on a
 front end that refuses everything; and eleven that run mutated copies of the
 real `run.sh` under real bash, four of them the round-two counterexamples above
 and the last three the round-two and round-three constructs on the real read
-set, with and without a front end) and `syn/ooc/ooc_tcl_selftest.py` (5 arms,
+set, with and without a front end) and `syn/ooc/ooc_tcl_selftest.py` (55 arms,
 which drive this .tcl under `tclsh` with the Vivado commands stubbed), plus both
 expansions for real.
 
--- a/syn/ooc/dp_srcs.py
+++ b/syn/ooc/dp_srcs.py
@@ -709,6 +709,16 @@ def main() -> int:
                     help="the run.sh entry to expand (default: milan_datapath)")
     ap.add_argument("--selftest", action="store_true",
                     help="prove each check fails on a planted defect")
+    ap.add_argument("--record", action="store_true",
+                    help="print the whole consumable record -- top=/define=/"
+                         "incdir=/src= lines, authority order -- rather than "
+                         "the bare source list. A consumer that must also "
+                         "PREPROCESS these sources (Vivado, sv2v) needs the "
+                         "defines and the include path, and an include path "
+                         "spelled by hand is the drift this file exists to "
+                         "prevent: syn/yosys/run.sh puts the elaboration-shape "
+                         "config dir FIRST for milan_datapath, and any other "
+                         "order silently selects a different entity shape.")
     args = ap.parse_args()
 
     if args.selftest:
@@ -728,6 +738,13 @@ def main() -> int:
         print(p, file=sys.stderr)
     if rc:
         return rc
+    if args.record:
+        out = ["top=" + t for t in rec["top"]]
+        out += ["define=" + d for d in rec["define"]]
+        out += ["incdir=" + i for i in rec["incdir"]]
+        out += ["src=" + f for f in files]
+        print("\n".join(out))
+        return 0
     print("\n".join(files))
     return 0
 
--- a/syn/ooc/milan_datapath_ooc.tcl
+++ b/syn/ooc/milan_datapath_ooc.tcl
@@ -25,7 +25,11 @@
 # elaboration shape the yosys gate also uses.
 
 set REPO [file normalize [file dirname [info script]]/../..]
-set TAG  [expr {[info exists ::env(TAG)] ? $::env(TAG) : "base"}]
+# TAG names the .rpt files. NOT via `expr`, which coerces its operand
+# numerically: TAG=007 wrote util_7.rpt and collided with a TAG=7 run, so a
+# campaign comparing two tags compared one file against itself.
+set TAG "base"
+if {[info exists ::env(TAG)] && $::env(TAG) ne ""} { set TAG $::env(TAG) }
 puts "milan_datapath OOC: tag=$TAG"
 
 # The source list is printed by syn/yosys/run.sh itself (`--emit`) and relayed
@@ -39,7 +43,53 @@ puts "milan_datapath OOC: tag=$TAG"
 # geometry packages are found IN this record, never spelled here by hand
 # ([R0] round three on PR #264 -- a spelled path is a hand list in waiting,
 # and the pp_srcs.py literal gate cannot see Tcl semantics).
-set SRC_LINES [split [string trim [exec python3 $REPO/syn/ooc/dp_srcs.py]] "\n"]
+#
+# --record hands over the PREPROCESSOR half too (top=/define=/incdir=), not
+# just src=. Deriving the sources and then spelling the include path by hand
+# left the two halves free to disagree, and they did: run.sh puts the
+# elaboration-shape config dir FIRST for milan_datapath, this recipe had it
+# LAST, and BOTH hdl/common/csr/gen/ and configs/generated/*/gen/ carry an
+# adp_shape_defaults.svh -- so synth_design priced the ax7101_1x1_tdm8 entity
+# (2 talker sources, 31 name entries, 8 wire channels) while the portability
+# gate proved the arty_current one (1 / 29 / 2). That is #246's own class one
+# layer up: a complete, plausible report for a design nobody asked for. The
+# define half is live too -- KL_gptp_engine.sv gates simulation-only $error
+# blocks on `ifndef SYNTHESIS`, which the gate defines and this recipe did
+# not.
+#
+# -ignorestderr: `exec` treats ANY child stderr byte as an error even on a
+# zero exit, so one DeprecationWarning out of python3 would abort the run
+# with that warning as its entire diagnostic. The exit STATUS is the
+# contract, here and at every generator below.
+set DP_SRCS $REPO/syn/ooc/dp_srcs.py
+if {[catch {exec -ignorestderr python3 $DP_SRCS --record} REC]} {
+  error "milan_datapath OOC: the derived record is not available -- \
+dp_srcs.py failed. Everything below derives from it, so this is a refusal,\
+ not a license to fall back to a hand list:\n$REC"
+}
+
+set DP_TOP    {}
+set DEFINES   {}
+set INCS      {}
+set SRC_LINES {}
+foreach line [split [string trim $REC] "\n"] {
+  if {[regexp {^top=(.*)$} $line -> v]}    { lappend DP_TOP    $v; continue }
+  if {[regexp {^define=(.*)$} $line -> v]} { lappend DEFINES   $v; continue }
+  if {[regexp {^incdir=(.*)$} $line -> v]} { lappend INCS      $v; continue }
+  if {[regexp {^src=(.*)$} $line -> v]}    { lappend SRC_LINES $v; continue }
+  error "milan_datapath OOC: unrecognized record line \"$line\". A key this\
+ recipe does not know is a refusal: silently skipping it is how a consumer\
+ stops consuming half of the record it asked for."
+}
+if {[llength $DP_TOP] != 1} {
+  error "milan_datapath OOC: the record names [llength $DP_TOP] tops,\
+ expected exactly one. The module this script reads and the module it\
+ synthesizes must not be two different strings (#235)."
+}
+if {[llength $INCS] == 0 || [llength $SRC_LINES] == 0} {
+  error "milan_datapath OOC: the record carries [llength $INCS] include\
+ dir(s) and [llength $SRC_LINES] source(s); both must be non-empty."
+}
 
 # Exactly one file in the derived record may carry the named basename; zero
 # means the record no longer contains the geometry source (refuse, do not
@@ -63,12 +113,14 @@ proc one_source {src_lines tail} {
 # transition ROM by the RELATIVE name "ltn_rom.hex", KL_aecp_ucpu its
 # microcode by "ucode.hex" (UCODE_HEX_P), and the default-on fabric gPTP
 # engine's KL_gptp_ucpu its microcode by "gptp_ucode.hex"
-# (GPTP_UCODE_HEX_P); Vivado resolves each against ITS OWN run directory. This recipe is documented to run from an EMPTY directory, so
+# (GPTP_UCODE_HEX_P); Vivado resolves each against ITS OWN run directory.
+# This recipe is documented to run from an EMPTY directory, so
 # it must generate what it requires: until #246 it generated only
 # ltn_rom.hex, and Vivado read the absent ucode.hex as an all-zero ROM behind
 # one CRITICAL WARNING (Synth 8-4445), constant-folded the AECP uCPU, and
-# completed rc=0 with a full, plausible utilization report 7,923 LUT under
-# the shipping design.
+# completed rc=0 with a full, plausible utilization report 8,012 LUT under
+# the shipping design (measured at PR #264's head; the 7,923 this comment
+# used to carry was the issue title's first estimate, not a measurement).
 #
 # The contract, per image ([R-parallel] and [R0] on PR #264 closed the
 # survivors):
@@ -99,11 +151,18 @@ proc pkg_num {path name} {
  $name in $path, found [expr {[llength $hits] / 2}] (comments stripped; a\
  spelling this parser does not recognize is a refusal, not a guess)"
   }
-  return [lindex $hits 1]
+  # SystemVerilog reads an unbased literal as DECIMAL; Tcl's expr reads a
+  # leading zero as OCTAL, so `UPC_W_C = 011` would be 9 here and 11 in the
+  # RTL. scan %d fixes the base at ten.
+  return [scan [lindex $hits 1] %d]
 }
 
 proc rom_check {path digits words} {
-  set fh [open $path r]; set lines [split [read $fh] "\n"]; close $fh
+  set fh [open $path r]; set text [read $fh]; close $fh
+  # $readmemh accepts BOTH comment forms and pkg_num strips both; stripping
+  # only // here made a generator banner written as /* */ a false refusal.
+  regsub -all {/\*.*?\*/} $text "" text
+  set lines [split $text "\n"]
   set n 0
   foreach line $lines {
     regsub {//.*$} $line "" line
@@ -145,15 +204,28 @@ proc positive_depth {name d} {
   return $d
 }
 
+# UPC_W_C is not a depth, it is the EXPONENT of one: the ucode ROM is
+# 1<<UPC_W_C words. It reached `expr` with no guard, so UPC_W_C = 0 asked for
+# a ONE-WORD image -- "the same lie as an absent one" by the contract above,
+# admitted through the single constant that skipped a check.
+proc positive_exp {name e} {
+  if {![string is integer -strict $e] || $e <= 0} {
+    error "milan_datapath OOC: $name = $e is not a positive ROM address\
+ width (the image is 1<<$e words). A zero exponent would accept a one-word\
+ image, which is the absent-image lie under another name."
+  }
+  return $e
+}
+
 set UCPU_PKG [one_source $SRC_LINES ucpu_pkg.sv]
 set ACMP_PKG [one_source $SRC_LINES pp_acmp_pkg.sv]
 set GPTP_PKG [one_source $SRC_LINES gptp_ucpu_pkg.sv]
 set UCODE_W [nibble_width UCODE_W_C [pkg_num $UCPU_PKG UCODE_W_C]]
-set UPC_W   [pkg_num $UCPU_PKG UPC_W_C]
+set UPC_W   [positive_exp UPC_W_C [pkg_num $UCPU_PKG UPC_W_C]]
 set TROM_W  [nibble_width TROM_W_C [pkg_num $ACMP_PKG TROM_W_C]]
 set TROM_D  [positive_depth TROM_DEPTH_C [pkg_num $ACMP_PKG TROM_DEPTH_C]]
 set GUCODE_W [nibble_width UCODE_W_C [pkg_num $GPTP_PKG UCODE_W_C]]
-set GUPC_W   [pkg_num $GPTP_PKG UPC_W_C]
+set GUPC_W   [positive_exp UPC_W_C [pkg_num $GPTP_PKG UPC_W_C]]
 
 foreach {img gen digits words} [list \
     ltn_rom.hex $REPO/protocol-processor/hdl/acmp/rom/gen_ltn_rom.py \
@@ -164,7 +236,16 @@ foreach {img gen digits words} [list \
                 [expr {$GUCODE_W / 4}] [expr {1 << $GUPC_W}]] {
   set tmp $img.gen.[pid]
   file delete -force $img $tmp
-  exec python3 $gen -o $tmp
+  # -ignorestderr: the exit STATUS is what "a failed generator aborts" means.
+  # A generator that writes a perfect image, exits 0 and emits one Python
+  # warning used to abort here -- with the warning as the whole diagnostic,
+  # $img already deleted at the line above, and $tmp orphaned because the
+  # cleanup below is only on the validation path.
+  if {[catch {exec -ignorestderr python3 $gen -o $tmp} out]} {
+    file delete -force $tmp
+    error "milan_datapath OOC: the generator for $img exited non-zero:\
+ $gen\n$out"
+  }
   set diag "the generator exited 0 leaving no file"
   if {[file exists $tmp]} { set diag [rom_check $tmp $digits $words] }
   if {$diag ne ""} {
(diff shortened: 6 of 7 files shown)


Mister-M-alt:
## Evidence at `bc7b62fb`

All runs local, in a clean worktree of this branch with the three submodules initialised. No Vivado.

### The suite, green

```
$ python3 syn/ooc/ooc_tcl_selftest.py
OOC .tcl refusal self-test: 49 arm(s) passed
rc=0
```

### The suite, RED on the pre-fix include path

The decisive control: restore the hand-spelled `$INCS` and change nothing else. This is the exact recipe the previous **40 arms certified green**.

```
$ python3 syn/ooc/ooc_tcl_selftest.py
  - SELF-TEST FAILED [dp-empty-dir-generates-and-passes]: the include path is not the record's, IN ORDER.
  record: [.../configs/generated/endstation_arty_current, .../hdl/common, .../hdl/common/csr,
           .../hdl/ieee8021q/ts, .../hdl/common/eth_event_counter, .../hdl/ieee17221/adp,
           .../hdl/ieee8021as/ptp_timestamp]
  passed: [.../hdl/common, .../hdl/common/csr, .../hdl/common/eth_event_counter,
           .../hdl/ieee17221/adp, .../hdl/ieee8021q/ts, .../hdl/ieee8021as/ptp_timestamp,
           .../hdl/ieee1722/avtp, .../configs/generated/endstation_arty_current]
Order decides which gen/adp_shape_defaults.svh wins: hdl/common/csr/gen/ and
configs/generated/*/gen/ both carry one, so a reordered path silently elaborates a
different entity shape than the portability gate proves.
rc=2
```

### The shape divergence, two front ends

```
$ sv2v -I configs/generated/endstation_arty_current -I hdl/common -I hdl/common/csr  shape_probe.sv
    localparam signed [31:0] ADP_TALKER_SRC_C = 1;
    localparam signed [31:0] AEM_NAME_ENTRIES_C = 29;
    localparam signed [31:0] TALKER_WIRE_CHANS_C = 2;
$ sv2v <the recipe's eight -I, in the recipe's order>                                shape_probe.sv
    localparam signed [31:0] ADP_TALKER_SRC_C = 2;
    localparam signed [31:0] AEM_NAME_ENTRIES_C = 31;
    localparam signed [31:0] TALKER_WIRE_CHANS_C = 8;
```

Full-design `sv2v --top=milan_datapath` over the record under each order differs in 58 lines, including `ADP_DMAP_OUT_NSRC_C` 8 → 17, `ADP_DMAP_IN_KEYS_C` 8 → 1, and ~40 signal-width changes. Independently corroborated with `yosys`, whose parse error names the file it actually opened.

### The authority, and what the recipe now passes

```
$ bash syn/yosys/run.sh --emit milan_datapath | grep -v '^src=\|^derived='
top=milan_datapath
define=SYNTHESIS
incdir=<REPO>/configs/generated/endstation_arty_current     <-- FIRST
incdir=<REPO>/hdl/common
incdir=<REPO>/hdl/common/csr
incdir=<REPO>/hdl/ieee8021q/ts
incdir=<REPO>/hdl/common/eth_event_counter
incdir=<REPO>/hdl/ieee17221/adp
incdir=<REPO>/hdl/ieee8021as/ptp_timestamp
```

Traced out of the patched recipe with a stubbed tool:

```
MODE: out_of_context          INC[0]: endstation_arty_current
TOP:  milan_datapath          INC[1]: common
DEFINE: SYNTHESIS             INC[2]: csr
                              INC[3]: ts
GENERIC: PP_TROM_HEX_P="<abs>/ltn_rom.hex"        INC[4]: eth_event_counter
GENERIC: PP_UCODE_HEX_P="<abs>/ucode.hex"         INC[5]: adp
GENERIC: GPTP_UCODE_HEX_P="<abs>/gptp_ucode.hex"  INC[6]: ptp_timestamp
CLOCK-OK
```

`define=SYNTHESIS` is live on today's `dev`: `gptp-processor/hdl/top/KL_gptp_engine.sv:695,725,732` gate `$error` blocks and `path_stage_pending_r` on `` `ifndef SYNTHESIS ``. It is the only such file in the 115-source record — it entered after #264 merged.

### The rest of the bar

```
syn/ooc/dp_srcs.py --selftest           rc=0  dp_srcs self-test: 35 arm(s) passed
scripts/pp_srcs.py --check --selftest   rc=0
scripts/lint_rtl.py --check             rc=0  LINT GATE: PASS (99 <= ratchet 99; 22 waived)
scripts/docs_check.py                   rc=0  0 finding(s) across 182 md + 813 scrubbed
scripts/check_doc_paths.py              rc=0  1273 cited paths all resolve
scripts/gen_toc.py --check              rc=0  132 page(s) carry an annotated contents list
scripts/ci_events.py --check            rc=0  658 contract item(s) across 4 workflow files
git diff --check                        rc=0  clean
```

`git status --porcelain` clean after a full suite run; no `.ooc-mut-*` residue.

### New arms (40 → 49)

| arm | plants | asserts |
|---|---|---|
| `dp-mut-incdirs-reordered` | the pre-fix include order | the positive detector fires |
| `dp-mut-top-cross-wired` | `-top KL_pp_shadow` | " |
| `dp-mut-defines-dropped` | `{*}$DEFARGS` removed | " |
| `dp-mut-sv-flag-dropped` | `read_verilog $SV` | " |
| `dp-mut-part-changed` | `-part xc7z020clg400-1` | " |
| `dp-pkg-upc-zero` | `UPC_W_C = 0` | `not a positive ROM address width` |
| `dp-pkg-trom-depth-zero` | `TROM_DEPTH_C = 0` | `not a positive ROM depth` |
| `dp-mut-one-source-missing` | a basename absent from the record | `expected exactly one …` |
| `dp-mut-clock-port-renamed` | `CLK_NAME renamed_clk` | the virtual-clock refusal |

Retargeted, not new: `dp-mut-srcs-hand-list` now plants a fabricated **record**; `dp-mut-srcs-deleted` → `dp-mut-srcs-empty-record`, which requires a refusal the recipe authors rather than the bare substring `SRC_LINES` that Tcl's own undefined-variable error satisfied. The three malformed-image arms gained a `check` asserting the refused image is absent from the run directory.

`arm()` gained `post_synth=` for the one refusal that legitimately follows `synth_design` (the clock constraint); every other arm still must beat synthesis to the run.

### Provenance

Found by a four-agent review of merged PR #264 (Tcl/Vivado semantics · self-test rigour · integration ground truth · adversarial). Three of the four independently reached the include-path finding. #264's three earlier rounds (11 findings) are not re-litigated here — none of them covered the `synth_design` call's arguments, which is why this survived.

Mister-M-alt:
## Round two — answering the review of `bc7b62fb`

Three cleared-context reviewers audited round one (Tcl/Vivado correctness · test rigour · claim audit & blast radius). They found **two majors in the fix itself, two in the test it shipped**, and several claims my own standard rejects. All closed at `a8d79890`. One finding I am **not** fixing, and it is the one to argue about.

### Majors in the fix

**The report invalidation was in the wrong place.** I put the `file delete` beside `synth_design`, so every refusal *above* it — the record, the geometry, the generators — still left the previous run's `util_$TAG.rpt` standing under the same name. That is precisely the defect the deletion was added to close, and no arm observed it.

```
refusal                     reports after the refused run
unknown record key          all three survive ("STALE NUMBERS FROM A PREVIOUS RUN")
generator exits 3           all three survive
UPC_W_C = 0                 all three survive
synth_design fails          deleted   <- the only path the fix covered
```
Now runs before the first guard can fire; **arm 44** plants stale reports and requires them gone.

**`-ignorestderr` threw away the diagnostic.** It fixed the false abort but replaced the child's own message with Tcl's `child process exited abnormally` — so a failing `dp_srcs.py` no longer reported its `missing sources:` list. Generators now use `2>@1` (their stdout is unused, so merging keeps the reason *in* the refusal); the record call sends stderr to a file and quotes it, because its stdout is the payload.

Also: the top was **counted, not identified** (a record naming one top that was not this one passed), and `DEFINES` had no non-empty guard while `INCS`/`SRC_LINES` did — so an authority that stopped emitting it would have silently reverted this recipe to its pre-fix state.

### Majors in the test

**`post_synth=` dropped the sentinel requirement instead of inverting it**, and the stub's `get_ports` answered before `synth_design` had run. Hoisting the clock guard above synthesis — which makes Vivado refuse *every* run, since there is no design to query — left all 49 arms green. Fixed both; replanted:

```
$ python3 syn/ooc/ooc_tcl_selftest.py     # clock guard hoisted above synth_design
21 arms FAIL (was: 49 passed)
```

**The `-sv` detector was one global boolean.** Putting `-sv` on the Verilog list and reading every `.sv` without it satisfied it. The stub now records the files read *as* SystemVerilog and the positive arm requires them to be the record's `.sv` half. Replanted:

```
$ python3 syn/ooc/ooc_tcl_selftest.py     # read_verilog $SV / read_verilog -sv $V
4 arms FAIL, including dp-empty-dir-generates-and-passes
```

Also: `synth_design`'s parser failed **open** in the same commit that made `set_msg_config` fail closed (`-max_bram 0`, `-no_lc`, `-flatten_hierarchy full` silently accepted — all of which move the number); the five inverted arms asserted "*some* detector fired" rather than the one they name; and `_mutant`'s anchor miss raised a bare `AssertionError`, the identical failure mode this PR fixed for the record, at 20+ sites that trip far more often. That last one **fired twice during this round**, on anchors these very fixes had moved — and it now reports as a named prerequisite failure with the arms already run intact.

### Arms added for guards this PR itself introduced

The review's sharpest structural point: seven of the nine unobserved guards were **new in this PR**. Arms added for an unrecognized record key, `-suppress` after the promotion (advertised in the body with nothing proving it), a generator that warns and succeeds, a leading-zero `UPC_W_C`, a `/* */` ROM banner, and the generator temp's cleanup on the failure path.

The banner arm immediately earned itself: it found that the **stub's own geometry check was stricter than the tool it models** — the recipe accepted the banner, the stub refused it.

**49 → 55 arms.**

### Claims corrected

| claim | correction |
|---|---|
| the gate "proved the arty_current one (1/29/2)" | true for `milan_datapath` only — see below |
| `:1530` / `:6907` feed the shape constants | stale; real feeds are `:6854-6855` and `:6865` (those line numbers were correct at #264's merge, not at this base) |
| 3 `` `ifndef SYNTHESIS `` sites | **5** (`:695,725,732,795,808`) |
| 7,923 was "the issue title's first estimate" | it was #246's **measured** calibration at the older head `6be50377` |
| `pp_shadow_ooc.tcl` is the live sibling defect | **wrong on this axis** — it has zero `` `include ``s. `syn/yosys/ooc.sh:48` is the real one |
| "~5 min" | ~7 min (measured 429 s / 432 s independently). CI step 5m30s → 6m51s against a 30-min cap |
| "Documentation is updated where needed" | was **not honest** — zero docs changed. Now done |
| linked-Issue DoD boxes | **vacuous**: #246 is closed and no open Issue tracks this work. Box unchecked |

### The finding I am NOT fixing

Aligning to the authority fixes `milan_datapath`'s shape. It does **not** make the elaboration coherent. `milan_csr.sv` sits beside its own `hdl/common/csr/gen/` copy, and a quoted `` `include `` resolves against the including file's directory before any `-I`:

| include order | `milan_csr` | `milan_datapath` |
|---|---|---|
| pre-fix (config last) | 2 / 31 / 8 | 2 / 31 / 8 — coherent, wrong shape |
| **authority / this PR** | 2 / 31 / 8 | **1 / 29 / 2** — right for the datapath, **incoherent** |

So the invariant `milan_datapath.sv:1526` asserts — *"milan_csr \`include-s the SAME file … are ONE constant"* — does not hold on this tree, **and does not hold for the Yosys gate either**. The remedy is `endstation_builder.py --write-rtl` regenerating the tracked copy for the config being built, or `run.sh` not special-casing one top; both are outside a recipe change. I chose to match the gate and say so in the file rather than mask it, but a reviewer who wants this PR to refuse a split shape instead has a real case, and it would block the instrument until the tracked copy is regenerated.

### Verification at `a8d79890`

```
python3 syn/ooc/ooc_tcl_selftest.py     55 arm(s) passed, rc=0
  replant pre-fix include path       -> RED (dp-empty-dir-generates-and-passes)
  replant hoisted clock guard        -> RED (21 arms)
  replant -sv on the Verilog list    -> RED (4 arms)
syn/ooc/dp_srcs.py --selftest           35 arm(s) passed
scripts/pp_srcs.py --check --selftest   rc=0
scripts/lint_rtl.py --check             PASS (99 <= ratchet 99)
scripts/docs_check.py                   0 finding(s)
scripts/check_doc_paths.py              1273 cited paths resolve
scripts/gen_toc.py --check              OK
scripts/ci_events.py --check            OK (658 items)
git diff --check                        clean
```

Still no Vivado. Everything about include-search order rests on first-match-wins plus the including-file-directory rule — demonstrated with `sv2v` and `yosys`, not with the tool that will produce the number.

Mister-M-alt:
## Correction: the remedy named above is insufficient

Round two's "the finding I am NOT fixing" section says the remedy is *"`endstation_builder.py --write-rtl` regenerating the tracked copy for the config being built"*. That is too weak, and the diagnosis behind it was wrong. Correcting both, since this PR's text is the record.

**The tracked copy is not stale.** `hdl/common/csr/gen/adp_shape_defaults.svh` is byte-identical to `configs/generated/endstation_ax7101_1x1_tdm8/gen/adp_shape_defaults.svh`, and `scripts/check_entity_shape.py` arm E deliberately permits that — it is a *last-writer-wins* artifact required only to name a config that exists and to match it exactly. (I had also cited `test_builder.py:3848`'s staleness assertion as covering it; it does not — `eb.CSR_DEFAULTS_REL` is `gen/lwsrp_csr_defaults.svh`, a different file.) Regenerating it would mask the split for one config and reintroduce it for every other.

**The actual cause is a front-end semantics mismatch.** `sw/builder/test_builder.py:3843` states the assumption the override rests on:

> *"every build/sim/synth flow that compiles milan_csr.sv carries `hdl/common/csr` as an include dir, so the \`include always resolves (Verilator searches -I/+incdir and the CWD only, NEVER the including file's directory)."*

That is Verilator's rule, and under it prepending a config dir overrides the shape for **both** includers — the mechanism is sound. Vivado, yosys and sv2v use the standard quoted-include rule (including file's directory first), and `milan_csr.sv` has `gen/adp_shape_defaults.svh` adjacent to it. So on those front ends the override reaches only `milan_datapath`.

Net: **the shape-override mechanism was designed against Verilator's include semantics and is used with front ends that do not share them.** Verilator sim proves a coherent `arty_current`; the yosys portability gate and any Vivado flow elaborate a hybrid. Confirmed for sv2v and yosys by reproduction; Vivado's rule is untested here, so the synthesis half of that statement is inference.

**The fix is to remove the ambiguity, not to re-point it**: relocate the tracked shape header so it is adjacent to no `` `include``r, leaving `-I` order to decide for both files identically on every front end; add a gate asserting that no relative `` `include `` is resolvable both from its own file's directory and from a config include dir; and give the two flows that compile `milan_datapath` with no config incdir (`syn/yosys/ooc.sh:48`, `bd/build.tcl`) one. Tracked separately — this PR's mechanism is a prerequisite for it and needs no change: because the recipe now derives its include path from the record, fixing the authority fixes the recipe for free.

Mister-M-alt:
[R8] REVIEW — head `a8d79890c609c6e82c2f56f6b64087488e8f7498`, reconstructed against current `dev` `3178b13638d11d67376a24ca52ee1332cbe23ad3`

Verdict: **NEGATIVE**.

Lenses covered: **Conformance, RTL, Robustness, Tests, Docs**. Clean lenses at this head/candidate: **none**.

[R8] BLOCKER Conformance — PR #305 / closed Issue #246 / open Issue #307 — no active public task contract
Requirement/evidence: AGENTS.md §§3–5 require a Ready Issue with frozen scope, acceptance criteria, validation, executor, and independent reviewer before implementation. #246 is closed and Done; this PR explicitly says no open Issue tracks this work, reuses #246's branch after #264 merged, and has no `[A<n>] TAKEN` / `REVIEW READY` handoff. #307 was opened for the separately deferred shape defect, not as the settled contract under which this lane began.
Impact: scope, authority, acceptance criteria, roles, and completion cannot be reconstructed or judged from public state.
Required change: create/assign the substantive work item, settle its scope and relationship to #307, record executor/reviewer and exact validation, and link this lane to it before implementation continues.
Verification: a cold reviewer can reconstruct the lane from a Ready/In-progress Issue and its prefixed takeover/review-ready comments alone.

[R8] BLOCKER Conformance — `syn/ooc/milan_datapath_ooc.tcl:74`, `hdl/common/csr/milan_csr.sv:1251`, `hdl/milan/milan_datapath.sv:1532` — the new recipe knowingly elaborates a split entity shape
Requirement/evidence: the PR changes Vivado from the old coherent-but-wrong order to the record's config-first include order, yet leaves `milan_csr.sv` adjacent to the tracked header. Vivado searches the including file's directory before `-include_dirs` ([AMD UG901, Include Files](https://docs.amd.com/r/en-US/2026.1/ug901-vivado-synthesis/Include-Files)). Thus CSR binds tracked `2/31/8`, while datapath binds config `1/29/2`. #307 records the resulting advertised-two/context-one compliance defect. AGENTS.md §7 explicitly says moving a finding to another Issue does not resolve it.
Impact: the area instrument can synthesize a protocol-incoherent design while reporting success.
Required change: land the ambiguity removal with this candidate, or make this recipe fail closed when the two includers can bind different headers until #307 is resolved.
Verification: Vivado, Yosys, sv2v, and Verilator probes show both real includers bind one identical selected header; the planted ambiguity is rejected.

[R8] BLOCKER RTL — same three RTL/Tcl artifacts — frontend-dependent shape binding violates the module invariant
Requirement/evidence: the two real `include "gen/adp_shape_defaults.svh"` sites do not resolve identically under the recipe this PR constructs; the recipe itself acknowledges this at line 74.
Impact: widths, generated arrays, descriptor constants, and the synthesis result are derived from incompatible shapes.
Required change: remove or refuse the ambiguous RTL binding; an explanatory comment is not a guard.
Verification: elaborate the real top under every supported frontend and compare the constants at both include sites.

[R8] BLOCKER Robustness — same three RTL/Tcl artifacts — a supported config/include collision is accepted rather than refused
Requirement/evidence: config and tracked headers coexist today and have different values, but no pre-synthesis check detects the collision.
Impact: the recipe silently succeeds on the exact invalid configuration it documents.
Required change: add a deterministic collision/coherence guard and a mutation that reinstates the ambiguity.
Verification: coherent configs pass; a divergent adjacent/config header pair refuses before `synth_design` and invalidates stale reports.

[R8] MAJOR Tests — `syn/ooc/ooc_tcl_selftest.py:127-146`, `hdl/milan/milan_datapath.sv:89,6391` — the 55-arm harness ignores an unmodeled synthesis generic
Requirement/evidence: the stub recognizes `-generic`, but line 146 `continue`s for every name outside its ROM `*_HEX_P` regex. I planted `-generic GPTP_PLANE_EN_P=0`; the full suite still reported **55 arms, 0 failures**. That parameter structurally removes the gPTP plane.
Impact: an option that changes the synthesized design can be added while the claimed recipe-equivalence suite remains green.
Required change: validate the exact generic multiset (names and values), reject every unexpected generic, and add this mutation.
Verification: the unmodified suite passes; adding `GPTP_PLANE_EN_P=0`, deleting a required ROM generic, duplicating one, or changing a value makes the named arm fail.

[R8] MAJOR Robustness — same self-test/top artifacts — synthesis option parsing fails open for unexpected generics
Requirement/evidence: the tested command model deliberately discards arbitrary `-generic` arguments instead of proving or refusing them.
Impact: future configuration-dependent behavior can evade the refusal harness.
Required change: fail closed for any unmodeled generic and make configuration-dependent generics explicit in the oracle.
Verification: boundary/malformed/duplicate/unknown generic cases are rejected by the intended detector.

[R8] BLOCKER Tests — PR checks / exact current candidate — mandatory exact-head evidence is absent
Requirement/evidence: this ready PR has an empty status rollup and `gh pr checks 305` reports no checks. No Vivado run was performed, no repository-owned `act_ci.py` replica is recorded, and no exact-head Verilator/Yosys long-gate evidence exists. GitHub reports the PR `CONFLICTING`; a local merge-tree is clean, but its candidate changes artifacts across every lens and is not the tested branch head.
Impact: neither the changed Vivado recipe nor the candidate that could land has the required reproducible evidence.
Required change: refresh the lane onto current `dev`, fix the open findings, run the complete local gates and trusted act-first replica, then obtain hosted `rtl-fast`, `verilator-suites`, and `yosys-portability` at the unchanged ready head.
Verification: recorded commands and protected contexts all name the exact merge candidate head and pass.

[R8] MAJOR Docs — candidate tree `d4f8fcaeb18ee18372b0d26dedadcb38f4a6e147:docs/design/AREA_BUDGET.md:41` — current-base merge creates a broken historical link
Requirement/evidence: the candidate retains `../findings/PP_SHADOW_AREA_0812.md`, but that path is absent in the candidate; current `dev` moved the target to `docs/history/v1/findings/PP_SHADOW_AREA_0812.md`.
Impact: the authoritative area document lands with a dead provenance link, showing the head-only docs result does not validate the candidate.
Required change: refresh/resolve the documentation against current `dev` and point to the candidate's actual historical path.
Verification: `scripts/docs_check.py`, `scripts/check_doc_paths.py`, ToC generation, and diff checks pass on the candidate merge result.

[R8] MINOR Docs — `.gitignore:131`, `syn/ooc/ooc_tcl_selftest.py:333` — claimed OOC mutation ignore rule is absent
Requirement/evidence: the only matching rule is `syn/yosys/.ooc-mut-*`; the self-test creates `syn/ooc/.ooc-mut-*`. `git check-ignore --no-index syn/ooc/.ooc-mut-probe.tcl` returns nonzero.
Impact: interruption/failure can leave untracked mutation files despite the PR's clean-residue claim.
Required change: add the actual `syn/ooc` ignore rule or move temporary files outside the worktree, and correct the claim.
Verification: `git check-ignore -v` identifies the intended rule and interrupted/failing runs leave no visible residue.

[R8] MAJOR Docs — commits `bc7b62fb…`, `a8d79890…` and PR discussion — public history violates the repository contract
Requirement/evidence: both commits carry multi-paragraph bodies, while AGENTS.md §8 and CONTRIBUTING require one-line commits with no trailers; public executor comments also omit the required `[A<n>]` identity/handoff format.
Impact: the lane cannot satisfy the documented review/completion procedure as published.
Required change: repair the lane history and publish the required prefixed handoff against the final exact head.
Verification: every lane commit is one line with no trailers, and the Issue/PR contains the required public role and evidence records.

Reproduced green at the branch head: OOC self-test 55 arms, `dp_srcs` 35 arms, `pp_srcs --check --selftest`, lint, docs check, cited-path check, ToC, CI-event contract, and `git diff --check`. Those head-only results do not cover the planted generic, the unresolved hybrid elaboration, or the current merge candidate.

Mister-M-alt:
[A0] Closing in favour of **#310**, which lands this lane and the other one together.

[R8]'s blocking objection to #305 was that changing the include order while leaving the shape ambiguity in place is a finding moved, not resolved (AGENTS.md §7). That is correct, so the two lanes are now one contract under #309.

Every [R8] finding from this round is answered at `91ad3e24`, including the ones that were real defects here:

- three Verilator Makefiles depended on the moved header (`SHAPE_DEF`), broken on a clean checkout — fixed, and `check_entity_shape.py` arm I now inventories every consumer, file-type blind, because the extension-filtered sweep is exactly what missed them;
- the `-generic` parser failed open — `-generic GPTP_PLANE_EN_P=0` rode through a green suite; now refused, with arms for the unmodelled and duplicated cases;
- the `syn/ooc/.ooc-mut-*` ignore rule had been deleted;
- the `AREA_BUDGET.md` link died against the archival move, and #307's criterion 5 is now answered by marking the `ooc.sh` figures;
- commits are one line with no trailers.

Branch left in place for reference.

